PR #6, opened 2016-06-20
Xapian indexing improvement

automactic:
- Created a new function that add  around an element.
- Add no index comment to footnote div 

The purpose of this pull request is to create better HTML content when using omega html parser, by adding no index comments to some html elements.[reference](https://xapian.org/docs/omega/overview.html)


kelson42:
Thx!

PR #14, opened 2016-09-01
Fixed small typo in parsoid article URL construction

bshishov:
Result URL was in incorrect format:
example before: http://parsoid.url/.../articleName&oldid=1234  (missing ? delimeter for queryset arguments)
example after: http://parsoid.url/.../articleName?oldid=1234 (queryset is now ok)

### DIFF
--- a/mwoffliner.js
+++ b/mwoffliner.js
@@ -1450,7 +1450,7 @@ function saveArticles( finished ) {
     }
 
     function saveArticle( articleId, finished ) {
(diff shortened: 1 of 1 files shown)


bshishov:
Well, just double checked it. It will only work if `parsoidUrl` is set from arguments, otherwise it will have proper configured arguments because `apiUrl `has that trailing '?'

kelson42:
THX

PR #33, opened 2016-11-14
Added switch that allows to use classic parse API instead of Parsoid …

osnard:
…backend; Mainly for private wikis

See also https://sourceforge.net/p/kiwix/discussion/604121/thread/2c71cd8a/


kelson42:
Great, sorry for my lack of reaction on your post on Sourceforge... Will have look to your work ASAP. Thank you.

PR #43, opened 2017-01-28
removed --strip from pngquant line. pngquant doesn't support the flag…

darkenvy:
… or stripping in general. mwoffliner was gracefully erroring on every png

> pngquant: unrecognized option '--strip'


kelson42:
@darkenvy you need to install last version of pngquant

darkenvy:
Okay, so thats my bad :P. Thanks Kelson.

PR #52, opened 2017-03-16
18/bug/mirroring ressource loader dep

Skylsmoi:
PR resolving https://github.com/kiwix/mwoffliner/issues/18

dont forget to $ npm install after merging the changes to get node-fetch new dependency


Skylsmoi:
package should be republished on npm once merged into master

lib/mwoffliner.lib.js
```diff
@@ -25,6 +25,7 @@ var os = require( 'os' );
 var crypto = require( 'crypto' );
 var unicodeCutter = require( 'utf8-binary-cutter' );
 var htmlMinifier = require('html-minifier');
+const fetch = require('node-fetch')
```
kelson42: A bit sad, that we introduce here a new download framework/soltuion as we already have download primitives in the code.

lib/mwoffliner.lib.js
```diff
@@ -25,10 +25,11 @@ var os = require( 'os' );
...
 
 module.exports = {
     getParametersList: () => [
-        { name: 'mwUrl', description: 'Mediawiki base URL', required: true },
+        { name: 'mwUrl', description: 'Mediawiki base URL. Dont forget the trailing /', required: true },
```
kelson42: Important indeed, but better make directly a check at the line : mwUrl =.... This is easier for user.

lib/mwoffliner.lib.js
```diff
@@ -911,6 +927,99 @@ module.exports = {
                 }
             }
 
+            function storeDependencies( parsoidDoc, articleId, finished ) {
+                const filenameRadical = computeFilenameRadical()
+
+                // the encodeURI bellow is mandatory for languages with illegal letters for uri (like fa.wikipedia.org)
+                fetch(encodeURI(`${apiUrl}action=parse&format=json&page=${articleId}&prop=modules|jsconfigvars`), {
```
kelson42: It looks like the list of modules can be directly parsed-out from the HTML delivered by Parsoid. See at the end of the DOM for example https://en.wikipedia.org/w/api.php?action=visualeditor&format=json&paction=parse&page=Emmanuel

lib/mwoffliner.lib.js
```diff
@@ -911,6 +927,99 @@ module.exports = {
                 }
             }
 
+            function storeDependencies( parsoidDoc, articleId, finished ) {
```
kelson42: This method - at least all the part which does not directly impact the footer of the article, should run only one time for each ZIM. This would save us a lot of file system access.

lib/mwoffliner.lib.js
lebouquetin: Kelson, I talked to Skylsmoi and it's not possible to avoid this call : the call to api also returns js modules which are not included in the html deliverd in the example you give.

Removing this API call will result in missing modules.

lib/mwoffliner.lib.js
lebouquetin: kelson42, this function download and write module on disk only if required (ie the first time a module appears). If not required, there are no file system access at all : the check is done through redis.

The difference with what you say is that the modules are downloaded during the process instead of at the end, but there are no extra-download or extra file system access

lib/mwoffliner.lib.js
kelson42: OK, laissons ce point pour l'instant.

lib/mwoffliner.lib.js
lebouquetin: This point is not related to the current ticket. @Skylsmoi improved the text because he faced the problem. We can fix it during next developements, please open another issue.

lib/mwoffliner.lib.js
lebouquetin: "fetch" is very common and about to be standard javascript (it is already available in all major web browser). node-fetch is a common tool for those who write nodejs code on a day-to-day base.

lib/mwoffliner.lib.js
kelson42: This is the only point where I see a real problem. To know if a module is there or not, this makes a disk access, which is slow.

What I have proposed is that instead of this, the track of modules could be handled through a a simple global hash, keeping the track of what module has already been downloaded (or need to be downloaded).

If you do that, then it might be cleaner - not faster - to just then download the modules sources at the end.

If that point is fixed, then we could probably go ahead with the merge as the other one are minors - or cleared.

lebouquetin: > This is the only point where I see a real problem. To know if a module is there or not, this makes a disk access, which is slow.
> 
> What I have proposed is that instead of this, the track of modules could be handled through a a simple global hash, keeping the track of what module has already been downloaded (or need to be downloaded).

This is not true anymore: the code now write/reads in Redis. 
See https://github.com/Skylsmoi/mwoffliner/blob/1475fda27411fb7d4556c733c0ae01453da046dc/lib/mwoffliner.lib.js

> If you do that, then it might be cleaner - not faster - to just then download the modules sources at the end.

I agree with you about spliting "detection of dependencies" and "downloading dependencies".

> If that point is fixed, then we could probably go ahead with the merge as the other one are minors - or cleared.

@Skylsmoi , let us know when the code is splited.

PR #61, opened 2017-04-20
7/mobile layout

Skylsmoi:
Added a yarg (buildForMobile) which is a boolean to choose mobile design or desktop design.
Default is desktop.
Mobile design content is loaded from, for exemple, https://fr.m.wikipadie.org/api/rest_v1/page/mobile-sections/Lyon so it can be separated into sections
Desktop design is the same as before


lib/mwoffliner.lib.js
```diff
@@ -623,6 +625,17 @@ module.exports = {
                 });
         }
 
+        function saveStaticKiwixFiles( finished ) {
+            try {
+                fs.readFile('./kiwix_mobile.css', (err, data) =>
```
kelson42: Please remove "kiwix" from the naming (for the file too), just call it "mobile.css".

kelson42:
Looks pretty promising :)

Few problems:
* On dekstop, there is an arrow beside the TOP title
* On desktop the arrows beside section titles do not work (tested with FF)... but strangely it works with mobile!

I also do not understand why we have a custom hardcoded css for mobile and why we do not load it dynamically like the mobile app does. I also do not understand the difference between kiwix_mobile.css and content.parsoid.css.

Skylsmoi:
About the arrows :
- The arrow beside the top title is meant to be here. I added it so that the first section can be reduced. The mobile version of wikipedia does not have it for the top title I don't know why.
- The arrows reduce mecanisme only works for mobile, there is a breakpoint in the css from wikipedia that allow the sections to be hidden only bellow 720px width. I can remove these rules so that the reduce will work on desktop but keep in mind that you were looking at a mobile design on a desktop, which is not designed for.

All the hardcoded css is style that is inserted inside the <head> with javascript. That's why I hardcoded it since js wont be able to download and insert it in an offline version.
The difference between kiwix_mobile.css and content.parsoid.css is that kiwix_mobile.css is some styling I created entirely to match the design of mwoffliner to the mobile version of wikipedia. content.parsoid.css is from parsoid
content.parsoid.css is the file [here](https://github.com/voidlinux/void-wiki/blob/master/resources/src/mediawiki.skinning/content.parsoid.less) converted from less to css

kelson42:
@Skylsmoi Thank you for the explanations:
* Could you please remove the title arrow, we do not want to remove the section-o (there is nothing to display then)
* The arrow have to work if they are displayed. So I would remove them on desktop or make them work. Up to you (I do not have a clear opinion on that point) to make the best decision.

Issue #69, opened 2017-05-12
Docker: separate redis from default command


automactic:
**Proposal**:

In the dockerfile of this repo, change the end of file from:
```
CMD service redis-server restart && /bin/bash
```
to 
``` 
RUN service redis-server restart
CMD /bin/bash && mwoffliner
```

Reason for change:
`CMD` is the default command, can be overridden but the user. So if the user run `docker run -it openzim/mwoffliner mwoffliner --mwUrl=https://bm.wikipedia.org/ --adminEmail=chris@kiwix.org`, the following error will be produced, since the redis server start command is overridden and never executed.
```
Error: Redis connection to /dev/shm/redis.sock failed - connect ENOENT /dev/shm/redis.sock
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1086:14)
```

I recently encountered this error while building [openzim/zimfarm](https://github.com/openzim/zimfarm). I have celery workers based on `mwoffliner` container and I have to manually boot up redis.

**Update**: looks like `RUN` won't work, since it only got executed in the build phase. I personally think it is a bad idea to put redis together with mwoffliner. mwoffliner should use a docker-compose and depend on an official redis image, so that different concerns are separated. It's just like if you have a website that use mysql as backend, you would have a container for the website and another for mysql. 

In zimfarm I could get around this by using `ENTRYPOINT service redis-server restart && celery -A tasks worker --loglevel=info`

kelson42:
I kind of like to have everything in one image for simple thing like mwoffliner.

What about ?
```
ENTRYPOINT service redis-server start
CMD /bin/bash && mwoffliner
```

automactic:
I don't think it will work, because in zimfarm, I will have to override `ENTRYPOINT ` and `CMD`. And also I think you cannot have both `ENTRYPOINT ` and `CMD`, because there is only one main process of a container

kelson42:
After a small discussion it seems to me that the most promising is:
* To have a worker which just run job in other "sibling" docker containers
* logs should be caughts with "docker logs" to be then transmitted to the dispatcher.

This would avoid us complex discussion about how to X scrapper dockerfile in one worker docker file.

kelson42:
This is basically done

PR #73, opened 2017-05-15
Allow both interactive and non-interactive usage of Docker image #69

kelson42:


kelson42:
Per default it works like before. But if you want to execute a script just run:
`docker run -t mwoffliner -c cmd1 && cmd2 && cmd3`

automactic:
I shouldn't be the reviewer of this. I am not familiar with this project and have no idea how changing the docker file could impact other parts of the project.

It looks good. But I am not sure if you still could pass params to mwoffliner in this way. And what does `CMD["-I"]` do?

automactic:
Have you tested this? I don't think it works. 

If I am understanding docker correctly, things after `mwoffliner `, which I presume is the image name, will be passed to `bin/bash`, along with `--init`. It will just print out error, since `bin/bash` does not understand `-c`.

I could be wrong though. I will update `openzim/mwoffliner` and test to see if it works

automactic:
OK, I just tried it. So if you run `docker run -t openzim/mwoffliner -c --mwUrl=https://bm.wikipedia.org/`, you are gonna get:
```
/bin/bash: --: invalid option
Usage:	/bin/bash [GNU long option] [option] ...
/bin/bash [GNU long option] [option] script-file ...
[omitted]
```

kelson42:
docker run -t openzim/mwoffliner -c "mwoffliner ...." will work

PR #88, opened 2017-06-10
Video Downloading

ISNIT0:
Sorry about the whitespace changes.

I'm fairly certain this is the correct way to do this, let me know if you want it done differently


ISNIT0:
Diff without whitespace:
https://github.com/openzim/mwoffliner/pull/88/files?w=1

kelson42:
I have tested this PR with the article "Johann Sebasitan_Bach" on WPEN. This articles has video and audio files which make it really good to test this feature. You can see the Parsoid output for it here:
https://en.wikipedia.org/api/rest_v1/page/html/Johann_Sebastian_Bach

It seems to already work almost perfectly because:
* I can play both audio and video files.

But we still have two blockers:
* The file for the video is the source OGV file and it should the low-bandwidth WEBM altnernative (proposed in the HTML DOM)
* The control buttons for the audio files are invisible (see screenshot)

screen shot 2017-06-10 at 20 55 54

ISNIT0:
Just updated PR, this is what I get:
image


Videos also select the lowest quality:
image

kelson42:
Now it seems the video webm is correctly scraped instead of the ogv, but we still have problems:
* I still not do see the control for audio files on Firefox and on Chromium, I see the control, but the file is downloaded instead of been lived played.
* In addition, and this is new in this version, the video has not thumbnail, I have instead a white square and I need to click one time to see something.

PR #117, opened 2017-08-15
Extract templates and other code into separate files

subbuss:
* Updated config to point to the template files
* Updated code to read the template files

### DIFF
--- a/lib/mwoffliner.lib.js
+++ b/lib/mwoffliner.lib.js
@@ -352,9 +352,6 @@ module.exports = {
             ? ['skins.minerva.base.reset|skins.minerva.content.styles|ext.cite.style|mediawiki.page.gallery.styles|mobile.app.pagestyles.android|mediawiki.skinning.content.parsoid']
             : [];
 
-        // this module has no css, blacklisting it avoids creating an empty file that generate an error in firefox
-        // error is "style sheet could not be loaded"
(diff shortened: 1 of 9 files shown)


lib/mwoffliner.lib.js
```diff
@@ -993,42 +993,42 @@ module.exports = {
...
+                    jsDependenciesList = genericJsModules.concat(modules, modulescripts);
+                    styleDependenciesList = [].concat(modules, modulestyles, genericCssModules);
 
-                    styleDependenciesList = styleDependenciesList.filter(oneStyleDep => !config.filters.blackListCssModules.includes(oneStyleDep))
+                    styleDependenciesList = styleDependenciesList.filter(oneStyleDep => !config.filters.blackListCssModules.includes(oneStyleDep));
 
-                    printLog(`Js dependencies of ${articleId} : ${jsDependenciesList}`)
-                    printLog(`Css dependencies of ${articleId} : ${styleDependenciesList}`)
+                    printLog(`Js dependencies of ${articleId} : ${jsDependenciesList}`);
+                    printLog(`Css dependencies of ${articleId} : ${styleDependenciesList}`).
```
kelson42: @subbuss That looks strange to put a point here? Same case a few lines below!

subbuss: Fixed

PR #141, opened 2017-10-03
white page background

automactic:
Override gray background of #mw-mf-page-center, changing it to white background.

But ultimately the solution should be to remove the background-color style of #mw-mf-page-center. In the wikimed zim I am testing, the style is defined in a file in `/-/s/css_modules/inserted_style_mobile.css`, line 1199-1201. But I cannot find out where it is.


subbuss:
@kelson42 this one is for you to review

kelson42:
@subbuss Absolutly. I'm not quite happy about the way how the bug is fixed. It should be IMO in a static CSS file included. The pb is that I'm not really able to check/do that properly as the CSS inclusion is broken currently since summer with --mobileLayout - see #158 (but we had other problems before)

kelson42:
@automactic thank you for the PR, that said I think it's better to modify directly inserted_style_mobile.css. This is what I have done here... hopefully it works properly (not really able to get the grey background by me).

PR #156, opened 2017-11-05
Convert str concat to str interpolation

darkenvy:
Big commit. The console log lines are assumed safe.

By using string interpolation, we can join the "+" strings together. This makes the code appear cleaner, easier to read, and gives the reader an idea of which variables are strings at a glance.

This pull request may require more scrutiny than the others. Let me know if we should shrink this pull request.

Following AirBNB's style guide as well: http://www.jstips.co/en/javascript/keyword-var-vs-let/


lib/mwoffliner.lib.js
```diff
@@ -2615,13 +2545,7 @@ function execute(argv) {
 
       var html = '<ul>\n';
       Object.keys(articleIds).sort().map(function(articleId) {
-        html =
-          html +
-          '<li><a href="' +
-          env.getArticleBase(articleId, true) +
-          '"\>' +
```
subbuss: Worth checking separately (from this PR) why \> vs >?

subbuss:
Looks like you need to resolve merge conflicts here

lib/mwoffliner.lib.js
darkenvy: I don't think I understand what you mean by `Worth checking separately (from this PR) why > vs >?`

But yes, I fixed the merge conflicts. I was conflicting against my own changes. All fixed up.

subbuss: github ate up my characters. "\\>" vs ">" . i.e. why s the a-tag closing with a "\\>"

darkenvy: That line isn't written by me originally. I'm not sure. Escaping the ">"?

PR #165, opened 2017-12-01
Remove unused requires + purge unused dependencies

subbuss:


subbuss:
Merging directly .. we can fix problems as they occur ... instead of doing PRs right now.

Issue #149, opened 2017-10-23
Can't create Bulbapedia zim


rashiq:
I'm trying to create a zim file of bulbapedia.bulbagarden.net but it's not working. 
For test purposes here is a command you can try to only download only a single article:
`mwoffliner --mwUrl=https://bulbapedia.bulbagarden.net/ --adminEmail=rashiq@kiwix.org --articleList articles.txt`

articles.txt:

```
KAORI
```

Output:

```
root@2656a79f440d:/bulba# mwoffliner --mwUrl=https://bulbapedia.bulbagarden.net/ --adminEmail=rashiq@kiwix.org --articleList articles.txt > output.txt
Unable to download content [1] http://bulba-ad-host-website.azurewebsites.net/Content/customstyles.css (statusCode=406).
Unable to download content [2] http://bulba-ad-host-website.azurewebsites.net/Content/customstyles.css (statusCode=406).
Unable to download content [1] https://bulbapedia.bulbagarden.net/wiki/Mediawiki:offline.css?action=raw (statusCode=403).
Unable to download content [3] http://bulba-ad-host-website.azurewebsites.net/Content/customstyles.css (statusCode=406).
Absolutly unable to retrieve async. URL: Unable to download content [3] http://bulba-ad-host-website.azurewebsites.net/Content/customstyles.css (statusCode=406).
Unable to download content [2] https://bulbapedia.bulbagarden.net/wiki/Mediawiki:offline.css?action=raw (statusCode=403).
Unable to download content [3] https://bulbapedia.bulbagarden.net/wiki/Mediawiki:offline.css?action=raw (statusCode=403).
Absolutly unable to retrieve async. URL: Unable to download content [3] https://bulbapedia.bulbagarden.net/wiki/Mediawiki:offline.css?action=raw (statusCode=403).
Failed to start to optim /bulba/tmp/bulbagarden_en_articles_2017-10/favicon.png. Size should be 6022 (4012)
Error by retrieving article: Unrecognized value for parameter 'action': visualeditor
```

I'm using the mwoffliner docker image.

kelson42:
I would try ```mwoffliner --mwUrl=https://bulbapedia.bulbagarden.net/ --adminEmail=rashiq@kiwix.org --verbose --withZimFullTextIndex --localParsoid```. The ```--localParsoid``` is mandatory because it's not installed server side.

rashiq:
[...]

PR #174, opened 2017-12-10
x3 all timeouts

kelson42:
In an attempt to make mwoffliner less sensitive to slow network/backends

### DIFF
--- a/lib/mwoffliner.lib.js
+++ b/lib/mwoffliner.lib.js
@@ -208,8 +208,36 @@ function execute(argv) {
                     parsoidConfig.setMwApi({
                         uri: '${mw.base + mw.apiPath}',
                     });
+
+                    parsoidConfig.timeouts = {
+
+                        // How long does a request have to generate a response?
+			request: 12 * 60 * 1000,
+
+                        // These are timeouts for different api endpoints to the mediawiki API
+                        mwApi: {
+                        // action=expandtemplates
+                        preprocessor: 90 * 1000,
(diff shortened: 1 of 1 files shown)


lib/mwoffliner.lib.js
```diff
@@ -220,8 +220,36 @@ function execute(argv) {
                     parsoidConfig.setMwApi({
                         uri: '${mw.base + mw.apiPath}',
                     });
+
+                    parsoidConfig.timeouts = {
+
+                        // How long does a request have to generate a response?
+			request: 12 * 60 * 1000,
```
subbuss: 12 mins seems very long! On the WMF cluster we use 2 mins, but in reality, p99 numbers are ~60 sec. But, not harmful to use 12 mins .. just noticing. Maybe start with 5 mins? In any case, I want to extract this into a separate config file as well and eventually add options to pass in other config files.

subbuss:
Overall, since scraping is not time-sensitive, increasing timeouts to these large values is not a problem per se. So, I am okay with it.

lib/mwoffliner.lib.js
kelson42: Yes... I just try to find where things go wrong with #149 ... and in fact this PR seems to no 100% fix the problem. See my last comment there.

kelson42:
OK, so I run this on a stronger machine and it works... something here is not fully under control, but this is good enought to me for now.

Issue #147, opened 2017-10-12
offliners should add categories to tags


automactic:
It would be very useful for offliners to add categories to zim files. For example, wikipedia, Wikibooks, etc. Kiwix apps could use this info to group books into categories to better manage them.

We could simply add this to tags attribute. But a better (and maybe correct way) is to create a category attribute and add to the book elements.

kelson42:
Yes, will do that soon.

tim-moody:
@kelson42 Yes. I prefer a category attribute to simply creating a tag and would like these to be human readable, so Wikipedia, not wikipedia, PhET, not phet, etc.

kelson42:
done, but everything lowercase as the information comes from the hostname (so far)

PR #175, opened 2017-12-10
Add hostname radical to the ZIM tags #147

kelson42:


lib/Zim.js
```diff
@@ -99,6 +100,24 @@ Zim.prototype.computeZimName = function() {
...
+    var tags = [];
+
+    /* Mediawiki hostname radical */
+    var mwUrlHostParts = urlParser.parse(this.env.mw.base).host.split(".");
+    var mwUrlHostPartsTag = mwUrlHostParts.length > 1 ?
+	mwUrlHostParts[mwUrlHostParts.length - 2] : mwUrlHostParts[mwUrlHostParts.length - 1]
+    tags.push(mwUrlHostPartsTag.toLowerCase());
+
+    /* novid/nopic */
+    this.env.nopic ? tags.push('nopic') : this.env.novid ? tags.push('novid') : undefined;
```
subbuss: I don't like using ternary operators for this ... Better to use the same if (this.env...) pattern you use below on lines 115 and 116.

lib/Zim.js
```diff
@@ -99,6 +100,24 @@ Zim.prototype.computeZimName = function() {
     return (this.publisher ? this.publisher.toLowerCase() + '.' : '') + this.env.computeFilenameRadical(false, true, true);
 };
 
+Zim.prototype.computeZimTags = function() {
+    var tags = [];
+
+    /* Mediawiki hostname radical */
+    var mwUrlHostParts = urlParser.parse(this.env.mw.base).host.split(".");
+    var mwUrlHostPartsTag = mwUrlHostParts.length > 1 ?
+	mwUrlHostParts[mwUrlHostParts.length - 2] : mwUrlHostParts[mwUrlHostParts.length - 1]
```
subbuss: I don't have an opinion on this .. depends on what the requirement are for tags.

lib/Zim.js
```diff
@@ -99,6 +100,24 @@ Zim.prototype.computeZimName = function() {
...
+    tags.push(mwUrlHostPartsTag.toLowerCase());
```
subbuss: https://github.com/openzim/mwoffliner/issues/147#issuecomment-350498365 wanted first letter to be uppercased ... not use how PhET will be generated however.

lib/Zim.js
kelson42: Agree

lib/Zim.js
kelson42: Kind of automatic solution to "categorize" somehow ZIM files.

lib/Zim.js
kelson42: Yes, I do not see how to do that everywhere, but I have created a new --customZimTags option which goes in that direction if we really want to do that in the future.

Issue #279, opened 2018-04-07
Small icon beside external links is not displayed anymore


kelson42:
In recently created Wikipedia ZIM files, it seems that the small icon displayed at the end of "external" (non Wikipedia) URLs is not displayed anymore. This icon is very important for the users to know if they need internet to open a link, there is otherwise no way to make the difference between an "internal" and an "external" link.

Here is an exemple with the online article:
```
https://fr.wikipedia.org/wiki/Potentiel_thermodynamique#Liens_externes
```
We can see the small icon beside the link "« Thermodynamique »".

If we look to the corresponding version in the ZIM file at:
```
http://library.kiwix.org/wikipedia_fr_all_novid_2018-03/A/Potentiel_thermodynamique.html
```
The small icon is not there. Here is the bug.

Usually (to be verified), this icon is applied with the `external` class in the `a` (link) node. The class seems to be properly set (even in duplicate !?!) but for some reason it seems to fails.

I suspect here a more profound problem in the stylesheet scrapping/loading.

What is interesting is that this bug seems to occur only in the desktop (default layout). If we use the --mobileLayout option, then it seems to work, see for example: http://library.kiwix.org/wikipedia_fr_medicine_novid_2018-03/A/Art%C3%A9riopathie_oblit%C3%A9rante_des_membres_inf%C3%A9rieurs.html

PR #284, opened 2018-04-08
Added mw-parser-output class to desktop HTML so MW CSS matches

ISNIT0:
Fixes https://github.com/openzim/mwoffliner/issues/279

Some of the MW CSS didn't match the HTML as it expects a class of `mw-parser-output`.


kelson42:
Thx a lot. Please do not forget to update the changelog next tome.

Issue #74, opened 2017-05-17
Mediawiki:offline.css?action=raw is not found


masgo:
I am trying to zim an older (internatl) MediaWiki installation by using the ZIMmakerVMv6.

Doing the following:
`node /usr/local/lib/node_modules/mwoffliner/bin/mwoffliner.script.js --mwUrl=https://my.url --adminEmail=my@mail.net --mwWikiPath /intern/mediawiki --mwApiPath /intern/mediawiki/api.php 
` (url censored)
gives me:
```
Unable to download content [1] https://my.url/intern/mediawiki/Mediawiki:offline.css?action=raw (statusCode=404).
Unable to download content [2] https://my.url/intern/mediawiki/Mediawiki:offline.css?action=raw (statusCode=404).
Unable to download content [3] https://my.url/intern/mediawiki/Mediawiki:offline.css?action=raw (statusCode=404).
Absolutly unable to retrieve async. URL: Unable to download content [3] https://my.url/intern/mediawiki/Mediawiki:offline.css?action=raw (statusCode=404).
TypeError: Parameter "url" must be a string, not undefined
    at Url.parse (url.js:88:11)
    at Object.urlParse [as parse] (url.js:82:5)
    at /usr/local/lib/node_modules/mwoffliner/lib/mwoffliner.lib.js:2484:41
    at /usr/local/lib/node_modules/mwoffliner/lib/mwoffliner.lib.js:2217:21
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:676:51
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:726:13
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:52:16
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:264:21
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:44:16
    at /usr/local/lib/node_modules/mwoffliner/node_modules/async/lib/async.js:723:17
```

The file `Mediawiki:offline.css` does not exist on the server. But I checked the real wikipedia and some other public mediawiki installations and it exists nowhere.

Am I doing something wrong or is this a bug?

The MediaWiki installation I want to capture is version 1.19.20+dfsg-2.3

kelson42:
The problem is not offline.css and your mwoffliner call seems ok. Not sure what happens exactly here.

[...]

PR #344, opened 2018-07-01
Make offline.css optional so that the usual errors are not reported

bradyhunsaker:
This addresses the title of #74 though not its underlying issue.
Rather than probing for offline.css, which seemed awkward, this change allows for optional URLs.  

An unintended consequence is that such URLs will only be tried one time.  If that is a concern, an alternative is to print a message that says it was not found but is not an error and add a few lines to suppress the final error message. I'm happy to convert to that solution if you prefer.

To suppress all messages while still trying multiple times would require more extensive changes, since currently it is the presence of a message that determines whether a retry is needed.


kelson42:
@bradyhunsaker Thank you very much for that fix too. New users will for sure enjoy it by not being disturbed by this useless error message.

PR #345, opened 2018-07-01
Include the error in the error message

bradyhunsaker:
While working on PR #344 I noticed that this error message would not be very helpful since we already know that the status code is 200.
I haven't tested it.  I'm not sure how to create something that would trigger this.


kelson42:
@bradyhunsaker Yes, thx. Just kept the dynamic error code for better maintainablity.

PR #348, opened 2018-07-07
Add missing autoescape-false tag

### LINKED ISSUE #178: HTML tags visible in headers on some Wikivoyage pages
@kelson42 : For example in `wikivoyage_en_all_novid_2017-12.zim`, visit the page entitled "Hot Springs". The HTML extracted from this ZIM looks like this on many of the headers:

`<span class="mw-headline" id="Africa }}">&lt;a href="/wiki/Africa" title="Africa"&gt;Africa&lt;/a&gt;</span>`

... resulting in a rendered view like the screenshot below. I don't know if the problem is caused by the `}}` in the id, in the original source of the page, or in mwoffliner's scraping process. I've seen something like this on a lot of Wikivoyage pages from the December ZIM, so I'm guessing it's a new issue or it would have been picked up before.

image

bradyhunsaker:
Fixes #178.
I was able to recreate the problem and confirm that this fixes it.  It also makes the autoescaping consistent with the other template files.


kelson42:
@bradyhunsaker Thank you very much for the patch. I'll bump-up the version of Parsoid used, we really have to stick to version used online (like you suggested).

PR #370, opened 2018-09-03
Create debugging advice for new contributors

bradyhunsaker:
@kelson42 This is background and advice that would have been helpful to me when I was first starting on mwoffliner bugs.

Of course, feel free to change the name and edit as much as you like.  I couldn't find any naming convention for this sort of advice.


ISNIT0:
Might also be worth mentioning `--format=nozim` and `--articleList`?

kelson42:
@ISNIT0 We need to be careful about not repeating `--help`

ISNIT0:
Agreed, but those two specifically are really handy for debugging (I don't have a working copy of zimwriterfs on my Mac, and don't want to download entire wikis :))

kelson42:
@ISNIT0 You are perfectly right. That is a good idea.

bradyhunsaker:
OK, I'll edit the change in the next day or two based on the feedback from
both of you.

bradyhunsaker:
Changes made.  Please take another look.

ISNIT0:
LGTM 🎉

Issue #324, opened 2018-04-12
The inline javascript in recently-generated ZIM files can not be executed with a restrictive CSP


mossroy:
In recent ZIM files, the sections of an article are collapsed when the screen is small.
If you click on the section name, it makes the content of this section visible.

It's a great feature, but it's currently implemented with some inline javascript, at least : 
- `<script>...</script>` tags
- `onclick` attributes with some javascript inside

In some contexts, the browser engines do not allow inline javascript, for security reasons. They block it through a https://en.wikipedia.org/wiki/Content_Security_Policy
It's the case in at least two contexts used by kiwix-js : Firefox OS apps (OK, I know this is dead), AND browser extensions (at least on Firefox and Chrome).

The consequence is that, on recent ZIM files, the sections remain invisible with kiwix-js. See https://github.com/kiwix/kiwix-js/issues/355
The first reason is that the javascript contents were not injected at all (but we're working on that). The second reason is that inline javascript is blocked (when inside the browser extension).

The proper way to solve this would be to replace the inline javascript code by some external javascript files (included in the ZIM file), that would add the event handlers on the section tags. It is the recommended way to comply with the CSP.

There is also a workaround (that needs to be tested), by adding a hash of each portion of inline code inside the manifest.json (file that declares the parameters of a browser extension). But it creates a dependency between the ZIM reader and the ZIM content, which is a problem. And I'm afraid we have a big amount of different inline javascript pieces of code, because it includes the contents of onclick attributes.

An even better way to solve this would be to not use javascript to toggle the sections, and use CSS3 selectors (I did not test, but it seems possible based on https://stackoverflow.com/questions/11023816/toggle-divs-without-using-javascript). It seems to be very widely supported : https://caniuse.com/#feat=css-sel3

ISNIT0:
Thanks for pointing this out!
The problem with the suggested fix is that various bits of MW content would need to be modified to add an input.

The existing fix was designed to be as non-invasive as possible.
What's the best way to replicate your issue? I'll try moving the code into another JS file.

mossroy:
Steps to reproduce : 
- install the kiwix-js browser extension, either on Firefox (https://addons.mozilla.org/firefox/addon/kiwix-offline/) or on Chromium/Chrome (https://chrome.google.com/webstore/detail/kiwix/donaljnlmapmngakoipdmehbfcioahhk)
- open the extension, and start the developer tools inside it (ctrl-shift-i), on the console tab
[...]

PR #385, opened 2018-09-18
On mobile devices with no JS, show all content by default

ISNIT0:
#324


bradyhunsaker:
The principle makes sense to me.  This approach looks cleaner and symmetric with the CSS handling.

I haven't been following the discussion of the main issue, however, so I'm not confident what the intended goal of this change is and whether this actually solves the problem there.  Kelson should be able to review for that.

Issue #347, opened 2018-07-06
Unable to scrape dandwiki.com


fattredd:
Running in the most recent docker image. The script runs for about 4 hours before it (seems like) it finishes downloading pages. I think this is at the transition from downloading to building the articles.

Here's the command I'm running:
`mwoffliner --redis="redis://redis" --mwUrl="https://www.dandwiki.com/" --adminEmail=admin@kiwix.com --localParsoid --verbose`

Looks like a replace() call is getting called on an empty string. I've been trying to figure out why for a while now. Any ideas?

```
. . .
Downloading https://www.dandwiki.com/w/api.php?action=query&meta=siteinfo&format=json...
Successfuly optimized /root/tmp/dandwiki_en_all_2018-07/s/Pathfinder_PFSRD_Banner.jpg
Successfuly optimized /root/tmp/dandwiki_en_all_2018-07/s/OGC_Banner.jpg
Downloading http://www.dandwiki.com/w/images/3/3f/D%26D_Wiki_Logo_-_final.svg at /root/tmp/dandwiki_en_all_2018-07/favicon.png...
Downloading http://www.dandwiki.com/w/images/3/3f/D%26D_Wiki_Logo_-_final.svg...
Successfuly optimized /root/tmp/dandwiki_en_all_2018-07/favicon.png
Successfuly optimized /root/tmp/dandwiki_en_all_2018-07/favicon.png
Create main page redirection...
Saving articles...
Getting (desktop) article from http://localhost:8000/www.dandwiki.com/v3/page/pagebundle/Main_Page?oldid=1028349
Getting (desktop) article from http://localhost:8000/www.dandwiki.com/v3/page/pagebundle/MediaWiki_talk%3ACopyrightpage?oldid=85398
Downloading http://localhost:8000/www.dandwiki.com/v3/page/pagebundle/Main_Page?oldid=1028349...
Downloading http://localhost:8000/www.dandwiki.com/v3/page/pagebundle/MediaWiki_talk%3ACopyrightpage?oldid=85398...
warn/api/main { logType: 'warn/api/main',
  wiki: 'wiki$0',
  title: 'Main_Page',
  oldId: null,
  reqId: null,
  userAgent: 'MWOffliner/HEAD (admin@kiwix.com)',
  msg: 'Config Request Unrecognized parameter: \'rawcontinue\'',
  longMsg: 'Config Request\nUnrecognized parameter: \'rawcontinue\'' }
[...]

PR #389, opened 2018-09-19
#347 Made favicon handling more robust

ISNIT0:


bradyhunsaker:
The ^^^^ above is supposed to point at the keyword "async".

ISNIT0:
@bradyhunsaker The `async` in this case is not referring to the async package - it's actually new in ES6 ([async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)).

The issue will disappear if you update your NodeJS.
Which version of Node are you using? (`node --version`)

ISNIT0:
@bradyhunsaker Also, could you create an issue so it's clearly documented for other people please? :)

bradyhunsaker:
You are correct.  Some newer node versions work correctly.  I was on 6.14.3 for some reason.  I think that at a hackathon a few months ago the 8. and 10. branches weren't working for some reason.

The package.json file says node >=7.4.0 in the "engines" part.  But that version doesn't work for me.
However, the latest LTS version, 8.12.0, does work for me (as does 10.5.0).
So should we update package.json to reflect that the minimum is 8.12.0 or something?  I'll create an issue for that.

Issue #386, opened 2018-09-18
Consider switching to TypeScript


ISNIT0:
This somewhat relates to #383 

One of the barriers to entry with the current codebase is that it's pretty monolithic, and it's quite easy to break other things when fixing bugs.

TypeScript is a tool that makes the maintenance and development of large JavaScript applications much easier.

## What is TypeScript and why should we use it?
If you've got 5 minutes: https://stackoverflow.com/questions/12694530/what-is-typescript-and-why-would-i-use-it-in-place-of-javascript
If you've got 10 minutes: https://hackernoon.com/typescript-vs-javascript-b568bc4a4e58
If you've got 50 minutes: https://channel9.msdn.com/posts/Anders-Hejlsberg-Introducing-TypeScript

### Summary of benefits
TypeScript allows us to sprinkle little bits of typing into our code.
This:
- makes it easier for new developers to understand what's going on
- helps IDEs give better intellisense
- catches bugs (related to types) that humans miss

## What it would mean
- adding a build-step (TypeScript -> JavaScript)
- introducing linting (a la #383)
- we commit to this repository (and to NPM) both the source, and a built copy of MWOffliner
- using ES2015+ in our source code

It would also be a good chance to standardise formatting etc.

## Alternatives to TypeScript
- Flow (https://flow.org/)
   As far as I know, Flow is the only other reasonably mature TypeScript competitor.
   I've used it in small amounts, but I believe it's much less used/popular than TypeScript. It's worth looking at comparisons online.

## Proposed plan
1. "Switch to TypeScript" (rename files etc.) but with most checks turned off.
  This allows us to quickly get some benefit without spending lots of time up-front fixing type errors
2. Slowly enable more and more checks in the TypeScript compiler and Linter
3. ...Profit?

I have created a branch which achieves step 1: https://github.com/openzim/mwoffliner/tree/ISNIT0/code-style

What are people's thoughts on moving to TypeScript? Are there better alternatives?

subbuss:
> One of the barriers to entry with the current codebase is that it's pretty monolithic, and it's quite easy to break other things when fixing bugs.

Indeed. That is the precise reason why I started refactoring the codebase and extracting components out of it last year in the Kiwix hackathon and did more work in spurts since then. Reno also contributed to some of that. The end goal was to be able to add unit and integration tests after that was done. But, I / we couldn't finish that work. There are other components that could be extracted from the codebase still.

[...]

PR #394, opened 2018-09-20
Convert to TypeScript

ISNIT0:
Please see #386

Feel free to comment and make suggestions, I see converting to TypeScript as being a longer process and not just a one-off change.


bradyhunsaker:
Because this changes so many files I'd like to help get it merged soon.

However, I'd really prefer to see it broken up into several PRs that will be easier to review separately.  For example, many files are both reformatted and moved to a new location.  This is very hard to review, since it's really hard to tell whether additional code was added or removed at the same time.

I'm not sure exactly what depends on what, but one possibility would be:
A. Move to new locations first, with only the minimal changes in file paths.  This should be easy to review. (Also, why are you moving lib to bin/lib?)
B. Reformats only, so we can expect no added or removed code and can flag any such occurrences.
C. Any changes other than reformatting existing code, so we can focus on new and deleted code.

ISNIT0:
@bradyhunsaker Good shout.

> A. Move to new locations first, with only the minimal changes in file paths. This should be easy to review.

This is complicated by the fact that the JavaScript is renamed to TypeScript and move to `src` which is then compiled into `bin` (for backwards compatibility). We can't really just move the JavaScript to `bin/` or `bin/lib/` because then we'll have to move it to `src/` when we convert it to TypeScript.
We can't just move them to `src/` because then it will need to be called differently.
We can't just rename the JS files to TypeScript because there will be compile errors.

> "Also, why are you moving lib to bin/lib?"

We're actually moving `lib/` to `src/lib/` and `bin/` to src/` which then gets compiled to `bin` and `bin/lib/`

> B. Reformats only, so we can expect no added or removed code and can flag any such occurrences.

Formats depend on the language, so converting to TypeScript (which has some extra syntax) will affect the formatting.

C. Any changes other than reformatting existing code, so we can focus on new and deleted code.

I think what you've suggested makes sense, but we might have to make them PRs into other branches, rather than 3 separate PRs into `master`?

PR #400, opened 2018-09-22
Update node required version and added warning

ISNIT0:
#398

PR #401, opened 2018-09-23
Update node required version and added warning

ISNIT0:
See #400


ISNIT0:
After checking the files, I noticed a stray `async` keyword, I've fixed this now.
AFAICT this is good to go :)

PR #414, opened 2018-10-19
Removed duplicate templates

ISNIT0:


ISNIT0:
Moving this change into another larger PR

PR #420, opened 2018-11-03
Dry-run mode to see exact calls to mwoffliner

bradyhunsaker:
This is the first step in providing a unit test for mwmatrixoffliner and for making some basic clean up.
I couldn't resist throwing in a correction to the --mwUrl parameter.  It seems it shouldn't even be working right now.

The underlying motivations are:
- to make it easier to know exactly which mwoffliner command is used when debugging 
- to do some general cleanup of mwmatrixoffliner, once a unit test is in place

Note that this is a dry-run in the sense that it does not call mwoffliner, but it does still get data from the passed mwUrl, so network access is still required and used.


ISNIT0:
LGTM 👍

PR #428, opened 2018-11-09
Downloader Simplification

ISNIT0:
There was quite a lot of logic in the downloader, I've switched to a 3rd party library which does a lot of the processing for us with a simpler API.


bradyhunsaker:
I was coincidentally thinking about a general diff testing framework when
thinking about a different issue.

Don't wait on me, of course, but what I'm planning to create is a tool that
will make a random sample of N articles, process them with two different
versions of the code (expected use case is master vs a change), and show
the diffs in the outputs.  That would give good confidence for refactors
like this, where we expect no diffs, and for targeted changes would give us
a reasonable way to detect if there are unintended consequences in addition
to the intended changes.

I'll send more info when I have explored technology options for each step.

kelson42:
@bradyhunsaker That seems to me to be a typical test to integrate to the CI (to validate any merge request). IMO a ticket should be open specifically about that.

PR #430, opened 2018-11-09
Simple test of mwmatrixoffliner

bradyhunsaker:
This is a step in cleaning up arg handling across mwmatrixoffliner and mwoffliner.

It's just a bash script, which I hesitated to use.  But this isn't a unit test of functions--it is testing the core actions of the whole mwmatrixoffliner script.  And I think bash is available on the command lines where mwoffliner would be run.

I am open to other ideas for how to have similar tests.


ISNIT0:
LGTM, I think bash is under-used anyway :) ✨

Issue #432, opened 2018-11-11
Changelog has not been updated


kelson42:
... while fixing bugs and implementing new features for 1.6:
https://github.com/openzim/mwoffliner/blob/master/Changelog

PR #434, opened 2018-11-11
Updated Changelog

ISNIT0:
#432


kelson42:
@ISNIT0 Would be nice to put the corresponding tickets number. That is why it is easier to do it just by fixing tickets.

PR #435, opened 2018-11-12
Added namespace fallback

ISNIT0:
@kelson42 
I'd appreciate your feedback here.

For some reason `mw.namespacesToMirror` ends up with a single empty string in it. I'm not sure if this is because of API changes or not.

The `pageGeneratorQueryUrl` method was being called with a `''` as namespace which means the `gapnamespace` query parameter was being set to `undefined`.

I've now added a fallback to `0`, but I think this doesn't solve the root issue.

In the `getNamespaces` method, there is an if statement:
```
if (isContent) {
    self.namespacesToMirror.push(name);
}
```
`isContent` is always set to false and so no namespaces are mirrored.

Do you have any thoughts as to what this could be?


ISNIT0:
I've found the root of the problem (not sure when it was introduced).

PR #450, opened 2018-11-27
Download simplifications and fixes for #347

ISNIT0:


ISNIT0:
While the test shows there are some differences, these seem to change every test run (I think because there are multiple sources of similar files, and whichever one is downloaded first wins)

PR #458, opened 2018-12-09
Better Nodejs install

kelson42:


ISNIT0:
I think the best way to install node is using NVM: https://github.com/creationix/nvm

Makes things much easier IMO

PR #461, opened 2018-12-11
Tamasfabi/article list

tamasfabi:


src/mwoffliner.lib.ts
```diff
@@ -2198,8 +2198,9 @@ async function execute(argv) {
           .replace('</head>',
             genHeaderCSSLink('mobile_main_page_nojs', 'nojs') + '\n' +
             genHeaderCSSLink('style') + '\n' +
-            genHeaderScript('article_list_home') + '\n' +
```
ISNIT0: @tamasfabi Why have you removed this script?

tamasfabi: I did not, it's on 2203.

ISNIT0: 👌

PR #463, opened 2018-12-12
Remote favicon and articleList support

### LINKED ISSUE #457: Support HTTP(S) links for options asking for a path
We need to expand support of options asking for a path to a file like:
* `--articleList`
* `--customZimFavicon`

These options should also support HTTP(S) links, in that case the content would be downloaded first to `/tmp` and the `mwoffliner` would read the file from here. If the download would failed, then `mwoffliner` would failed. The temporary files should be deleted at the end of the run.

This feature is quite important/urgent to be able to make properly ZIM of sub-selection of articles from Wikipedia.

ISNIT0:
Closes #457

### DIFF
--- a/bin/mwoffliner.script.js
+++ b/bin/mwoffliner.script.js
@@ -39,6 +39,29 @@ mwofflinerLib.execute(argv)
         process.exit(0);
     })
     .catch(err => {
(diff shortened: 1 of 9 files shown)


kelson42:
@ISNIT0 I can hardly review the code, but maybe updating the usage() too would be recommended?

PR #468, opened 2018-12-18
Improved welcome page

### LINKED ISSUE #449: Improve selection welcome page
Using `mwoffliner` with selection of articles, ie. with command line argument `--articleList` has currently following consequences:
* Welcome page is a simple list of the articles (in the ZIM)
* Fill the ZIM only with the articles of the list

This is good, but the welcome page is really a problem because:
* it is "ugly" and not really helpful
* it does not scale.

I propose to following things: Replace the list of title by a full page of tilled pictures. Each picture would represent an article and in fact point to it. Because we can not create a page with 1 millions images, we would display only the first one, and scrolling down will extend the page with more pictures. All of this would be sorted from the first article in the list given to the last one.

tamasfabi:
Closes #449


lib/Utils.js
```diff
@@ -104,7 +105,7 @@ function checkDependencies(env) {
         'gifsicle --version',
         'advdef --version',
         'file --help',
-        'stat --version',
+        // 'stat --version',
```
ISNIT0: Can you un-comment this please?

src/Utils.ts
```diff
@@ -104,7 +105,7 @@ function checkDependencies(env: OfflinerEnv) {
     'gifsicle --version',
     'advdef --version',
     'file --help',
-    'stat --version',
+    // 'stat --version',
```
ISNIT0: Un-comment please :)

ISNIT0:
Closed in favour of #469

Issue #451, opened 2018-11-28
Snyk problem with publishing


kelson42:
I can not publish without snyk run and snyk wants a `node_modules` directory, but I do not want to publish the `node_modules` directory. I had to remove Snyk to publish `1.6.0`.

```
$ npm publish

> mwoffliner@1.6.0 prepare /tmp/mwoffliner
> npm run snyk-protect


> mwoffliner@1.6.0 snyk-protect /tmp/mwoffliner
> snyk protect

Missing node_modules folder: we can't test without dependencies.
Please run 'npm install' first.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mwoffliner@1.6.0 snyk-protect: `snyk protect`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the mwoffliner@1.6.0 snyk-protect script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/kelson/.npm/_logs/2018-11-28T06_40_27_164Z-debug.log
npm ERR! addLocal Could not install /tmp/mwoffliner
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mwoffliner@1.6.0 prepare: `npm run snyk-protect`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the mwoffliner@1.6.0 prepare script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/kelson/.npm/_logs/2018-11-28T06_40_27_184Z-debug.log
```

ISNIT0:
It's not necessary to publish the `node_modules` folder, but Snyk checks the installed packages for vulnerabilities and applies patches.

This kind of tool is especially important in light of https://github.com/dominictarr/event-stream/issues/116#issuecomment-442217436

kelson42:
@ISNIT0 So, what is the step-by-step publishing procedure? Could you please update the README? The classical npm login and npm publish does not work anymore.

ISNIT0:
@kelson42 What is the error you get? The publish process should be un-changed (other than that `snyk protect` runs when you run `npm publish`)

kelson42:
[...]

PR #476, opened 2018-12-23
Add publishing instructions to the README

ISNIT0:
We use Snyk to protect the users of MWOffliner, in order for Snyk to function, it needs to be able to check local dependencies. We should clone a new version of the code, install the latest versions of packages, then publish (runs `snyk protect` automatically).

Closes: #451


kelson42:
@ISNIT0 Are you really sure that way only the strict necessary will be uploaded on npmjs.org?

ISNIT0:
`npm publish` respects `.gitignore` files https://docs.npmjs.com/cli/publish

Issue #263, opened 2018-03-30
MWoffliner should be able to work without a cache


kelson42:
In many cases we do not want a cache. Having one just makes the whole stuff work slower and cache needs to be deleted manually afterward.

ISNIT0:
The `--useCache` defaults to true, to disable the cache, use `--useCache=false`

kelson42:
@ISNIT0 ok, this is a bit "complicated", would be better to just use nocache per default and then just have `--useCache` to activate it?

ISNIT0:
@kelson42 
Okay, I will do this

PR #482, opened 2019-01-02
Made minor tweak to --useCache

ISNIT0:
See #263


ISNIT0:
@kelson42 These errors are related to Snyk. I'll investigate later

kelson42:
@ISNIT0 It is mandatory to the CI to be reliable, otherwise we should just abandon this idea. No merge without green CI. So here, either the test is broken and should be (temporary) removed OR it is not and the pb should be fixed.

ISNIT0:
@kelson42 Agreed, I haven't had a chance to look yet. The command was breaking on my local machine too, I think there is a temporary problem with Snyk

PR #543, opened 2019-02-07
Renamed withoutZimFullTextIndex to withZimFullTextIndex

### LINKED ISSUE #489: Replace --withZimFullTextIndex with --withoutZimFullTextIndex
Which means that per default `zimwriterfs` is called with the option to create the FT index

ISNIT0:
Closes #489


ISNIT0:
Not merging

PR #554, opened 2019-02-11
:sparkles: Implemented proper http request backoff

ISNIT0:
Added backoff to all http requests, retries maximum 10 times with an exponential backoff pattern (see https://github.com/MathieuTurcotte/node-backoff)

Closes: #496 

I have tested this with a custom rate-limiting proxy

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1083,6 +1083,14 @@
         "js-tokens": "^3.0.2"
       }
     },
+    "backoff": {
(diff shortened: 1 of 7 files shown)


kelson42:
@ISNIT0 Does that slow down the overall speed?

ISNIT0:
@kelson42 Yes, if the requests fail

PR #582, opened 2019-02-21
:arrow_up: Updated @openzim/libzim and run `npm audit fix`

ISNIT0:


ISNIT0:
The `snyk` warning above is not important for us

PR #606, opened 2019-02-25
Strip html from title #605

### LINKED ISSUE #605: Title should be based on the DOM node <title>
Based on https://github.com/kiwix/kiwix-js/issues/475#issuecomment-467119951

AFAIK this was done properly before. We should also secure that if we have something like `<title>2<b>b</b></title>` then the `<b>` will be removed.

ISNIT0:
E.g. https://framadrop.org/r/axTlBZU4BQ#83x6hS7OxNitfbXcXEwqXOxnerN+MQDw+GTFQieXgTk=

@mossroy

Closes #605

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -3,6 +3,7 @@ import logger from './Logger';
 
 import * as urlParser from 'url';
 import ServiceRunner from 'service-runner';
+import * as domino from 'domino';
 import * as imagemin from 'imagemin';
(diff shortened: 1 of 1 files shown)


mossroy:
Sorry, I had not seen you were expected something from me here.
What did you need me to do? I read this file with kiwix-js and do not see the difference with the one you mentioned in https://github.com/kiwix/kiwix-js/issues/475#issuecomment-467093651

PR #641, opened 2019-03-30
Reduce complexity of duplicate image detection

### LINKED ISSUE #635: localeCompare in src/util/dump.ts seems to be unnecessarily slow
removeDuplicatesAndLowRes()  takes a very long time to compute if the number of dependencies is fairly large (say >100k).  Do we really need localeCompare to do the sorting?  And if so, wouldn't [obtaining a collator object beforehand](https://stackoverflow.com/questions/14677060/400x-sorting-speedup-by-switching-a-localecompareb-to-ab-1ab10/25775469) be of help?

We're comparing URLs -- localeCompare() should not be needed in this place, as URLs are non-localized.

kelson42:
Fix #635

### DIFF
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -100,9 +100,9 @@ export function removeDuplicatesAndLowRes(items: Array<{ url: string, path: stri
 
         const e = m.get(it.path);
         if (e === undefined) {
-            m.set(it.path, { ...it, mult, });
+            m.set(it.path, { ...it, mult });
         } else if (e.mult < mult) {
-            m.set(it.path, { ...it, mult, });
+            m.set(it.path, { ...it, mult });
         }
     });
     const itemsWithHighestRequiredRes = Object.values(m);
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -85,41 +85,27 @@ export async function getAndProcessStylesheets(downloader: Downloader, links: Ar
 }
 
 export function removeDuplicatesAndLowRes(items: Array<{ url: string, path: string, namespace: string }>) {
-    items = items.sort((a, b) => {
-        return a.url.localeCompare(b.url);
+    const items = items.sort((a, b) => {
+        return a.url < b.url ? -1 : (a.url > b.url ? 1 : 0);
     });
 
-    const uniqueItems = items.filter((it, index, arr) => {
-        if (index >= arr.length - 1) {
-            return true;
+    let m = new Map();
+    items.map((it) => {
+        const hasMult = item.url.includes('x.');
+        let multiplier = '1x';
+        if (hasMult) {
+            multiplier = item.url.split('-').slice(-1)[0].split('.').slice(0, -1).join('.'); // e.g. "1.5x"
         }
-        return it.path !== arr[index + 1].path;
+        const mult = Number(multiplier.slice(0, -1));
+
+        let e = m.get(it.path);
+        if (e === undefined) {
+            m.set(it.path, { ...item, mult, });
+        } else if (e.mult < mult) {
+            m.set(it.path, { ...item, mult, });
+        }
     });
-
-    const itemsWithHighestRequiredRes = uniqueItems.map((it) => {
-        const similarItems = items.filter((item) => item.path === it.path);
-
-        // const uniqueNamespaces = items.map((it) => it.namespace).sort().filter((it, i, arr) => it !== arr[i + 1]);
-        // This de-dup could cause issues if the duplicate files are in different namespaces, ignoring for now.
-
-        const itemsWithMult = similarItems.map((item) => {
-            const hasMult = item.url.includes('x.');
-            let multiplier = '1x';
-            if (hasMult) {
-                multiplier = item.url.split('-').slice(-1)[0].split('.').slice(0, -1).join('.'); // e.g. "1.5x"
-            }
-
-            const mult = Number(multiplier.slice(0, -1));
-
-            return {
-                ...item,
-                mult,
-            };
-        });
-
-        const itemsSortedByMult = itemsWithMult.sort((a, b) => a.mult < b.mult ? 1 : -1);
-        return itemsSortedByMult[0];
-    });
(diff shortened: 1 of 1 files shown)


ISNIT0:
The CI errors are linting and a merge issue

ISNIT0:
I've fixed the compile and lint errors, but the code doesn't work (the images aren't downloaded). The change doesn't behave in the same way the original did @cm8 Further changes would be appreciated

cm8:
"The images aren't downloaded" - as in "no images" or "not all images".  I'll have a look, but in theory it should work.  What did the debug output say and which command did you test with?

ISNIT0:
@cm8 Sorry, vague comment on my part - hand't had any coffee yet 😉 
Using the default VSCode launch config which is in `.vscode/launch.json` (Article list on en.wikipedia.org), no images show for the generated main page tiles.
I didn't look any further than that, unfortunately I don't have the ⌛️

cm8:
Ok, the problem lies in `Object.values(m)`:  If m is a Map this will not return the values of the map as an array.  `Array.from(m.values())` should be used instead.  I have disclosed that I did not test the patch initially, so thanks for having done review.

I have tested the following snippet with `.vscode/launch.json` now and it is ok to work.  It also does deduplication fully without prior sorting of the input items.  I have not done a 10+ hour run with a wikibooks dump to prove that it is faster (but it should be by all means, since localeCompare is eliminated):

```javascript
export function removeDuplicatesAndLowRes(items: Array<{ url: string, path: string, namespace: string }>) {
    let m = new Map();
    items.map((it) => {
        const hasMult = it.url.includes('x.');
        let multiplier = '1x';
        if (hasMult) {
            multiplier = it.url.split('-').slice(-1)[0].split('.').slice(0, -1).join('.'); // e.g. "1.5x"
        }
        const mult = Number(multiplier.slice(0, -1));

        const e = m.get(it.path);
        if (e === undefined) {
            m.set(it.path, { ...it, mult });
        } else if (e.mult < mult) {
            m.set(it.path, { ...it, mult });
        }
    });
    const itemsWithHighestRequiredRes = Array.from(m.values());
    //console.log(itemsWithHighestRequiredRes);

    logger.info(`Not downloading [${items.length - itemsWithHighestRequiredRes.length}] low-res images`);
    return itemsWithHighestRequiredRes;
}
```
Sorting `items` is absent, which is a major difference to the state the PR is currently in.  And if we should not modify `items`, consider returning it in `map()` - does not affect the result of the function.  Since we are not getting a `const items` as input, I think its ok to mutilate it, though.

I will consider using a PR next time if I have anything more to contribute to mwoffliner, but for now that should be it.

ISNIT0:
Thank you very much @cm8 🙌  I'll test and if all looks good I'll merge it

ISNIT0:
This now seems to be using only the lowest-res images for the main page 😢  (even when I add the sorting back)

Did you run `./build.sh` or `npm run build` before running? It may be that you tested an old build

ISNIT0:
Adding the sort back in fixed this, so I've pushed and will merge in a bit

cm8:
> This now seems to be using only the lowest-res images for the main page cry (even when I add the sorting back)
> 
> Did you run `./build.sh` or `npm run build` before running? It may be that you tested an old build

I've used `npm rebuild mwoffliner` for sure and double checked.  It did work here, e.g. I've seen `enwiki-2x.png` as part of the result list, when 1x and 1.5x were also available.

You should (in theory, if any imageset uses multipliers) not need the extra sorting step, because the sorting is essentially done inline by the `if (e.mult < mult)` part.

That said, it may well be that the initial sorting step chose/chooses higher resolution variants for the same path by other means than the res-factor (multiplier) logic (1x, 1.5x, 2x as part of the url).  Obviously you were not talking about the enwiki-*.png logo image.

That sorting step kills a lot of time if the project has lots of media dependencies, so we should double check if its truely needed, even though it does run ok with it now, when the performance difference on small scale is not noticable.

-----
Let's assume for a while, that we indeed have an image set of the same image, but none of the images in the set use a multiplier as part of the url.  In this case they will all be assigned a multiplier of 1x, and the first one that appears is taken (without pre-sorting).

If your high-res variants depend on sorting, without a multiplier set, then this means some part of the URL is responsible for getting it right (like `-300px.png`, `-450px.png` or `-600px.png`) - and this can go wrong, because `-1000px.png` will sort before `-900px.png` with normal string sorting.

In fact I've seen some rare images using lowres-quality in the Blender-Wikibook dump I did and I wondered what may have caused it, because it did not seem to consistently choose either a low- or highres variant.  The explanation in the last paragraph could give a cause for that observation and if it holds true, that behaviour isn't related to the latest patches.

cm8:
@ISNIT0 Please give the following replacement a try.  It should correctly parse pixel widths in thumb urls.  The hyphens and slashes in the match strings are important, because math images consist only of a hash and thus may match a simple `[0-9]+` as well (which we don't want in these cases).

```javascript
export function removeDuplicatesAndLowRes(items: Array<{ url: string, path: string, namespace: string }>) {
    const m = new Map();
    items.map((it) => {
        let mult = 1;
        let width = 1*10e6; // dummy value for unscaled media
        const widthMatch = it.url.match(/\/([0-9]+)px-/);
        if (widthMatch) {
            width = Number(widthMatch[1]);
        } else { 
            const multMatch = it.url.match(/-([0-9.]+)x\./);
            if (multMatch) {
                mult = Number(multMatch[1]);
            }
        }

        let e = m.get(it.path);
        if (e === undefined) {
            m.set(it.path, { ...it, mult, width });
        } else if (e.width < width || e.mult < mult) {
            m.set(it.path, { ...it, mult, width });
        }
    });
    const itemsWithHighestRequiredRes = Array.from(m.values());
    console.log(itemsWithHighestRequiredRes);

    logger.info(`Not downloading [${items.length - itemsWithHighestRequiredRes.length}] low-res images`);
    return itemsWithHighestRequiredRes;
}
```

ISNIT0:
@cm8 This change looks good, thanks a bunch 👍

Issue #654, opened 2019-04-03
Duplicate jsConfigVars.js entries in ZIM file


ISNIT0:

ISNIT0:
It seems the old logic for this wrote to `jsConfigVars.js` once for every article, but because it wrote to the filesystem no duplicates were made. This logic was then preserved and causes an issue because libzim creates a new entry each time.


In the `1.7` logic, we end up losing any uniqueness, and just use the data from the final article. For this fix, I won't try to merge, but use a single jsConfigVar value.

PR #655, opened 2019-04-03
:bug: Fix for #654

ISNIT0:


kelson42:
@ISNIT0 I guess this is not the right ticket number

PR #730, opened 2019-05-11
Smaller Docker image

kelson42:


ISNIT0:
@kelson42  It's better to have the installs as separate lines.
Docker caches after each `RUN` command, which means there will never be a cache before running `npm install ...`

See:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds

kelson42:
Why should we care here about the compile time. The important goal is a smaller image, that is what matters for users?

ISNIT0:
It makes a difference for developing, and testing locally. That aside, it's also one of docker's own recommended best practices

PR #753, opened 2019-05-19
Respect absolute paths for output and cache dirs

sivaraam:
We use 'path.join' to expand the path for the output and cache
directories even when an absolute path was given. This results
in the absolute path being joined with the current directory. As
a consequence, a chain of folders as mentioned in the absolute
path are created relative to the current directory. This
makes no sense as absolute paths should be treated as it is.

Correct this by correctly identifying absolute paths and expanding
only relative paths with respect to the current directory.

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -92,9 +92,16 @@ async function execute(argv: any) {
   const publisher = _publisher || config.defaults.publisher;
   let customZimFavicon = _customZimFavicon;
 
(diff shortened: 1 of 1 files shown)


ISNIT0:
Thanks for making these changes @sivaraam, looks good to me!
Would you be interested in helping with a couple of other small issues?
😄

sivaraam:
>  Would you be interested in helping with a couple of other small issues?

Sure, if I get the some time to spare :wink: 

On what issues do you want me to help you? Any links to issues (or) tags?

sivaraam:
> Sure, if I get the some time to spare :wink:

I mostly will :slightly_smiling_face:

PR #756, opened 2019-05-29
:sparkles: add pagination for subcategories

ISNIT0:
Example:
https://framadrop.org/r/taT5kYRAfz#uOVi5c+XqoocqOHh0fMjJ2Dk5j2Cbi55a44bVUl1cLg=

Look at the "Category:Container categories" article


kelson42:
Works good in general but:
* Main page has black border, the whole background should be white
* the (previous page) - (next page) should be put on a new ligne and aligned properly like online

PR #796, opened 2019-06-10
Add nscd service

kelson42:
@rgaudin So I have added nscd finally... but impossible to stop nscd "polluting" the terminal. Maybe you wil be smarter than me


kelson42:
After discussing with @rgaudin, we decided to abandon this approach in favour of another one: creating a DNS cache with dnsmasq and setup it in the ZIM farm.

PR #807, opened 2019-06-11
Codecov

kelson42:
* Ingore files which should be ignored
* Does not all big code coverage regressions (0.1% tolerance)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    patch:
+      default:
+        threshold: 0.1%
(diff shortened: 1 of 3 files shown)


ISNIT0:
@kelson42 The `.codecov.yml` is not necessary this is already defined here: https://github.com/openzim/mwoffliner/blob/master/package.json#L27

The coverage report (generated by NYC) only looks for `.ts` files not in `node_modules` or `test`

kelson42:
@ISNIT0 I have removed that part which is already handled via `nyc`.

PR #845, opened 2019-06-25
:sparkles: add check for offline.css (#614)

### LINKED ISSUE #614: offline.css retrieval generates errors
To avoid that we should properly test if the content exists before downloading it.

ISNIT0:
Closes #614


kelson42:
@ISNIT0 Would be nice to have a unit test for that method, that way codecov would probably also not complain.

Issue #795, opened 2019-06-10
geo.position HEAD node entries are not written anymore.


kelson42:
In `wikivoyage_en_all_novid_2019-04.zim` "Paris" article HTML you can read
```<meta name="geo.position" content="48.856;2.351">```

In a newly created Wikivoyage ZIM file (with 1.9.3-rc2) this entry does not exists anymore.

**Comment by kelson42** [comment]
@ISNIT0 I don't know where is the problem if this is by the retrieving of information or by the writting of the `geo.position` node entry but please create automated tests for both things and let me review the PR.

**Comment by ISNIT0** [comment]
Seems to be being written on master: `<meta name="geo.position" content="48.856;2.351">`

**Comment by kelson42** [comment]
@ISNIT0 Quite sure both code + test are wrong. I will be convinced I we have remade a full Wikivoyage ZIM like reported in the ticket and tested it properly.

```
$./node_modules/.bin/ts-node ./src/cli.ts --mwUrl=https://en.wikivoyage.org --adminEmail=kelson@kiwix.org --format=novid &> /dev/null
$ zimdump -f Paris -d out/wikivoyage_en_all_novid_2019-06.zim  | grep "geo.position"
$
```

PR #847, opened 2019-06-25
:bug: fix coordinate fetching continuation (#795)

ISNIT0:
Looks like #795 was caused because we weren't fetching the coordinates, not because of failing to write them.

Closes: #795

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -139,7 +139,7 @@ class Downloader {
     const queryOpts = {
       ...this.getArticleQueryOpts(),
       titles: articleIds.join('|'),
-      ...(this.canFetchCoordinates ? { colimit: '1' } : {}),
+      ...(this.canFetchCoordinates ? { colimit: 'max' } : {}),
       ...(this.mw.getCategories ? {
         cllimit: 'max',
         clshow: '!hidden',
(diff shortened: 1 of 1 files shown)


kelson42:
@ISNIT0 Can we please unit test this properly, we keep coming back and fixing that part of the source code. This is obviously a very sensitive one which should be really secured.

ISNIT0:
@kelson42 It's not that the code is sensitive. It's that we changed it un-necessairily.

I set the max coordinates to `1`, thinking that meant 1 per article, it actually meant one per request (so only the first of 500 articles would have coordinates).

kelson42:
@ISNIT0 How do we secure that next fix/change/bug we won't break it? If not with an automated test?

kelson42:
@ISNIT0 I propose to run to make a run a a small wiki? and check afterward that the values in the redis are ok:
* Good number of articles
* Good number of redirects
* Good number of geo coordinates
* whatever else is retrieved at that time from the `query` api.

That way we secure that we will never again make a scrape and forgetting a big part.

ISNIT0:
Redis is automatically cleared, so I don't think it's a good idea to try and use that.

I would like to do this using `node-libzim`. I spent a few hours improving the reader a while ago, but I got stuck with C++ stuff

ISNIT0:
Also, I think your above comment should be a different issue. It's a good suggestion, but I'd like to implement it as part of the existing "e2e" tests (there is already some commented out half working code).

kelson42:
@ISNIT0 Thx. Good to me, please for #851, let us discuss first about this e2e level. I don't understand why this has to be done at such a high level.

ISNIT0:
@kelson42 Are you happy with the solution here: https://github.com/openzim/mwoffliner/pull/862

kelson42:
@ISNIT0 Considering that your are handling the testing of this ticket (and a bit more) in an other ticket, please just merge this PR if you think it works fine. If you detect problem in your testing PR then please then fix them straight in that other testing PR.

PR #855, opened 2019-06-28
:whale: fixing docker build to use committed code

### LINKED ISSUE #722: Build mwoffliner from source in Docker
To avoid sync problem with npmjs.org

ISNIT0:
Closes #722

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+node_modules
+cac
+out
+test
+.vscode
+.github
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,19 +1,23 @@
 FROM node:10
 
-# Basics
-RUN apt update && apt install -y --no-install-recommends make g++ curl git
+# Install dependences
+RUN apt update && apt install -y --no-install-recommends make g++ curl git imagemagick
 
 # Install mwoffliner
-RUN apt install -y --no-install-recommends imagemagick
-RUN npm --global config set user root
-WORKDIR /mwoffliner
+WORKDIR /tmp/mwoffliner
 COPY package*.json ./
-RUN npm install
-COPY . .
+COPY src src
+COPY lib lib
+COPY res res
+COPY translation translation
+COPY index.js .
+COPY build.sh .
+RUN npm --global config set user root
+RUN npm i
 RUN npm i -g .
-RUN apt remove -y make g++
 
-# Boot commands
+# Configure launch environment
+WORKDIR /
 RUN mv /root/.bashrc /root/.old-bashrc
-COPY .custom-bashrc /root/.bashrc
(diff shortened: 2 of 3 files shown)


kelson42:
@ISNIT0 Why no CI report here?

kelson42:
This seems to fail for me:
```
$ docker build docker/ -t test
Sending build context to Docker daemon  5.632kB
Step 1/13 : FROM node:10
 ---> 90fa0611d45b
Step 2/13 : RUN apt update && apt install -y --no-install-recommends make g++ curl git
 ---> Using cache
 ---> eaa1975ed541
Step 3/13 : RUN apt install -y --no-install-recommends imagemagick
 ---> Using cache
 ---> b630fea25d78
Step 4/13 : RUN npm --global config set user root
 ---> Using cache
 ---> 381d74b826e2
Step 5/13 : WORKDIR /mwoffliner
 ---> Using cache
 ---> 592e3fce4edf
Step 6/13 : COPY package*.json ./
COPY failed: no source files were specified
```

ISNIT0:
@kelson42 Can we have a look at the dockerhub configuration please? Looks like the build context is wrong.

ISNIT0:
The CI is running, click "Show all checks"

kelson42:
@ISNIT0 What is the local command to build the image?

ISNIT0:
From that directory: `docker build . -f docker/Dockerfile`

kelson42:
This does not work for me and the Dockerfile is broken, but I will have a look.

kelson42:
@ISNIT0 I have fixed the `Dockerfile`. I believe the hub.docker.com is properly configured, but we will have to see that when we are so far.

ISNIT0:
@kelson42 
It seems you've replaced `COPY . .` with lots of individual copies... This is not idea. Please just use the `.dockerignore` file that I've already created.

kelson42:
@ISNIT0 I prefer to be explicit here. Relying only on the `.dockerignore` is implicit copying which leads to the following situation:
* Difficult to immediately know what is copied or not to the image
* Does not secure that you never publish here something which is not intended
* No way to forget to maintain the COPY instructions (which might be perfectly the case with the .dockerignore)
* Not more work to maintain the COPY instructions in the Dockerfile than to maintain the `.dockerignore`

That said:
* I might COPY too much?
* Not against keeping the `.dockerignore` like it is.

ISNIT0:
Using `.dockerignore` is explicitly listed under Docker best practices.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#exclude-with-dockerignore

If you want to keep the multiple `COPY`s, it's best to remove the `.dockerignore` as it's confusing.

Please merge when you're happy with it.

kelson42:
I want to keep the `.dockerignore` like you made it. This is a security.

PR #856, opened 2019-06-28
:sparkles: support WPRU desktop mainpage CSS

### LINKED ISSUE #841: Russin desktop main page rendering in-correctly
With 1.9.3. Look for example homepage of https://download.kiwix.org/zim/wikipedia/wikipedia_ru_all_novid_2019-06.zim

ISNIT0:
The problem (#841) is caused by the use of inline CSS on the WPRU main page that specifies the class `mw-parser-output`.

This pull request fixes the issue, but I think in the wrong way. The better solution is for WPRU to not specify the `mw-parser-output` class in it's page, as this makes the content fragile.

This CSS change was made recently, hence when testing on 1.7.1, we get the same result

Closes #841

### DIFF
--- a/res/templates/page.html
+++ b/res/templates/page.html
@@ -18,7 +18,7 @@
     <div id="mw-mf-page-center">
       <div id="content" class="mw-body">
         <a id="top"></a>
-        <div id="bodyContent" class="content">
+        <div id="bodyContent" class="content mw-parser-output">
           <h1 id="titleHeading" style="background-color: white; margin: 0;"></h1>
           <div id="mw-content-text">
           </div>


kelson42:
@ISNIT0 This seems to be a good move. Considering that this part of MWoffliner is hardcoded, we have indeed to keep it in-sync with prod version... otherwise the kind of problem we have will happen. That said, I have no better solution that doing that manually (for now).

kelson42:
But the CI fails :(

ISNIT0:
It's not a problem of the MediaWiki HTML, it's that a specific class has been used to namespace the query. The CSS would work fine on WPRU without the namespacing.

PR #861, opened 2019-07-01
:sparkles: :white_check_mark: re-write empty paragraph removal logic

ISNIT0:
@kelson42 I don't really understand the old logic, but I believe this does the same thing...


kelson42:
What is the differrence beetween a section and a paragraph? This should apply at any level.

ISNIT0:
When reading the code, "paragraph" to me means `<p>...</p>`, but section would include the `<hx></<hx>` too

kelson42:
@ISNIT0 My bad. Will do my best to use the proper English word next time.

PR #883, opened 2019-07-08
:sparkles: Check mw capabilities before starting scrape

### LINKED ISSUE #670: Availibility of coordinates should be check
Scraping with `1.8.2` seems, depending of the wiki, to generate a lot of following warnings:
```
The response contained warnings: { main: { '*': 'Unrecognized parameter: colimit.' },
  query:
   { '*': 'Unrecognized value for parameter "prop": coordinates.' } }
```

The reason is that it seems that Mediawiki is configurable to support - or don't support - API queries about geo-coordinates. If not, we get this warnings. This is the case for example with de.wikibooks.de. See #663.

We should, at start, detect if the remote Mediawiki support this or not, and if not, stop asking about the coordinates in our API requests. This should avoid the ''silly'' warnings.

ISNIT0:
Closes #670

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -30,6 +30,16 @@ const imageminOptions = {
   ],
 };
 
+interface DownloaderOpts {
+  mw: MediaWiki;
+  uaString: string;
+  speed: number;
+  reqTimeout: number;
+  useCache: boolean;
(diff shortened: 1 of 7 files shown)


ISNIT0:
Need to call `getCapabilities` in the tests

PR #906, opened 2019-07-22
Various redis improvements

### LINKED ISSUE #899: [REGRESSIONS] Redis memory consumption is far too high
I don't understand what is exactly going on here... but we have a waste of memory (in comparison to earlier version I have coded) by at least a factor x10 I guess.

I also see this kind of things in the redis:
```
{"pageid":12612419,"ns":0,"title":"Drown  in My Own Tears","redirects":[{"pageid":13209454,"ns":0,"title":"Drown  In My Own Tears"},{"pageid":14367420,"ns":0,"title":"Drown in my own  tears"},{"pageid":41493354,"ns":0,"title":"I'll Drown in My Own  Tears"}],"revisions":[{"revid":870146860,"parentid":864011370,"user":"86.147.194.118","anon":"","timestamp":"2018-11-22T18:41:50Z","comment":"/*  Other recordings */"}]}
```

Most of this should never be in the Redis, we only need title, revisionid, (date?) coordinates.

ISNIT0:
Closes #899

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -56,6 +56,7 @@ class Downloader {
   private activeRequests = 0;
   private maxActiveRequests = 1;
   private noLocalParserFallback = false;
+  private urlPartCache: KVS<string> = {};
 
   constructor({ mw, uaString, speed, reqTimeout, useCache, cacheDirectory, noLocalParserFallback }: DownloaderOpts) {
     this.mw = mw;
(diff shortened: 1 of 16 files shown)


kelson42:
* The "title", "pageid", ... should be shorten to one character.
* pageid is not needed IMO
* "ns=0" has to be replace by a simple flag if this is a category, other just nothing.
* "Thumbnail" info should be saved only if welcome page need it (selection) AND we still need one (total thumbnails found <100).
* Move redirects to the redirects
* Rename "redirects" to "redirect"

ISNIT0:
@kelson42 I've updated this, would appreciate another look 👍

kelson42:
> * The "title", "pageid", ... should be shorten to one character.

* "ti" should be replaced by optimal "i"
* "th" should be replace by optiml "h"

Still not optimized in redirect table:
```
"Friends_Society" "{"targetId":"Quakers","title":"Friends Society"}"
```

> * "Thumbnail" info should be saved only if welcome page needs it (selection) AND we still need one (total thumbnails found <100).

Seems not to work properly, I have scraped 1000 articles from WPEN, and had 189 entries in redis with a thumbnail. It should be only 100.

ISNIT0:
> Seems not to work properly, I have scraped 1000 articles from WPEN, and had 189 entries in redis with a thumbnail. It should be only 100.

They're fetched in batches, and the checking happens on each batch. Is this really such a huge issue?

kelson42:
@ISNIT0 What is the max number then?

ISNIT0:
Depends on the batch sizes and how many batches are launched in parallel, but it isn't proportional to the size of the wiki

Issue #726, opened 2019-05-10
Badly rewritten URLs in case of "/" in the article title


kelson42:
The forward slash in the URL is causing problems in Kiwix JS (see https://github.com/kiwix/kiwix-js/issues/494). I wonder if it might be causing problems for you too. For example, the article `Singapore` references its assets like this:
```
<link href="../-/s/css_modules/ext.kartographer.link.css" …>
```
Whereas the article `Singapore/Riverside` references its assets like this:
```
<link href="../../-/s/css_modules/ext.kartographer.link.css" …>
```
Note the extra `../../`, because the browser thinks it's accessing an article Riverside in the subdirectory Singapore. This would all be fine, except that the hrefs in the `Singapore/Riverside` article are written like this:
```
<p>The <b>Singapore River</b> forms a central artery in <a href="Singapore"
title="Singapore">Singapore</a>'s densely packed Central Business District.
```
Note that to access the hyperlink on this page from a browser, we would need to write that link as `<a href="../Singapore" ...>`.

So there seems to be some inconsistency. Are we inside a subdirectory from the browser's perspective, or not? The subroutine that writes the location of the assets seems to think so, while the one that writes the location of the hyperlinks doesn't! Do you think this is an mwoffliner issue?

kelson42:
I reopen the ticket as this is still not done properly. Same example with Wikivoyage `Singapore/Riverside` article. The link to `Suntec City and the Esplanade` is wrong. It is `../../A/../Singapore%2FMarina_Bay` and it should be `../../A/Singapore/Marina_Bay` (So two errors). @ISNIT0 Please create a proper unit test to test URL writing completely so we can be sure that we don't do any regression on that anymore in the future.

ISNIT0:
Yea, this is some code I've tried not to touch so far - it's very complicated and old :(

I'd like to refactor once we have good tests for it

kelson42:
Just tried again and CSS is not loaded properly in Wikivoyage "EN Paris/2nd arrondissement" article

kelson42:
Seems to work now... not sure what happend here.

kelson42:
No, it's not, I have created a new ZIM file with git master head of wikinews FR, and here is what `zimcheck` reports (we have broken internal links):
```
$ zimcheck out/wikinews_fr_all_2019-07.zim 
[INFO] Checking zim file out/wikinews_fr_all_2019-07.zim
[INFO] Verifying Internal Checksum.. 
[...]

PR #954, opened 2019-08-25
:bug: rewrite '/'s in urls for Parsoid articles, not just MCS

ISNIT0:
#726

### DIFF
--- a/src/util/rewriteUrls.ts
+++ b/src/util/rewriteUrls.ts
@@ -172,14 +172,14 @@ export async function rewriteUrl(articleId: string, mw: MediaWiki, dump: Dump, l
             }
         } else { // This is MediaWiki HTML
             await removeLinksToUnmirroredArticles(mw, dump, linkNode, href);
+        }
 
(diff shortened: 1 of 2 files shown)


kelson42:
* No automated testing to avoid that for the nth time we have url rewriting problems
* The way to rewrite URLs is really dirty, slow and highly probably buggy
-> I don't want to release a new version of mwoffliner without part tested and written properly

ISNIT0:
> - No automated testing to avoid that for the nth time we have url rewriting problems

You're right, this is something that should be tested with `zimcheck`, I'll add a `customMainPage` test.

> - The way to rewrite URLs is really dirty, slow and highly probably buggy
>     -> I don't want to release a new version of mwoffliner without part tested and written properly

Although I agree that it'd be nice to replace the existing code with the native function, there is no evidence that the current logic is buggy, the recent re-open of #726 is nothing to do with the url handling logic, it's that the logic wasn't even applied to the HTML.
I don't want to make unnecessary changes in an already delayed release.

Issue #871, opened 2019-07-03
Wikivoyage cartography links are wrong


kelson42:
All links with the class `mw-kartographer-maplink` have hte name `1`... but online they are numbered from 1 to n. See for example article "Berlin" in Wikivoyage EN.

kelson42:
@ISNIT0 I suspect this is a bug upstream, but could you please confirm?

kelson42:
To extend knowledge about this problem read the following comments on #494 https://github.com/openzim/mwoffliner/issues/494#issuecomment-453163347 and https://github.com/openzim/mwoffliner/issues/494#event-2464381379

ISNIT0:
I *think* the problem comes from the MCS response - all of the `mw-kartographer-maplink` anchor tags contain the text value of `1`.
E.g.: https://en.wikivoyage.org/api/rest_v1/page/mobile-sections/Berlin

`<a class="...... data-overlays=..... .....>1</a>`

kelson42:
@ISNIT0 If there is a bug upstream this should be reported. I suspect this is the case too here. But I'm still interested in a solution as this is a problem which has been reported many times in the past and this is really important for offline Wikivoyage versions.

For example could we replace the 1 with an character 🌍  with a simple CSS rule?

ISNIT0:
If this is not a new issue, it shouldn't be in 1.9 or 2.0

Jaifroid:
> I think the problem comes from the MCS response - all of the mw-kartographer-maplink anchor tags contain the text value of 1.

The reason they contain a text value of `1` is because numbers only make sense if you have a map visible on the page that shows all the different map markers, and you have to distinguish them so the user can easily find them. As we don't reproduce the map, we must rely on the browser / OS / app to launch a mapping app and show the requested location when the user clicks on the link. In this situation (no map), the link itself should either show a map pin: 
![image](https://user-images.githubusercontent.com/4304337/61812745-0401ac00-ae3c-11e9-818c-d9cfb01e180a.png)
or another character like 🌍 as @kelson42 suggests. The `1` can simply be replaced with this character or png, and then it's up to the reader software to interpret the vCard geo: link and launch the OS default mapping app centred on the given co-ordinates.

PR #957, opened 2019-08-27
:sparkles: replace kartographer maplink "1" with world emoji (#871)

ISNIT0:
#871


ISNIT0:
@kelson42 The problem was that the code specifically replaced only items with a "1", I've now changed the logic.
It was an attempt to not add the emoji to empty kartographer links which seem to be quite common

kelson42:
@ISNIT0 Thx, hope it works because I don't have tested again.

Issue #938, opened 2019-08-11
Handle properly articles list with redirect with local anchors


mgautierfr:
Article http://library.kiwix.org/wikipedia_en_medicine_nodet_2019-08/A/Retina and http://library.kiwix.org/wikipedia_en_medicine_nodet_2019-08/A/Lipemia_retinalis are very close.

They have the same title and the content is the same.
There are some difference in the html. (Especially a `<link href="../-/s/css_modules/mediawiki.action.view.redirectPage.css" rel="stylesheet" type="text/css" class="">` in `A/Lipemia_retinalis`.

kelson42:
@ISNIT0 "Lipemia retinalis" should be a redirect to "Retina"

kelson42:
@ISNIT0 The problem here is that we don't have a simple redirect from "Lipemia retinalis" should be a redirect to "Retina", we have a redirect to a specific paragraph"Retina#Diseases and disorders". This is not possible to do with the built-in ZIM redirect system. This should be done with a normal HTML page redirect.

ISNIT0:
@kelson42 Do you mean the problem is that the articleList contains redirects?

kelson42:
@ISNIT0 Yes... I confirm it is in https://ftp.nluug.nl/pub/kiwix/wp1/enwiki_2019-08/customs/medicine. You mean it might be a duplicate of  #889?

ISNIT0:
I think it is, yes

kelson42:
@ISNIT0 It is, but the problem is a bit more complex here because of the hash at the end of the URL... But I would be agree to handle it together in 2.0 if this is what you prefer. This is indeed really similar.

ISNIT0:
Surely that's fine? The reader serves the content from the redirect, and the hash is preserved?
I have a fix nearly ready

mgautierfr:
For information, zimwriterfs parse the content of the html to detect redirect and create a redirect article.
(See https://github.com/openzim/zimwriterfs/blob/master/src/article.cpp#L143-L159)

ISNIT0:
This issue should have been closed by the PR above. I will close for now, we can re-open if it's not fixed :)

kelson42:
@ISNIT0 Thx for the fix, I just reopen it, because I want to verify this by myself.

kelson42:
The problem here is how we create the list of articles for the medicine selection, the list is full of redirects.

PR #960, opened 2019-08-27
:sparkles: remove duplicate article data for redirects (#889)

### LINKED ISSUE #889: Article list redirects are not handled properly
MWoffliner can not deal properly with `--articleList` specified list of articles which contains redirects:
* The redirects target is not scraped
* There is no check that the targeted article is not already in the list.

The consequences are `libzim` generated errors like for example:
```
$ cat articles 
Enter_The_Pitch
Pinewood_Studios
$ mwoffliner --mwUrl=https://en.wikipedia.org/ --adminEmail=kelson@kiwix.org --verbose --articleList=articles
```

will generate this kind of errors:
```
...
[log] [2019-07-10T12:58:03.517Z] Writing Article Redirects
Impossible to add AEnter_The_Pitch
  dirent's title to add is : Enter The Pitch
  existing dirent's title is : Pinewood Studios
[log] [2019-07-10T12:58:03.522Z] Finishing Zim Creation
...
```

I propose to make the request for article details (only if we have a hardcoded list given) able to detect if this is a redirect. If this is a redirect, then find the target and add it to the list of articles to mirror (if not already in the list).

This is a follow up of #828

See also openzim/wp1_selection_tools#18

ISNIT0:
Closes #889 

Relates to #938

### DIFF
--- a/src/ZimCreatorFs.ts
+++ b/src/ZimCreatorFs.ts
@@ -20,7 +20,7 @@ class ZimCreatorFs extends ZimCreator {
         const articleFileName = path.join(this.fileName, article.aid);
 
         if (article.redirectUrl) {
-            const target = article.redirectUrl.split('/').slice(1).join('/'); // Hack
+            const target = article.redirectUrl;
             try {
                 const ret = await symlinkPromise(target, articleFileName);
                 return ret;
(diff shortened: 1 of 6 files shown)


ISNIT0:
@kelson42 This is a path issue with the auto-generated imports. I've fixed the issue

kelson42:
@ISNIT0 OK, hope it works because I don't have tested again.

Issue #948, opened 2019-08-22
WPEN is not scrapeable anymore in 16GB of RAM


kelson42:
See https://farm.openzim.org/tasks/5d5c519f1339d1137618f4b1
An additional effort should be done on memory consumption

kelson42:
@ISNIT0 Any idea how we could do better?

kelson42:
Here is the monitoring at that time
![image](https://user-images.githubusercontent.com/1029718/63546275-b7d48500-c529-11e9-8ca7-a0091ff6dd29.png)

kelson42:
@ISNIT0 I'm not so sure anymore mwoffliner has crashed because of lack of memory considering that we had still a lot of memory cached.

kelson42:
@ISNIT0 I have asked @rgaudin to restart a scrape, the error is quite weird to me. I want to see if the problem is reproducable.

ISNIT0:
@kelson42 Any news?

rgaudin:
Failed again in the favicon resize. https://farm.openzim.org/tasks/5d611a341339d1137618f565

ISNIT0:
Looks like it's related to this: https://github.com/nodejs/node/issues/25382

kelson42:
@ISNIT0 You told me that the problems comes probably from the fact that a process fork happens here at at time the Node.js process is already quite big. Maybe the solution is simply to move this image resize to the start of the scraping process?

ISNIT0:
It seems imagemin (our image compression library) also suffers from this problem, so we could potentially save even more memory by finding a node configuration variable. https://github.com/nodejs/node/issues/25382#issuecomment-497083280

The favicon is done at the end because favicon downloading etc. is handled the same way any other media file is downloaded.

ISNIT0:
This PR _may_ help https://github.com/openzim/mwoffliner/pull/968

kelson42:
@ISNIT0 Still same problem
```
T:91842; 16000/19032
T:91844; 17000/19032
T:91846; 18000/19032
T:91848; 19000/19032
T:91900;  write checksum
T:91900; finish
[log] [2019-09-09T15:10:53.780Z] Summary of scrape actions: {
	"files": {
		"success": 30889,
		"fail": 0
	},
	"articles": {
		"success": 5925375,
		"fail": 11
	},
	"redirects": {
		"written": 8408528
	}
}
[log] [2019-09-09T15:10:53.787Z] Finished dump
[...]

PR #968, opened 2019-09-02
:arrow_up: use execa instead of execPromise

ISNIT0:
Fix #968


kelson42:
@ISNIT0 Does that fix on of the pending issues?

ISNIT0:
@kelson42 It is a general improvement in the way we execute commands, but it _may_ improve #948

Issue #873, opened 2019-07-04
Wikivoyage DE homepage is not displayed properly


kelson42:
Here is how it looks like after a scrape with git master head:
![FireShot Capture 021 - Hauptseite - localhost](https://user-images.githubusercontent.com/1029718/60667336-96c8af80-9e69-11e9-9ef9-96089fa215f8.png)

To be compared to https://de.wikipedia.org/wiki/Wikipedia:Hauptseite

ISNIT0:
It seems the live page (https://de.wikivoyage.org/wiki/Hauptseite) has lots of inline styles in the article `<head>`.
This inline style isn't returned by the visual editor api: https://de.wikivoyage.org/w/api.php?action=visualeditor&mobileformat=html&format=json&paction=parse&page=Hauptseite

ISNIT0:
@kelson42 What's the sensible thing to do in this case?

kelson42:
@ISNIT0 I don't see anything special with that page in the wiki code. This is a standard Wikivoyage page. We should not consider this "inline style" as an exception, but as something which should be supported. It as working before, it should still work. Or do we have a technical problem here?

ISNIT0:
The api doesn't return the inlined styles, I'm not sure how else to get it...

kelson42:
@ISNIT0 Which nodes are we talking about exactly? Can you copy there please?

ISNIT0:
In the `<head>` element on this page: https://de.wikivoyage.org/wiki/Hauptseite
There are 7 `<style>` elements. Normally the css is exposed as modules, which we then have access to, but in this case we don't

kelson42:
@ISNIT0 I don't see any `<style>` element. Please copy them here.

ISNIT0:
Looks like these style elements aren't causing this specific issues (though might cause other issues).
The most important thing now is that the API returns no style modules: https://de.wikivoyage.org/w/api.php?action=parse&format=json&page=Hauptseite&prop=modules%7Cjsconfigvars

The specific module we need is `site.styles`, but this isn't returned for the article

ISNIT0:
I don't think this is a 1.9 issue

ISNIT0:
https://phabricator.wikimedia.org/T232092

PR #974, opened 2019-09-05
:art: add styles to css request (#873)

ISNIT0:
@kelson42 I'm not certain this is the correct fix, but it does seem to work without causing obvious issues in EN10 and BM scrapes


kelson42:
@ISNIT0 Good :)

PR #977, opened 2019-09-06
:see_no_evil: update .npmignore

ISNIT0:


kelson42:
`Changelog` should also be ignored

Issue #989, opened 2019-09-18
Sample custom plugin for wiktionary


rgaudin:
In order to produce better wiktionary ZIM ([#147](/openzim/zim-requests/issues/147) [#148](/openzim/zim-requests/issues/148)), we need to be able to manipulate the received DOM for each article.

This is why #941 was opened: to allow custom article DOM processing in mwoffliner through a plugin mechanism. Discussion about how to implement this mechanism is up to #941 so I am documenting here a sample requirement for the *demo wiktionary plugin* we'll use to bootstrap the feature.

### Main objective

* simplify [fr-wiktionary](https://fr.wiktionary.org) ZIM with french-only info.
* ideally it should work with all wiktionaries but for this use case, we'll test over FR version.
* the *target* language would be the main wiktionary language (FR in this case) but we might want to keep it an option to extract specific foreign-to-french language ZIM (BM-FR or IT-FR for instance).
* in this regard, we might want to set a list of languages to keep? @kelson42 @Popolechien what do you think?
  * In `fr`, we might need to include `fro` (Ancien français)…
* articles without a section of the target language(s) should not be included.
  * Ex: with `fr` target, [buru article](https://fr.wiktionary.org/wiki/buru) should not be included.
* the TOC should be removed if we have a single language target and reduced if not.
* only the target-language(s) section(s) should be kept.
* in-section content should be kept as is.

### DOM Manipulations

* [pain](https://fr.wiktionary.org/wiki/pain) is a good example
* we'll need a mapping for `targetLanguage(s)` to `#AnchorLink(s)` as this is what's usable (in the rendered version).
* `h1#firstHeading` is kept
* rest happens solely within `div.mw-parser-output`
* if we can know that `div.bandeau-voir` will not have valid links, remove it. (otherwise, it will be a link-less sentence). this classname being french, I bet it's specific.
* if `div#toc > ul li a[href]` (without `#`) is not in our `anchorLinks` then the whole `li` is removed.
* if there's only `div#toc > ul > li` left, then remove `div#toc`
* if there's no `div#toc > ul > li`, then the page doesn't contain wanted content and should not be included.
* for the rest of the article, if `h2 > span.mw-headline[id]` not in `anchorLinks` then every elements from that `h2` up to the next `h2` (or end of `div.mw-parser-output`)

Popolechien:
> In fr, we might need to include fro (Ancien français)…

[...]

PR #992, opened 2019-09-30
:sparkles: implement customProcessors for filtering and post/pre processing articles (#941)

### LINKED ISSUE #941: Allow custom flavour plugins
This is necessary to allow to make https://github.com/openzim/zim-requests/issues/148 properly

ISNIT0:
Closes #941 
Relates to #989 

This is not intending to fully implement #989, just to get it started.

`CustomProcessor`s must implement the `CustomProcessor` interface:
```typescript
interface CustomProcessor {
    shouldKeepArticle?: (articleId: string, doc: Document) => Promise<boolean>;
    preProcessArticle?: (articleId: string, doc: Document) => Promise<Document>;
    postProcessArticle?: (articleId: string, doc: Document) => Promise<Document>;
}
```

The `--customProcessor` argument takes a path to a JS file, e.g.: `--customProcessor=./extensions/wiktionary_fr.js`

Please confirm I've understood the specification correctly?
If so, I'll start writing tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -30,6 +30,7 @@
                 // "--articleList=https://ftp.nluug.nl/pub/kiwix/wp1/enwiki_2019-01/tops/100",
                 // "--articleList=https://download.kiwix.org/wp1/enwiki_2019-04/projects/Football",
                 "--redis=redis://127.0.0.1:6379",
+                // "--customProcessor=./extensions/wiktionary_fr.js"
                 // "--getCategories=true"
                 // "--customZimFavicon=test.png"
             ]
--- a/extensions/wiktionary_fr.js
+++ b/extensions/wiktionary_fr.js
@@ -1,0 +1,14 @@
+module.exports = class WiktionaryFR { // implements CustomProcessor
+    async shouldKeepArticle(articleId, doc) {
+        const frenchTitle = doc.querySelector(`#fr.sectionlangue`);
+        return !!frenchTitle;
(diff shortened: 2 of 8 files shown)


kelson42:
@ISNIT0 From the principle it looks good to me, but please call it `--customFlavour`

ISNIT0:
@rgaudin Tests added

kelson42:
> I have no problem with the processor name ; seems more appropriate than flavour to me but meh.

Why this is not a "Flavour" to your opinion?

rgaudin:
ATM flavour is used to indicate an “interpretation” of a source.

Currently we have a 1:1 relationship with formats to indicate flavours:
maxi flavour is novid format.

This processing can be paired with formats. Let’s say we add a processor
that removes offensive words from WP and we want that in both maxi, nopic
and mini ; what’s the flavour then?

I think flavour is too vague of a name and we should have rules somewhere
on the wiki to describe what’s the intended use for this classification.

kelson42:
> ATM flavour is used to indicate an “interpretation” of a source.

This is it, everything doing this interpretation and called differently is misnamed. `--format` should be probably renamed, this is a legacy stuff.

If we want to have a builtin badword filter, lets build it and just add a new `nobadword` keyword, but this is still everything about the flavour. `nobadword` is a way to specify a flavour which can be custom named or not.

rgaudin:
Ok so formats are just bundled flavours.

Issue #531, opened 2019-01-23
Should give a nicer error when we can't connect to Redis


ISNIT0:
See #530

kelson42:
@ISNIT0 Agree, no scenario should lead to an unmanaged stop/fail.

vaibhavmatta:
May I start working on this issue if anyone else is not?

kelson42:
@vaibhavmatta Yes, would be really welcome.

Fontirroig:
Hello, I like to help with this issue, I created a pull request

PR #1003, opened 2019-10-26
#531 Should give a nicer error when we can't connect to Redis

Fontirroig:


kelson42:
@Fontirroig Thank you for your PR. Does it fix/implement a specific ticket? #531 maybe?

PR #1012, opened 2019-10-29
single container for mwoffliner and redis

rgaudin:
**WARNING**: not intended to be merged-in as is.

* mwoffliner has a runtime dependency to redis.
* most people run mwoffliner via our docker image
* redis-server is a lightweight service (when idle) with close to no output
* most people using mwoffliner have to start a detached redis container for mwoffliner use and stop it afterwards.

For this last reason, and the fact that in my (zimfarm) use case, I have to write and maintain specific code to handle that two-containers situation (so far, all other scrapers are single-containers), I figured it would be handy to offer an all-inclusive image: **a docker image containing both a running redis and mwoffliner**.

This PR offers exactly this: it adds a new dockerfile (and entrypoint) to create a new _node-redis_ base image that is basically the `node:10` image with the redis binaries and an entrypoint to start it in the background.
Additionally, it changes our `mwoffliner` base image to this one.

Using this:
* regular user would just have to run mwoffliner container directly (bash alias adding `--redis redis.sock` automatically).
* advanced user would still be able to run a dedicated redis (specifying it with `--redis`).

**advantages**:

* simpler for most users.
* simpler resources monitoring/restriction of the _scraping job_.

**drawbacks**:

* not following docker dogma (but not preventing it)
* a non-used running redis-server in the background

--

Let me know what you think. maybe we could provide this in addition to the normal docker image on a separate tag but that looks like maintenance work with low benefits ; hence this version.

Haven't updated doc in the PR on purpose. Will do should we agree on doing this.

### DIFF
--- a/docker/.custom-bashrc
+++ b/docker/.custom-bashrc
@@ -1,2 +1,3 @@
 source ~/.old-bashrc
-alias mwoffliner='mwoffliner --redis="redis://redis"'
+alias mwoffliner='mwoffliner --redis=/dev/shm/redis.sock'
+alias redis-cli='redis-cli -s /dev/shm/redis.sock'
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,11 @@
-FROM node:10
+FROM openzim/node-redis
 
 # Install dependences
-RUN apt update && apt install -y --no-install-recommends make g++ curl git imagemagick
+RUN apt-get update && \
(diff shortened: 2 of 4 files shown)


rgaudin:
OK will update documentation then. I might submit a new PR with just he redis-node container, have it merged and image created on docker hub **then** submit the changes to our Dockerfile (as it depends on the other one existed).

kelson42:
@rgaudin Good to me.

PR #1029, opened 2020-01-28
Image Url Optimisation using Wasabi

### LINKED ISSUE #996: Get a better optimized content cache management
With the current MWoffliner and without `--useDownloadCache`, more than 2/3 of the whole process duration, MWoffliner optimised pictures, reducing the size of them to secure as smaller ZIM files as possible. This is too long, considering the amount hardware resources we have and the one release/month throughput we want to achieve.

If we use a download cache, things could be better because the cache keeps the optimised version of the pictures. But this has a few weaknesses:
* This can not be shared through different servers (because this is a filesystem based cache)
* There is a mix betweetn optimisation cache and download cache (this is semantically wrong and might bring further problems)
* The cache works also for text/pictures... and here we are mostly talking from pictures (caching the text might slow down the proces - to be confirmed).

Another point, it that this cache is obviously MWoffliner specific and we might need that kind of solution for all our other scrapers. So we probably need a better solution generic and network based.

bakshiutkarsha:
Fixes #996 
- Wasabi Plugin added
- Uploading new images to wasabi 
- Checking if the image exists, if yes, downloading it from wasabi and writing it to zim file.(Not done via Etag comparison because aws generates its own etag and it will never match.)
- If image doesn't exists, again following step 2.


src/util/const.ts
```diff
@@ -1,3 +1,9 @@
 
 export const MEDIA_REGEX = /^(.*\/)([^/]+)(\/)(\d+px-|)(.+?)(\.[A-Za-z0-9]{2,6}|)(\.[A-Za-z0-9]{2,6}|)$/;
 export const MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE = 10;
+export const WASABI_CONFIG = {
```
kelson42: Oooops, really bad to put credentials in git.

PR #1049, opened 2020-03-15
Internal wikis link broken fixed

### LINKED ISSUE #1017: [REGRESSION] URL rewritting completly broken with gbf.wiki and other wikis
With 1.9.12:
```
mwoffliner --mwUrl=https://gbf.wiki/ --adminEmail=kelson@kiwix.org --filenamePrefix=granbluefantasy_en_all --format=nopic --format=novid:maxi --mwWikiPath=index.php --mwModulePath=load.php --mwApiPath=api.php --useCache --verbose
```

All the internal links are broken.

bakshiutkarsha:
I have changed the end-point of the API for the wikis having mwApi path:
http://localhost:6927/${domain}/v1/page/mobile-sections/ **(Producing the wrong hrefs)**
to
http://localhost:8000/${{domain}}/v3/{+path}

I am not sure if every wiki has the same mwApi path like:
--mwApiPath = "api.php"

And if the wikis doesn't have same mwApiPath, on what other basis should we differentiate? @kelson42 
Should resolve #1017

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -235,6 +235,11 @@ class Downloader {
     });
     const domain = (urlParser.parse(this.mw.base)).host;
     this.mcsUrl = `http://localhost:6927/${domain}/v1/page/mobile-sections/`;
+
+    if(this.mw.apiPath == 'api.php'){
+      this.mcsUrl = `http://localhost:8000/${{domain}}/v3/{+path}`
+    }
+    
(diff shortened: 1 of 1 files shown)


kelson42:
@bakshiutkarsha This seems to be really wrong. You have changed the whole API used to get content! This is a URL rewriting problem, so you should fix the URL rewriting in a generic manner.

PR #1052, opened 2020-03-17
Internal Links Broken for some wikis

### LINKED ISSUE #1017: [REGRESSION] URL rewritting completly broken with gbf.wiki and other wikis

bakshiutkarsha:
Should resolve #1017

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -263,6 +263,11 @@ class Downloader {
     });
     const domain = (urlParser.parse(this.mw.base)).host;
     this.mcsUrl = `http://localhost:6927/${domain}/v1/page/mobile-sections/`;
+
+    if(this.mw.apiPath == 'api.php'){
+      this.mcsUrl = `http://localhost:8000/${{domain}}/v3/{+path}`
+    }
+    
     if (forceLocalParsoid) {
       const webUrlHost = urlParser.parse(this.mw.webUrl).host;
       this.parsoidFallbackUrl = `http://localhost:8000/${webUrlHost}/v3/page/pagebundle/`;
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -263,7 +263,6 @@ class Downloader {
     });
     const domain = (urlParser.parse(this.mw.base)).host;
     this.mcsUrl = `http://localhost:6927/${domain}/v1/page/mobile-sections/`;
-
     if (forceLocalParsoid) {
       const webUrlHost = urlParser.parse(this.mw.webUrl).host;
       this.parsoidFallbackUrl = `http://localhost:8000/${webUrlHost}/v3/page/pagebundle/`;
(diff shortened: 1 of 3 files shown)


bakshiutkarsha:
The problem lies in API response when we use parsoid.
Scenario 1: Where links are broken
- Atricle List: Damage_formula, Hades
Now Damage_Formula API response for Hades link is(only posting the relevant data here) :
<a href="/wiki/Hades" title="Hades" data-parsoid="{&quot;stx&quot;:&quot;simple&quot;,&quot;a&quot;:{&quot;href&quot;:&quot;./Hades&quot;},&quot;sa&quot;:{&quot;href&quot;:&quot;Hades&quot;},&quot;dsr&quot;:[6938,6947,2,2]}">Hades</a>.

and the linkNode getting created(again posting the relevant data):
_attrsByLName:
   [Object: null prototype] {
     '|href':
      { localName: 'href',
        prefix: null,
        namespaceURI: null,
        data: '/wiki/Hades',
        _ownerElement: [Circular],
        onchange: undefined },
     '|title':
      { localName: 'title',
        prefix: null,
        namespaceURI: null,
        data: 'Hades',
        _ownerElement: [Circular],
        onchange: undefined },
     '|data-parsoid':
      { localName: 'data-parsoid',
        prefix: null,
        namespaceURI: null,
        data:
         '{"stx":"simple","a":{"href":"./Hades"},"sa":{"href":"Hades"},"dsr":[6938,6947,2,2]}',
        _ownerElement: [Circular],
        onchange: undefined } },
  _attrKeys: [ '|href', '|title', '|data-parsoid' ]

here if you will notice the href is also "/wiki/Hades", but in data-parsoid data is "a":{"href":"./Hades"} which we want actually. But right now the code does not consider that data-parsoid href and goes straight to first href(you can see above), which is "/wiki/Hades".

Now here the question is:
1. How do we differentiate when to get the data from href and when from data-parsoid?
2. Is there we can know that the local server is creating the right hrefs or not?

bakshiutkarsha:
> IMO this can simply not be the right manner of fixing this bug. This should be fixed without hardcoding anything related to the URL.

Yes I know that and we have discussed this earlier as well, and that is why my above comment mentioned some queries which needs answers so that I can proceed in right direction.

kelson42:
@bakshiutkarsha Sorry I forgot about your message. But all this discussion should be in the ticket, not the PR. I will try to answer it but will close this PR.

I think for the moment we rewrite the pure HTML and we ignore the `data-parsoid` attribute. I see no reason why it should change.

I see also no argument to say that `/wiki/Hades` is a broken `href` value, it is just that we fail to rewrite it properly.

I think the most important question is: why we fail URL rewritting with a few wikis (because of absolute URLs?) and other work. Is the data returned by the Parsoid API not the same? If "yes" how and why?

PR #1066, opened 2020-04-01
Clearer download retry behaviour

kelson42:


src/util/saveArticles.ts
```diff
@@ -75,8 +75,8 @@ export async function downloadFiles(fileStore: FileStore, zimCreator: ZimCreator
         });
     });
 
-    if (!isRetry) {
-        await downloadFiles(filesToRetryXPath, zimCreator, dump, downloader, true);
+    if (retry_laster) {
```
bakshiutkarsha, line 78: typo error: should be "retry_later"

kelson42: fixed

PR #1082, opened 2020-04-22
Logs for getting total articles

### LINKED ISSUE #993: getArticlesByNS is not finding all articles on EN Wikipedia
EN Wikipedia scrapes are finding about half of the existing articles.

bakshiutkarsha:
In this PR, both the logs are included:
- sum of total articles found by namespace
- Length of articles stored in Redis

@kelson42 Let me know if we need both the logs, or either of them?

Should resolve #993

### DIFF
--- a/src/util/RedisKvs.ts
+++ b/src/util/RedisKvs.ts
@@ -157,7 +157,6 @@ export class RedisKvs<T> {
           reject(err);
         } else {
           resolve(val);
-          logger.log(`Total Articles stored in redis ${val}`);
         }
       });
     });
--- a/src/util/RedisKvs.ts
+++ b/src/util/RedisKvs.ts
@@ -1,6 +1,7 @@
 import {cpus} from 'os';
 import {mapLimit} from 'promiso';
 import type {RedisClient} from 'redis';
+import logger from '../Logger';
 
 interface ScanResult {
   cursor: string;
@@ -156,6 +157,7 @@ export class RedisKvs<T> {
           reject(err);
         } else {
           resolve(val);
+          logger.log(`Total Articles stored in redis ${val}`);
         }
       });
     });
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -98,13 +98,7 @@ export async function getArticlesByNS(ns: number, downloader: Downloader, contin
         totalArticles += numDetails;
 
         // Only for testing purposes
-        if (typeof continueLimit !== 'undefined') {
-            if (continueLimit > 0) {
-                continueLimit = continueLimit - 1;
-            } else {
-                break;
-            }
-        }
+        if (--(continueLimit as number) < 0) break;
     } while (_gapContinue);
 }
 
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -72,35 +72,40 @@ export async function getArticlesByIds(_articleIds: string[], downloader: Downlo
     );
 }
 
-export async function getArticlesByNS(ns: number, downloader: Downloader, _gapContinue?: string, continueLimit?: number): Promise<void> {
+export async function getArticlesByNS(ns: number, downloader: Downloader, continueLimit?: number): Promise<void> {
+    let _gapContinue: string;
+    do {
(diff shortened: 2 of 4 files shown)


src/util/RedisKvs.ts
```diff
@@ -156,6 +157,7 @@ export class RedisKvs<T> {
           reject(err);
         } else {
           resolve(val);
+          logger.log(`Total Articles stored in redis ${val}`);
```
midik, line 160: `RedisKvs.len()` is being used not only for articles but has number of usages across the code (for example, for the files), while the original issue is about articles (and also the message logged). I think would be better to move the logging out of this class and put to the outer level.
Another reason to do that would be to keep RedisKvs class low coupled with application logic as possible :)

kelson42:
@bakshiutkarsha @midik There is an other problem, none of this is tested and codecov decreased over the limit.

src/util/RedisKvs.ts
bakshiutkarsha: Done.

src/util/mw-api.ts
```diff
@@ -72,35 +72,40 @@ export async function getArticlesByIds(_articleIds: string[], downloader: Downlo
...
-    const canContinue = typeof continueLimit === 'undefined' || continueLimit > 0; // used for testing
+        logger.log(`Got [${numDetails}] articles from namespace [${ns}]`);
+        totalArticles += numDetails;
 
-    if (gapContinue && canContinue) {
-        const nextContinueLimit = typeof continueLimit === 'undefined' ? undefined : continueLimit - 1;
-        return getArticlesByNS(ns, downloader, gapContinue, nextContinueLimit);
-    }
+        // Only for testing purposes
+        if (typeof continueLimit !== 'undefined') {
```
midik, line 101: My personal opinion is that having some code only for testing purposes (outside the tests) would be considered as anti-pattern. However, I'd not mind about this in this particular case and I don't see another reasonable way to limit the number of iterations here.
Anyways, that would be shortened to the following one-liner:
```
if (--(continueLimit as number) < 0) break;
```
There's a [sandbox](https://www.typescriptlang.org/play/?module=1&ssl=25&ssc=52&pln=25&pc=9#code/DYUwLgBAdhC8EEYAMBuAsAKFJAlnCqmA9EdhAMYD2UYOUAriADI4C2OYAXNPawEYgATvgCs6DCTJUadRi3aR49KABMQAMzogV4zJnVQAFCICUujOuXla1CAcPTaDZmw4B+bg35CTEAN6YEEEQKpT+gcGR0gDOlKAAdMCUAOaGAAYcQhAAJH44AL5pZhGRQTgA1OXmRABUJaU46hCGYACeAA4glE2Osi4KEACEsPAA5MpqmlDao74BGKWLEI3Nvc7yHBAAfARz9Uula3KuihTUTscDALSI4geR+RAgwNEg4Qv3izFxIIkphqMQAAPDizO6fUp8QQgACGAGtwZ98vtgsiMDUiHoPosVoYrlcHOc+htIDDojxvIJfAAeXYQKGwhEooL1R4AdwAFjhQM08LSoMUMPVvgkkqlRqFpmDMPkgA) where I've tested this.

kelson42: Agree with that comment

bakshiutkarsha: Solved.

kelson42:
@bakshiutkarsha Can you please secure that codefactor does not report new problems? We should not have trailing spaces in our code.

bakshiutkarsha:
@kelson42 Locally lint was not giving this issue, so I think we should make the lint checks more rigorous. Meanwhile, I am fixing these.

kelson42:
> @kelson42 Locally lint is not giving these issues, so I think we should make the lint checks more rigorous. Meanwhile, I am fixing these.

Thx, agree with you. I recommend you just configure your code editor to just highlight these trailing spaces. It is a good practive anyway.

kelson42:
@bakshiutkarsha BTW, please avoid this kind of situation where you have two commits doing exactly the same with the same commit msg. This "pollutes" the revision log. Merge them.

Issue #1043, opened 2020-03-07
WPEN all scraping (and a few others) is stuck at the end of the process


kelson42:
```
[log] [2020-03-05T18:45:00.772Z] Worker [21] Processing batch of article ids [["SC_Eendracht_Aalst","+8 more +","Alexej_Rubcov"]]
[log] [2020-03-05T18:45:04.479Z] Worker [6] Processing batch of article ids [["Luňákovec","+8 more +","Mexická_rallye_2017"]]
[log] [2020-03-05T18:45:06.326Z] Progress downloading articles [446670/447110] [99.9%]
[log] [2020-03-05T18:45:07.095Z] Worker [8] Processing batch of article ids [["107_(rozcestník)","+8 more +","Římskokatolická_farnost_Mělník-Pšovka"]]
[log] [2020-03-05T18:45:07.162Z] Worker [0] Processing batch of article ids [["Jugoslávská_liga_ledního_hokeje_1968/69","+10 more +","Kanton_Honfleur"]]
[log] [2020-03-05T18:45:08.162Z] Worker [17] Processing batch of article ids [["Studio_6","+8 more +","Poitou-Charentes"]]
[log] [2020-03-05T18:45:08.294Z] Worker [5] Processing batch of article ids [["Octan_olovnatý","+9 more +","Mistrovství_Evropy_juniorů_v_ledním_hokeji_1989"]]
[log] [2020-03-05T18:45:11.203Z] Worker [23] Processing batch of article ids [["Erste_Bank_Open_2018_–_mužská_čtyřhra","+9 more +","O_životě_na_vysokých_školách_pražských_knihy_dvoje"]]
[log] [2020-03-05T18:45:11.695Z] Worker [19] Processing batch of article ids [["Provincie_v_Itálii","+8 more +","Seiko_Hašimotová"]]
[log] [2020-03-05T18:45:15.526Z] Worker [1] Processing batch of article ids [["Seznam_písní_Karla_Šípa","+9 more +","Kanton_Garches"]]
[log] [2020-03-05T18:45:17.217Z] Worker [16] Processing batch of article ids [["Atletika_na_Letních_olympijských_hrách_1968_–_100_metrů_ženy","+8 more +","Indžera"]]
[log] [2020-03-05T18:45:19.197Z] Worker [10] Processing batch of article ids [["Jan_Fischer_(ilustrátor)","+8 more +","Andijské_Kojsu"]]
[...]

PR #1084, opened 2020-04-22
Fix/1043/stucks reprod

### LINKED ISSUE #1079: MWoffliner 1.10.6 dying because of Mediawiki API timeouts
See for example https://farm.openzim.org/recipes/wikipedia_en_all_maxi

midik:
fixes #1079 
add debug logging that probably help to solve #1043

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -632,6 +632,9 @@ class Downloader {
   }
 
   private async claimRequest(): Promise<null> {
+    // @ts-ignore
+    logger.info(`[queue] RSS=${process.memoryUsage().rss / 1024 / 1024} / AH=${process._getActiveHandles().length} / AR=${process._getActiveRequests().length}`);
+
     if (this.activeRequests < this.maxActiveRequests) {
       this.activeRequests += 1;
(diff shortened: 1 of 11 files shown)


src/Downloader.ts
```diff
@@ -466,9 +467,13 @@ class Downloader {
           resolve(val);
         }
       });
-      call.retryIf((err: any) => err.response && err.response.status !== 404);
+      // todo pull this logic up to the class
```
bakshiutkarsha, line 470: As you have already mentioned it as Todo, I think it's better if we can create a ticket for this to note it.

bakshiutkarsha:
@midik the codecov report for the PR seems to decrease the coverage.

src/Downloader.ts
midik: You're right, however, that's just a kind of refactoring so I don't think it deserve the ticket. I redesigned that.

midik:
@bakshiutkarsha slightly stepped up the coverage (technically) by covering redisKvs.keys(). We'll rethink the testing strategy soon, once we get couple critical bugs fixed.

src/Downloader.ts
bakshiutkarsha: Great 👍

Issue #1058, opened 2020-03-20
Edutechwiki has not title at the top of its pages


kelson42:
See for example http://library.kiwix.org/edutechwiki_fr_all_maxi_2020-03/A/Instrumentation

danielkschneider:
If this can help (sorry I am not a developer: I maintain this wiki and do not modify anything in the code or the theme (that being said, only tiny little things if something breaks + MediaWiki:Common.css). But we do use extensions that wikipedia does not have. https://edutechwiki.unige.ch/fr/Sp%C3%A9cial:Version - greetings and thanks !

kelson42:
@danielkschneider Thank you for your comment. I'm convince this is a bug on our side, even if this is the first time I see it at all. In an attempt to keep a good overview of problems we have with Edutech Wiki scraping, I have created an "edutech" tag. Feel free to tag any already existing bug using it, if you are impacted as well.

midik:
@danielkschneider would you please consider to upgrade MediaWiki to the latest version? That will allow us to use REST API introduced in v.1.35 and will fix this issue.

kelson42:
@midik You mean Mediawiki or Parsoid?

midik:
@kelson42 
I mean Mediawiki. My bad, REST API introduced in 1.34 not 1.35 that I said above. 
https://www.mediawiki.org/wiki/API:REST_API#Enabling_the_REST_API_on_your_wiki

kelson42:
For the record, the problem is the following. With edutechwiki, the only way to get the HTML is the API `action=visualeditor`. This is not unusual, in fact this the normal way to do for non Wikimedia wikis which have Parsoid/VisualEditor installed. The problem is that this approach does not deliver the `<h1>` DOM not which at the top of each page contains the title of an article. This makes sense because this is not editable, so the VisualEditor does not need it. We face here an old bug in MWoffliner and which impact a lot of ZIM files.

Three solutions can be implemented:
1 - We stop using this, and we use a local Parsoid instead. This has proven to work but use a lot of CPU and can be fragile.
2 - We retrieve the information about the title do display using the Mediawiki API and insert manually a `<h1>` node. See https://www.mediawiki.org/wiki/API:Info for example.
3 - We insert the `<h1>` node an compute an approximation of the title based on the articleid (just replacing the underscores with spaces should be good enough).

I think we should really try to do the (2) smartly (to avoid overhead for all other wikis which don't need this)... in particular Wikipedia.

[...]

PR #1089, opened 2020-04-28
fix source select logic + inject header to the page

midik:
- Fix: fallback / source select logic and related
- Fix: title rendering for the sources scraped using `visualeditor` (#1058)
- Refactoring

### DIFF
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -73,10 +73,20 @@ export const renderArticle = async (json: any, articleId: string, dump: Dump, us
 };
 
 
+const injectHeader = (content: string, articleId: string): string => {
+    const doc = domino.createDocument(content);
+    const header = doc.createElement('h1');
+    header.appendChild(doc.createTextNode(articleId));
+    const target = doc.querySelector('body.mw-body-content');
+    target.insertAdjacentElement('afterbegin', header);
+    return doc.documentElement.outerHTML;
+};
+
+
 const renderDesktopArticle = (json: any, articleId: string): string => {
     if (!json) { throw new Error(`Cannot render [${json}] into an article`); }
     if (json.visualeditor) {
-        return json.visualeditor.content;
+        return injectHeader(json.visualeditor.content, articleId);
     } else if (json.contentmodel === 'wikitext' || (json.html && json.html.body)) {
         return json.html.body;
     } else if (json.parse && json.parse.text) {
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -18,7 +18,7 @@ export const renderArticle = async (json: any, articleId: string, dump: Dump, us
     const articleDetail = await articleDetailXId.get(articleId);
 
     if (useParsoidFallback) {
-        const html = renderDesktopArticle(json, articleId);
+        const html = renderDesktopArticle(json, articleId, articleDetail);
         const strippedTitle = getStrippedTitleFromHtml(html);
         return [{
             articleId,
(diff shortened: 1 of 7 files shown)


src/util/articleRenderers.ts
```diff
@@ -109,9 +191,10 @@ export function renderMCSArticle(json: any, dump: Dump, articleId: string, artic
     }
     html = html.replace(`__SUB_LEVEL_SECTION_${json.remaining.sections.length}__`, ''); // remove the last subcestion anchor (all other anchor are removed in the forEach)
     return html;
-}
+};
+
 
-function groupAlphabetical(items: Array<{ name: string, url: string }>) {
+const groupAlphabetical = (items: Array<{ name: string, url: string }>) => {
```
bakshiutkarsha, line 197: In this PR, there are some functions which are changed from normal function to arrow functions. Is there a particular reason why it is done? IMK, simple functions are faster, takes less memory and are good for optimisation.

midik: My only reason is code readability.

Regarding to performance: those days, while arrow functions implementation in v8 was on early experimental stages, they could be significantly slower. Nowadays, it's not that simple. 

Anyways, I tested our own case with the following [benchmark](https://www.typescriptlang.org/play/#code/MYewdgzgLgBGMF4YEYAM7UG4CwAoPoksAJgIZSmIwDaeMMA3mKQLYCmAXDAOQBmIIAEzcANDACuAJwA2XbgAsoUAA7cAviLqNm7OfxDJREmXMUqI6zbnpNWnHvoDMRqbJ7jiqjVtu6eAI1JJNBcTHihgLysbHXtuQMkAVlC3bgBLYBYovABdHHwCgHpCmCDJEAB3AnBoGF4wAEFJcoqqAAo0qDYWCC4myVIATwAeBjg7LmhJNLAAczFXSahpuZg1AD4ASkR1xi1CWtny8WUIBulleVJ-NigM0mkqTu6IADpJNmJxYDY2ttJgMAuKQwIMxM8WNsELsGFp6AdYEcQCcAJLEJ5dFivWLUVA5V5QEAAGRAwAebAAqsplGxJABhUgQX6bfL0egA4DUJGo4g5docrnHZRovkAH1FNBym1ehDJUA6mJZcJgHygUngHNZazEDDUSoK9FV6pgAHl-AArNjAKCvADWbEGEDa3NO50u11u92km2VrwgIEk8p91jZrxYpGUf2kty6kihMOVhtuxthIbZbLuUGjXGjSlp0XTbIhvRgLrOFyuNzuZOk1FzsZyBfTai19D1+RbeDwxTgAfD0jwvHEYGtaXAdTAADk+w8FS8+s0hqNxn4pjN5sY3GvVhttqn4TVEULy+6q16MS93p9vr9-oCOCCwTAIfG9mmYAjS0K0ResTi8QSxKkuSVI0vSjLMsqAouiK-KAoKyLCryMDipK0qyuQc6QsqRqSBqgJaBojB6l2Ia4fAZqWtadoOk6ZZupWno1sG6Z+gGQa+uGkZtPWtKvvuhbkW+hYZp02YwLxkhNoWxZcPRFYetWDx1jGtKNomazKiRuBqKRPZgDOjwVAM1K0tURC9pI-btPx+yHjAAAeVAALLkPI7wgsQIAsG0+oHpAIDRgSaTsG03AGVZDzcH5dQBjAPG3M+VBYElwxwJgz4ANSZXuyr1NOkXSG0ZAUDFukhgcgVsMF7AAKJgMQYURf20Udvk3YlGUlQwMZEZgeZtRda0SC+TswkfvZTlIK5UDuQMDXeb5WqVUFdyhdwQ2tVo-CSPFuZJUgKVpDAaVgBlaTZbl771P0lTFeQpBlXZAWrSFbD1Y1G3NJUW06e1uA9pIw6DnFCWwMdh3nSdKBHZd43NbOMVDUteDlUAA).

Here is some results with 5 elements and 10000 runs:

Normal first
image

Arrow first
image
As we could see, the variance is within the margin of error in both direction (differences between first two runs and the following ones is V8 optimizations).

Another point of view from V8 developer: https://stackoverflow.com/a/44031830/11372576

bakshiutkarsha: Thanks for making the comparison and I totally agree that it improves the readability. These are sources I was referring to:
https://www.reddit.com/r/javascript/comments/9f3hhf/bound_function_vs_arrow_function_performance/

But I am happy that in our use-case it is not impacting the performance. :)

Issue #1095, opened 2020-05-02
Urls with `:` are not properly encoded


kelson42:
In http://library.kiwix.org/edutechwiki_fr_all_maxi/A/Accueil, we have this link `<a href="Guidelines:Règles_d'édition" title="Guidelines:Règles d'édition"> règles d'édition</a>` which is a broken link, because the character `:` is not URL encoded... and interpreted as protocol scheme separator.

bakshiutkarsha:
@kelson42 I highly doubt there is any problem with ":", I am able to scape the articles with ":", successfully. I tried with this article as well:
The Combat: Woman Pleading for the Vanquished.

If you just inspect and "open in new tab" from that href, the link is working perfectly its not broken:
http://library.kiwix.org/edutechwiki_fr_all_maxi/A/Guidelines:R%C3%A8gles_d'%C3%A9dition

But, when you will hover on that text on browser you will see "guidelines:Règles_d'édition" at the left corner of your screen(indicates the link)which starts from small "g", and that's the problem. It should have started from capital "G". I am trying to figure out the cause behind it.

kelson42:
@bakshiutkarsha You are wrong here. On way to immediatly see is to replace `G` by `g` in the DOM and you still see then that the URL is still broken.

bakshiutkarsha:
I don't think my comment was able to explain what I am trying to say. The href is correct in DOM, it is with capital G.
`g` is wrong anyways, so we don't have to replace it with `G`

kelson42:
@bakshiutkarsha If the URL is correct how do you explain that it does not open the right article if we click on it?

bakshiutkarsha:
<img width="827" alt="Screenshot 2020-05-07 at 3 01 45 AM" src="https://user-images.githubusercontent.com/13726478/81264403-56990700-900f-11ea-970f-8bcdde76880d.png">

WIth this screenshot, if you see on left corner, it highlights the link with small "g".

kelson42:
@bakshiutkarsha This is not an explanation where is the problem, this is a screenshot. Do you try to tell me the browser is buggy?

bakshiutkarsha:
This is the screenshot explains one possible reason of this problem.

As I mentioned in my previous comment, I am trying to figure out the cause of why it is happening(could be the buggy browser or something else), but I am pretty sure it isn't happening because of  ":"  not encoded.

kelson42:
[...]

PR #498, opened 2019-01-10
Fixes for zim.tmpDirectory and remote articleList

ISNIT0:
Closes: #488

PR #1096, opened 2020-05-04
Fix for the articles with question mark

### LINKED ISSUE #1024: MWoffliner does not scrap articles with question mark at end of title
See for instance [Have I been pwned?](http://library.kiwix.org/wikipedia_en_all_maxi_2018-10/I/Have_I_Been_Pwned%3F.html) or [Que faire?](http://library.kiwix.org/wikipedia_fr_all_maxi_2019-09/A/Que_faire_?)
The both appear in their respective indexes, though, and the error is seemingly limited to titles with a question mark at the end: [Saint-Louis du Ha! Ha!](http://library.kiwix.org/wikipedia_fr_all_maxi_2019-09/A/Saint-Louis-du-Ha!_Ha!) is ok, as well as [...and then there were three...](http://library.kiwix.org/wikipedia_en_all_maxi_2018-10/A/...And_Then_There_Were_Three....html)

bakshiutkarsha:
should resolve #1024 

Tested with following articles:
Que faire ?
Que faire ? (Lénine)
Avanti!
and some regular articles.

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -611,7 +611,7 @@ async function execute(argv: any) {
     }
 
     function createMainPageRedirect() {
-      logger.log(`Create main page redirection from [index] to[${'A/' + mainPage}]`);
+      logger.log(`Create main page redirection from [index] to [${'A/' + mainPage}]`);
       const article = new ZimArticle({
         url: 'index',
         shouldIndex: true,
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -22,7 +22,7 @@ import { config } from './config';
 import Downloader from './Downloader';
 import MediaWiki from './MediaWiki';
 import Redis from './Redis';
-import { writeFilePromise, mkdirPromise, isValidEmail, genHeaderCSSLink, genHeaderScript, genCanonicalLink, saveStaticFiles, readFilePromise, makeArticleImageTile, makeArticleListItem, getDumps, getMediaBase, MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE, downloadAndSaveModule, getSizeFromUrl, getRelativeFilePath, sanitizeString } from './util';
+import { writeFilePromise, mkdirPromise, isValidEmail, genHeaderCSSLink, genHeaderScript, genCanonicalLink, saveStaticFiles, readFilePromise, makeArticleImageTile, makeArticleListItem, getDumps, getMediaBase, MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE, downloadAndSaveModule, getSizeFromUrl, getRelativeFilePath, sanitizeString, getArticleBase} from './util';
 import { mapLimit } from 'promiso';
 import logger from './Logger';
 import { getAndProcessStylesheets } from './util';
@@ -380,7 +380,7 @@ async function execute(argv: any) {
     const zimCreator = new ZimCreator({
       fileName: outZim,
       fullTextIndexLanguage: dump.opts.withoutZimFullTextIndex ? '' : dump.mwMetaData.langIso3,
-      welcome: (dump.opts.mainPage ? dump.getArticleBase(dump.opts.mainPage) : 'index'),
+      welcome: (dump.opts.mainPage ? getArticleBase(dump.opts.mainPage) : 'index'),
     }, {
         Tags: dump.computeZimTags(),
         Language: dump.mwMetaData.langIso3,
@@ -612,7 +612,7 @@ async function execute(argv: any) {
     }
 
     function createMainPageRedirect() {
-      logger.log(`Create main page redirection from [index] to[${'A/' + dump.getArticleBase(mainPage, true)}]`);
+      logger.log(`Create main page redirection from [index] to[${'A/' + encodeURIComponent(mainPage)}]`);
       const article = new ZimArticle({
         url: 'index',
         shouldIndex: true,
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -22,7 +22,7 @@ import { config } from './config';
 import Downloader from './Downloader';
 import MediaWiki from './MediaWiki';
 import Redis from './Redis';
-import { writeFilePromise, mkdirPromise, isValidEmail, genHeaderCSSLink, genHeaderScript, genCanonicalLink, saveStaticFiles, readFilePromise, makeArticleImageTile, makeArticleListItem, getDumps, getMediaBase, MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE, downloadAndSaveModule, getSizeFromUrl, getRelativeFilePath, sanitizeString, getArticleBase} from './util';
+import { writeFilePromise, mkdirPromise, isValidEmail, genHeaderCSSLink, genHeaderScript, genCanonicalLink, saveStaticFiles, readFilePromise, makeArticleImageTile, makeArticleListItem, getDumps, getMediaBase, MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE, downloadAndSaveModule, getSizeFromUrl, getRelativeFilePath, sanitizeString } from './util';
 import { mapLimit } from 'promiso';
 import logger from './Logger';
 import { getAndProcessStylesheets } from './util';
@@ -376,11 +376,10 @@ async function execute(argv: any) {
     logger.log(`Flushing Redis file store`);
     await filesToDownloadXPath.flush();
     await filesToRetryXPath.flush();
-
     const zimCreator = new ZimCreator({
       fileName: outZim,
       fullTextIndexLanguage: dump.opts.withoutZimFullTextIndex ? '' : dump.mwMetaData.langIso3,
-      welcome: (dump.opts.mainPage ? getArticleBase(dump.opts.mainPage) : 'index'),
+      welcome: (dump.opts.mainPage ? dump.opts.mainPage : 'index'),
     }, {
         Tags: dump.computeZimTags(),
         Language: dump.mwMetaData.langIso3,
@@ -612,7 +611,7 @@ async function execute(argv: any) {
     }
 
     function createMainPageRedirect() {
-      logger.log(`Create main page redirection from [index] to[${'A/' + encodeURIComponent(mainPage)}]`);
+      logger.log(`Create main page redirection from [index] to[${'A/' + mainPage}]`);
       const article = new ZimArticle({
         url: 'index',
         shouldIndex: true,
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -347,3 +347,11 @@ export function objToQueryString(obj: KVS<any>): string {
 export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ');
 }
+
+export function getArticleBase(articleId: string) {
+    if (typeof articleId === 'undefined') {
+        return undefined;
+    }
+    // Encoding ? char because it is not allowed directly in hrefs.
+    return articleId.replace('?', encodeURIComponent('?'));
+}
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -348,10 +348,7 @@ export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ');
 }
 
-export function getArticleBase(articleId: string) {
-    if (typeof articleId === 'undefined') {
-        return undefined;
-    }
-    // Encoding ? char because it is not allowed directly in hrefs.
-    return articleId.replace('?', encodeURIComponent('?'));
+export function encodeArticleId(articleId: string) {
+  // Encoding ? and # from the articleId to prevent path termination in browser.
+  return articleId.replace('?', encodeURIComponent('?')).replace('#', encodeURIComponent('#'));
 }
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -349,6 +349,6 @@ export function sanitizeString(str: string) {
 }
 
 export function encodeArticleId(articleId: string) {
-  // Encoding ? and # from the articleId to prevent path termination in browser.
-  return articleId.replace('?', encodeURIComponent('?')).replace('#', encodeURIComponent('#'));
+  // Encoding articleId string and then decoding / char from it.
+  return encodeURIComponent(articleId).replace(/%2F/g, '/');
 }
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -351,5 +351,5 @@ export function sanitizeString(str: string) {
 // We will need the encoded URL on article load so that we can set the hrefs of anchor tag correctly,
 // but we must not encode the '/' character or else relative links may fail
 export function encodeArticleIdForZimHtmlUrl(articleId: string) {
-  return encodeURIComponent(articleId).replace(/%2F/g, '/');
+  return articleId && encodeURIComponent(articleId).replace(/%2F/g, '/');
 }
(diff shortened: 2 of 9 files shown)


bakshiutkarsha:
> everything is ok except of codecov complains (it seems strange)

Yeah, even I am not sure why codecov is complaining, I wrote decent amount of test cases too :(

kelson42:
@bakshiutkarsha @midik The codecov "miscomputation" shoudl be understood. I suspect a misconfiguration on our side.

kelson42:
@bakshiutkarsha A few remarks: 
* I don't really understand why we have a special treatment here for `?`. Yes `?` is a URL reserved character but there is others and I don't get why we should have a special treatment for that one in particular. Do you know why?
* We should not make a special automated test for '?' we should make a test for other/all special characters.
* Why we have a function `e()` considering this used only one time? We should probably remove it?
* `getArticleBase()` second parameter is call `escape`... but this is not escaping, this is URL encoding.
* `getArticleBase()` is called only one time with `escape` true AFAIK, we should ask ourself if this is proper to do this within the method. Probably not.
* We should remove the part deal with the 250 character limit of the filesystem. We use node-libzim since a year now, no reason to do that anymore
* Last but not least: why we have this method at all... why the `articleId` can not be used straight? We should have least put a comment IMO to explain why this method is needed at all.

bakshiutkarsha:
@kelson42  I have done the requested changes and for this question specifically, 

> I don't really understand why we have a special treatment here for ?. Yes ? is a URL reserved character but there is others and I don't get why we should have a special treatment for that one in particular. Do you know why?

The answer is the URL path is terminated by the question mark ("?"), which means after the question mark it won't take the rest of path. 
Eg: For **Que faire ? (Lénine)** article, it will only take show "Que_faire_" in path, that is the reason we are specifically encoding question mark in href, so that it becomes "Que_faire_%3F_(Lénine)".

For more info, this can help:
https://tools.ietf.org/html/rfc3986#section-3.3

src/mwoffliner.lib.ts
```diff
@@ -380,7 +380,7 @@ async function execute(argv: any) {
     const zimCreator = new ZimCreator({
       fileName: outZim,
       fullTextIndexLanguage: dump.opts.withoutZimFullTextIndex ? '' : dump.mwMetaData.langIso3,
-      welcome: (dump.opts.mainPage ? dump.getArticleBase(dump.opts.mainPage) : 'index'),
+      welcome: (dump.opts.mainPage ? getArticleBase(dump.opts.mainPage) : 'index'),
```
kelson42, line 383: I see no reason why we should have anything URL encoded here.

src/mwoffliner.lib.ts
```diff
@@ -612,7 +612,7 @@ async function execute(argv: any) {
     }
 
     function createMainPageRedirect() {
-      logger.log(`Create main page redirection from [index] to[${'A/' + dump.getArticleBase(mainPage, true)}]`);
+      logger.log(`Create main page redirection from [index] to[${'A/' + encodeURIComponent(mainPage)}]`);
```
kelson42, line 615: Same here? I don't understand why the URL in the ZIM should have any encoding.

src/util/misc.ts
```diff
@@ -347,3 +347,11 @@ export function objToQueryString(obj: KVS<any>): string {
 export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ');
 }
+
+export function getArticleBase(articleId: string) {
+    if (typeof articleId === 'undefined') {
```
kelson42, line 352: What is the legit scenario to get `undefined` here?

src/util/misc.ts
```diff
@@ -347,3 +347,11 @@ export function objToQueryString(obj: KVS<any>): string {
 export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ');
 }
+
+export function getArticleBase(articleId: string) {
```
kelson42, line 351: I would like to see here a comment explaining what does this function exactly and probably a better name as well.

src/util/misc.ts
```diff
@@ -347,3 +347,11 @@ export function objToQueryString(obj: KVS<any>): string {
...
+    if (typeof articleId === 'undefined') {
+        return undefined;
+    }
+    // Encoding ? char because it is not allowed directly in hrefs.
+    return articleId.replace('?', encodeURIComponent('?'));
```
kelson42, line 356: I still believe this is wrong. I believe this function makes the unencoded article_id ready for a URL, so for example `C♯_(musical_note)` shoudl be probably changed to `C%E2%99%AF_(musical_note)`!?

kelson42:
@bakshiutkarsha @midik Any reason to not just put `encodeURIComponent(articleId)` where this is needed adn remove the whole `getArticleBase()`?  What would fail if we do that?

Jaifroid:
@kelson42 Just remember the long discussions about not encoding forward slashes `/` in URLs as they are meaningful for ZIMs with subdirectories (and even a Wikipedia ZIM treats titles with forward slash as if they contained a subdirectory, which is what readers now assume). So forward slash must not be encoded.

kelson42:
@Jaifroid You enlight me. `encodeURIComponent(articleId)` is definitly a bad idea if we use for URLs in the ZIM. I will let @bakshiutkarsha clarify the purpose/usage/name of this function to continue the discussion.

kelson42:
@bakshiutkarsha I have open recently an other ticket #1095 which seems really related to this one.

To extend a bit what I have written earlier today, here are the questions I have regarding this method:
* Is that use to generate only URLs?
* In this URLs, the result is used in the URL body or in a variable value?
* Do we use it to make requests to the MW? Do we use is only to prepare URL which will be stored in the ZIM file? both?
* Would a better name be `encodeArticleId()`?

src/mwoffliner.lib.ts
bakshiutkarsha: Yes, even I can't see any good reason for this to be encoded.
Done.

src/mwoffliner.lib.ts
bakshiutkarsha: Removed.
Anyways it's just in log, seems sense to remove this.

src/util/misc.ts
bakshiutkarsha: The function name is changed to encodeArticleId(), which seems more explanatory and mentioned a comment which explains this function only seems to encode "?" and "#" characters.

src/util/misc.ts
bakshiutkarsha: No, it is not required. That's not a "#" character if you see it closely.

First of all, I tried to create the page with # character on wiki, I couldn't. It just wouldn't allow me and trim everything after the hash.

Secondly, if the case appears I have handled it by encoding "#" char as well. But in my opinion, no article id would have "#" in it.

bakshiutkarsha:
> @bakshiutkarsha I have open recently an other ticket #1095 which seems really related to this one.
I have commented on that ticket presenting my findings about that issue.
> 
> To extend a bit what I have written earlier today, here are the questions I have regarding this method:
> 
> * Is that use to generate only URLs?
Yes, only in href's in anchor tag.
> * In this URLs, the result is used in the URL body or in a variable value?
It is directly used in URL body.
> * Do we use it to make requests to the MW? Do we use is only to prepare URL which will be stored in the ZIM file? both?
We use it to only prepare and store url in zim.
> * Would a better name be `encodeArticleId()`?
Yes

src/mwoffliner.lib.ts
```diff
@@ -612,7 +611,7 @@ async function execute(argv: any) {
     }
 
     function createMainPageRedirect() {
-      logger.log(`Create main page redirection from [index] to[${'A/' + encodeURIComponent(mainPage)}]`);
+      logger.log(`Create main page redirection from [index] to[${'A/' + mainPage}]`);
```
kelson42, line 614: A space is missing here before the `[`

test/unit/util.test.ts
```diff
@@ -32,21 +32,23 @@ test('util -> Throttle', async (t) => {
...
-    const checkAddEqualChar = getArticleBase('2_+_2_=_5');
+    const checkExclamationChar = encodeArticleId('Avanti!');
+    const checkAndChar = encodeArticleId('McCormick_Tribune_Plaza_&_Ice Rink');
+    const checkAddEqualChar = encodeArticleId('2_+_2_=_5');
+    const checkMixChar = encodeArticleId('Saint-Louis-du-Ha!_Ha!');
 
     t.equal(checkExclamationChar, 'Avanti!', 'Not esacping ! char');
     t.equal(checkAndChar, 'McCormick_Tribune_Plaza_&_Ice Rink', 'Not escaping & char');
     t.equal(checkAddEqualChar, '2_+_2_=_5', 'Not escaping + and = char');
+    t.equal(checkMixChar, 'Saint-Louis-du-Ha!_Ha!', 'Not escaping mix char');
```
kelson42, line 53: I think I miss a test for the `#` character!

kelson42:
@bakshiutkarsha Thank you for the explanation and improvments. All of this go in the right direction but I believe the work is not over. Like we see `:` needs to be encoded as well. IMO you need to think twice about which character should be encoded and which not and for what reason? I'm not sure, but it might be here we need a `encodeURIComponent` for the whole string but excepting for the `/` (which handling is different within the ZIM than online). @Jaifroid Maybe you have a wise advise here as well?

Jaifroid:
@kelson42 @bakshiutkarsha I think I need to know what the aim is here. The ZIMs store article URLs (in the directory entry) without any percent encoding at all, and that still needs to be the case, but if we're talking about (relative) html hyperlinks inside an article, then they do need to be percent encoded other than forward slashes.

One option for the latter would be to use encodeURIComponent, and then re-convert "%2F" back into a forward slash?

bakshiutkarsha:
> @bakshiutkarsha Thank you for the explanation and improvrments. All of this goes in the right direction but I believe the work is not over. Like we see `:` needs to be encoded as well. IMO you need to think twice about which character should be encoded and which not and for what reason? I'm not sure, but it might be here we need a `encodeURIComponent` for the whole string but excepting for the `/` (which handling is different within the ZIM than online). @Jaifroid Maybe you have a wise advise here as well?

@kelson42 For now, I have encoded these characters:
1. `#`
2. `?`
3. `:`

And I am going to see what all characters can affect the url, will post in this discussion.

test/unit/util.test.ts
bakshiutkarsha: Done

src/mwoffliner.lib.ts
bakshiutkarsha: Done

kelson42:
> @kelson42 @bakshiutkarsha I think I need to know what the aim is here. The ZIMs store article URLs (in the directory entry) without any percent encoding at all, and that still needs to be the case, but if we're talking about (relative) html hyperlinks inside an article, then they do need to be percent encoded other than forward slashes.

Yes we are talking about the URLs within the HTML within the ZIM. The URLs of the directory stays unencoded.
 
> One option for the latter would be to use encodeURIComponent, and then re-convert "%2F" back into a forward slash?

This seems to be the best I can think about for the moment as well.

Jaifroid:
@bakshiutkarsha @kelson42 I've just checked the code we use for this in Kiwix JS reader. It's like this:
```
// We will need the encoded URL on article load so that we can set the iframe's src correctly,
// but we must not encode the '/' character or else relative links may fail [kiwix-js #498]
var encodedUrl = dirEntry.url.replace(/[^/]+/g, function (matchedSubstring) {
    return encodeURIComponent(matchedSubstring);
});
```

bakshiutkarsha:
> @bakshiutkarsha @kelson42 I've just checked the code we use for this in Kiwix JS reader. It's like this:
> 
> ```
> // We will need the encoded URL on article load so that we can set the iframe's src correctly,
> // but we must not encode the '/' character or else relative links may fail [kiwix-js #498]
> var encodedUrl = dirEntry.url.replace(/[^/]+/g, function (matchedSubstring) {
>     return encodeURIComponent(matchedSubstring);
> });
> ```

@Jaifroid Thanks for the help. Now, I have done the same in typescript.

kelson42:
@bakshiutkarsha We still need:
* codecov to stop complaining
* A clear comment on the top of the function to explain its purpose and limitations
* The name of the function is still a bit unclear to me. I would prefer something more explicit like `encodeArticleIdForZimHtmlUrl()`

src/util/rewriteUrls.ts
```diff
@@ -1,6 +1,6 @@
 import * as urlParser from 'url';
 import { isMirrored } from './saveArticles';
-import { migrateChildren, getMediaBase, getFullUrl, getRelativeFilePath } from './misc';
+import { migrateChildren, getMediaBase, getFullUrl, getRelativeFilePath, encodeArticleId} from './misc';
```
kelson42, line 3: A space is missing here after `encodeArticleId`.

test/unit/util.test.ts
```diff
@@ -30,3 +30,29 @@ test('util -> Throttle', async (t) => {
     t.equals(sanitizeString('SELECT * FROM db WHERE something="Poler"'), 'SELECT   FROM db WHERE something  Poler ', 'Escaping query characters');
 
 });
+
+test('Question Mark escape', async(t) => {
```
kelson42, line 34: Test name is not correct anymore. The test name should be directly linkable to what is tested, so to the function which is tested.

kelson42: I don't understand here why we have two test for one single function. Does not make sense IMO.

src/util/rewriteUrls.ts
bakshiutkarsha: Done

test/unit/util.test.ts
bakshiutkarsha: Resolved.

src/util/misc.ts
```diff
@@ -347,3 +347,9 @@ export function objToQueryString(obj: KVS<any>): string {
 export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ');
 }
+
+// We will need the encoded URL on article load so that we can set the hrefs of anchor tag correctly,
+// but we must not encode the '/' character or else relative links may fail
+export function encodeArticleIdForZimHtmlUrl(articleId: string) {
+  return encodeURIComponent(articleId).replace(/%2F/g, '/');
```
kelson42, line 354: What about:
```
return articleId && encodeURIComponent(articleId).replace(/%2F/g, '/');
```
That way we would have a better handling of the `undefined` use case. Returning `'undefined'` as string is really hiddeous.

test/unit/util.test.ts
```diff
@@ -30,3 +31,27 @@ test('util -> Throttle', async (t) => {
     t.equals(sanitizeString('SELECT * FROM db WHERE something="Poler"'), 'SELECT   FROM db WHERE something  Poler ', 'Escaping query characters');
 
 });
+
+test('Encoding ArticleId for Zim html Url', async(t) => {
```
kelson42, line 35: The way the test is build is not really nice because:
* It repeat a lot of time the same piece of code
* It put the input value, output value and comment in difference places

A better way to do so is creating an array with all the values an iterate through it.

test/unit/util.test.ts
```diff
@@ -1,8 +1,9 @@
 import './bootstrap.test';
 
 import test from 'blue-tape';
-import { throttle, sanitizeString } from 'src/util';
+import { throttle, sanitizeString, encodeArticleIdForZimHtmlUrl } from 'src/util';
 import { sleep } from 'test/util';
+import logger from '../../src/Logger';
```
kelson42, line 6: We shoudl not have `../..`

test/unit/util.test.ts
kelson42: The right way to write HTML is HTML (https://en.wikipedia.org/wiki/HTML) not `html`. It is a free string, we are free to write it properly.

PR #1097, opened 2020-05-07
move to node-libzim 2.2 + change build environments

midik:
Fixes  #707
@kelson42 I added node v12 build env and removed old ones. Please correct me if we still need for them.

PR #1104, opened 2020-05-08
Better imagemagick handling + refactoring

### LINKED ISSUE #1069: Saving favicon fails without imagemagick installed
Saving `favicon` without `imagemagick` installed fails with non-relevant error:

image

I think would be great to either add `imagemagick` to Readme/Prerequisites or/and make the error meaningful.

midik:
Fixes #1069


kelson42:
@midi Thank you for the PR, but IMO this is incomplete. All the deps which can not be controlled via `npm` should be checked at the start of the process.

kelson42:
@midik I wonder here why we don't use a builtin feature (without relying on the filesystem and an external binary).

midik:
> This was done in the past

@kelson42 What do you mean here? If that was done in the past, why we have no check today?

> why we don't use a builtin feature

I guess it's just because historical reasons

> Why not using sharp?

in a couple of previous projects, my personal experience with sharp was painful. However, there's no objective reasons to not to give it another chance. IMO better to move to sharp once we settled down [with node 12](https://github.com/openzim/mwoffliner/pull/1097).

kelson42:
@midik We had a check if exterinal binaries where here at the start of MWoffliner. Even the version was checked if I properly remember. I think it has been removed around 1.9.0 release because we have started to use binding for `pngquant`, `jpegoptim`, ... but this was a bit abusive because we still have dependence to external `convert`. Considering that we will move to nodejs12 within the next hours, I'm in favour of using `sharp` (or any other well maintained image library) which will make the whole stuff more solid with less code (and no verification at start necessary anymore).

midik:
oops :)

midik:
Before yesterday, one test (out of 3) failed because of that old S3 error. I've re-run them **today**, but, they failed because we've removed (renamed) `report` npm script:
```
$ REDIS=redis://127.0.0.1:6379 npm run report
npm ERR! missing script: report
```
The rest steps of that test are green. Branch test also green. I'd suggest to don't mind.

kelson42:
@midik Would you please able rebase on origin/master ?
@bakshiutkarsha We are waiting to your review!

kelson42:
I'll just merge, this is a trivial thing and the PR is around for long enough.

PR #1106, opened 2020-05-08
Add support for parametrized translatable strings

dnohales:
This allows to have translatable strings with parameters like:

```
My name is ${name}
```

This is crucial for languages which the variable part of the translatable string is not necessarily at the end of the string like assumed in the `ISSUED_BY` string.

I'm also refactored the footer to use this kind of strings, the bad news is that the creator and license links couldn't be made inside the Swig template.


kelson42:
@midik Can you please rebase on `master` next time, instead of merging `master`?

midik:
@dnohales could you please make sure that the coverage isn't decreased?

dnohales:
I added a test for the new function I've made. I'm not sure how `S3.ts` has downgraded in coverage since it's unrelated to this.

midik:
I have no idea why codecov complains on that, except that's probably because we've changed tests entry point [recently](https://github.com/openzim/mwoffliner/commit/b447bc2bf6f25e38a37792efcf21829a55051e26).

PR #1114, opened 2020-05-13
removed duplicate test

### LINKED ISSUE #1110: Travis "S3 image not found" test Failure
Sometime, due to parallel running of different node version in Travis CI, the build fails with error :

S3 image not found, it fails because one version is deleting the image and other is trying to find the same image in S3 bucket at the same time, hence the test fails.

bakshiutkarsha:
should resolve #1110 .
This test case has already been taken care in S3 test file.


kelson42:
@bakshiutkarsha Where is exactly the duplicate of the part of have removed?

bakshiutkarsha:
> @bakshiutkarsha Where is exactly the duplicate of the part of have removed?

As mentioned in my comment earlier "This test case has already been taken care in S3 test file." 
S3.test.ts > line 30

PR #1115, opened 2020-05-14
Add Spanish translation

dnohales:
Not much to say here :)


kelson42:
We should seriously think about implementing #1116

dnohales:
> We should seriously think about implementing #1116

Ah, I didn't know about this, I suppose we could safely close this then.

kelson42:
> > We should seriously think about implementing #1116
> 
> Ah, I didn't know about this, I suppose we could safely close this then.

No, it's helpful. We could move the translations to TW when we are so far. Thx for the one in Spanish.

PR #1117, opened 2020-05-15
Subtitle Downloading

### LINKED ISSUE #877: Mediawiki video subtitles should be scraped too
See for example https://commons.wikimedia.org/wiki/File:Carcinoid_syndrome.webm. It has arabic subtitle. Article of WPEN integrates it https://en.wikipedia.org/wiki/Carcinoid_syndrome. Currently the subtitle links are in the HTML code, but the subtitles files are not downloaded and URL not rewritten.

bakshiutkarsha:
should resolve #877


kelson42:
@bakshiutkarsha What is the status of this PR?! There is not WIP or draft... but nobody is assigned to review it?!

bakshiutkarsha:
> @bakshiutkarsha Please secure that the the development process is respected properly. Missing the review assignement like all other kinds of thing going forgotten in the process (not putting draft, not reviewieng, forgotting to merge, forgotting to rebase properly, etc...) just slows down the whole process and creates overhead (often on my side).
> 
> Concerning the testing, the instructions were:
> 
> > Here is how your test should looks like IMO: (1) you have a small wikicode for a video with subtitle (2) you send it to https://en.wikipedia.org/api/rest_v1/#/Transforms/post_transform_wikitext_to_html (3) you get the result rewrite it and verify that the track URL is proper.
> 
> So far I can see, (3) is there but not (1) and (2) which makes the whole test relatively weak. What will happen if for some reason the HTML generated by Mediawiki changes?! The answer is that it might go through unoticed!

@kelson42 I will make sure in future, these things won't affect the development timeline and the process is respected properly.

src/util/saveArticles.ts
```diff
@@ -402,9 +403,32 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
 
     sourceEl.setAttribute('src', newUrl);
+
+    /* Scrape subtitle */
+    const trackEle = videoEl.querySelector('track');
+    if (trackEle) {
+        const { content, trackTitle } = await treatSubtitles(trackEle, downloader, webUrlHost, mw);
+        const article = new ZimArticle({ url: `${trackTitle}.vtt`, mimeType: 'text/vtt', data: content, ns: 'I' });
+        zimCreator.addArticle(article);
+        trackEle.setAttribute('src', `${getRelativeFilePath(articleId, trackTitle, 'I')}.vtt`);
```
midik, line 413: are you sure should we put subtitles into `I` namespace? ;)

src/util/saveArticles.ts
```diff
@@ -291,10 +292,10 @@ async function getModuleDependencies(articleId: string, mw: MediaWiki, downloade
     };
 }
 
-async function processArticleHtml(html: string, downloader: Downloader, mw: MediaWiki, dump: Dump, articleId: string) {
+async function processArticleHtml(html: string, downloader: Downloader, mw: MediaWiki, dump: Dump, articleId: string, zimCreator: ZimCreator) {
     let mediaDependencies: Array<{ url: string, path: string }> = [];
     let doc = domino.createDocument(html);
-    const tmRet = await treatMedias(doc, mw, dump, articleId);
+    const tmRet = await treatMedias(doc, mw, dump, articleId, downloader, zimCreator);
```
midik, line 298: not sure it's a good idea to carry `downloader` and `zimCreator` 3 levels down into `treatMedias`. Normally, we get a set of links back to upper level then handle them by iterating mediaDependencies [here](https://github.com/openzim/mwoffliner/blob/9ea0be2c9fed9463c9f380627e608af882c6cddf/src/util/saveArticles.ts#L165). This way we don't deal with zim file in `treat*()` functions which follow OOP spirit. I belive the function called `processArticleHtml` should know nothing about the entity like zim, so I feel there's better way to handle this ;)
Could we handle subtitles outside this function?

src/util/saveArticles.ts
bakshiutkarsha: If there is no namespace for the resource, default is "I", do you suggest something different?

src/util/saveArticles.ts
bakshiutkarsha: I am not sure I understood this correctly, I agree with point that downloader and zimcreator got dragged 3 level down, but trackEle is tightly coupled with videoEle, so IMO it makes sense to have them in a same place. But le me know if you suggest differently?

src/util/saveArticles.ts
bakshiutkarsha: For all the content and image download, if no namespace is provided the default is "I", so I followed that, but as I can see for other module files(css, js) default convention is "-", so I will change it to that.

src/util/saveArticles.ts
```diff
@@ -402,9 +403,32 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+        zimCreator.addArticle(article);
+        trackEle.setAttribute('src', `${getRelativeFilePath(articleId, trackTitle, 'I')}.vtt`);
+    }
     return { mediaDependencies };
 }
 
+export async function treatSubtitles(trackEle: any, downloader: Downloader, webUrlHost: string, mw: MediaWiki) {
+    try {
+        const sourceUrl = getFullUrl(webUrlHost, trackEle.getAttribute('src'), mw.base);
+        const trackformat: any = QueryStringParser.parse(sourceUrl).trackformat;
```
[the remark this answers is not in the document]
bakshiutkarsha: Yeah even I was confused to get this weird casting error, if I put type string:

`Type 'ParsedUrlQuery' is not assignable to type 'string'`, so the quick fix was to change to any type. Can you suggest how to fix it?
Screenshot 2020-05-17 at 1 37 40 AM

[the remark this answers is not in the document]
midik: @bakshiutkarsha try this:
image
it's not required to specify the types of these two properties because they follows from `parse()` definition. However, `trackformat` defined as `string | string[]` there, so that we have to add explicit type guard bellow while calling `replace()`.
Extra point: please follow common variable naming style (`trackformat` -> `trackFormat`) as well ;)

src/util/saveArticles.ts
midik: I just mean that I was wondering why we put the subtitles in one namespace next to the images while subtitles are the text by nature ;)

src/util/saveArticles.ts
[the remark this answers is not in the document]
midik: UPD: same thing for `trackTitle` bellow ;)
Another approach would be to specify the types after all, but it wouldn't look decent IMO:
```
const {trackTitle, trackformat} = QueryStringParser.parse(sourceUrl) as {trackTitle: string, trackformat: string};
```

src/util/saveArticles.ts
```diff
@@ -402,9 +403,32 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+    if (trackEle) {
+        const { content, trackTitle } = await treatSubtitles(trackEle, downloader, webUrlHost, mw);
+        const article = new ZimArticle({ url: `${trackTitle}.vtt`, mimeType: 'text/vtt', data: content, ns: 'I' });
+        zimCreator.addArticle(article);
+        trackEle.setAttribute('src', `${getRelativeFilePath(articleId, trackTitle, 'I')}.vtt`);
+    }
     return { mediaDependencies };
 }
 
+export async function treatSubtitles(trackEle: any, downloader: Downloader, webUrlHost: string, mw: MediaWiki) {
```
midik, line 418: Please consider to add returnable types for the methods and functions as possible, both to newly created ones and existing ones that you make changes in. It's worth spending a little time.

src/util/saveArticles.ts
midik: Probably it does make sense to keep status quo because that would require so much changes to make the order here in `processArticleHtml()`. Let's table this until we get a resources to catch up a technical debt.

src/util/saveArticles.ts
bakshiutkarsha: Yes, I am aware that the rest of the text content file(js, css) uses "-" as namespace, but how they produce links are also different [here]. So, that was the reason I found using "I" namespace best for this usecase.(https://github.com/openzim/mwoffliner/blob/9ea0be2c9fed9463c9f380627e608af882c6cddf/src/util/misc.ts#208)

src/util/saveArticles.ts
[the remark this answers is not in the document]
bakshiutkarsha: @midik Thanks a lot for the help, I used the latter solution, because it only requires to define the type once, which is more efficient.

Also, for following common variable naming style, after doing the props destructing changes, I am getting the "trackformat" from url directly, so can't do much there.

test/util.ts
```diff
@@ -64,3 +66,15 @@ export async function zimcheckAvailable() {
 export async function zimcheck(filePath: string) {
     return execa.command(`${zimcheckPath} ${filePath}`);
 }
+
+export async function convertWikicodeToHtml(wikicode: string, dump:Dump): Promise<string>{
```
midik, line 70: @bakshiutkarsha please don't forget to fix eslint errors here ;)

test/unit/saveArticles.test.ts
```diff
@@ -238,3 +238,16 @@ test('--customFlavour', async (t) => {
     t.ok(ParisDocument.querySelector('#PRE_PROCESSOR'), `Paris was correctly pre-processed`);
     t.ok(PragueDocument.querySelector('#POST_PROCESSOR'), `Prague was correctly post-processed`);
 });
+
+test('Test Subtitles', async(t) => {
+    const { downloader, mw, dump } = await setupScrapeClasses({ format: '' });
+
+    // Wikicode is taken from article "Mechanical energy" which has a video with subtitle
+    const wikicode = `==Conservation of mechanical energy==
```
kelson42, line 246: I don't understand why we have a paragraph title here.

test/unit/saveArticles.test.ts
```diff
@@ -238,3 +238,16 @@ test('--customFlavour', async (t) => {
...
+    [[File:Physicsworks.ogv|thumb|200px|alt="Lecture demonstrating conservation of mechanical energy"|MIT professor [[Walter Lewin]] demonstrating conservation of mechanical energy]]`;
+
+    const htmlDoc = domino.createDocument(await convertWikicodeToHtml(wikicode, dump));
+    const contentRes = await treatSubtitles(htmlDoc.querySelector('track'), downloader, 'en.wikipedia.org', mw);
+
+    t.ok(contentRes, 'Successfully downloaded the subtitles');
```
kelson42, line 252: * We should have in test a function which just takes as argument `wiki`, `html` `comment` and make the comparison.
* Never write "Successfully" or similar in a test comment. You could write it everywhere and this is useless and in particular if it fails. Just write in a few words what the test does like for example "video with subtitle rewriting".

src/util/saveArticles.ts
kelson42: I reopen this because I'm not happy about that either. We need to architecture this in a smarter way.

test/unit/saveArticles.test.ts
```diff
@@ -238,3 +238,16 @@ test('--customFlavour', async (t) => {
     t.ok(ParisDocument.querySelector('#PRE_PROCESSOR'), `Paris was correctly pre-processed`);
     t.ok(PragueDocument.querySelector('#POST_PROCESSOR'), `Prague was correctly post-processed`);
 });
+
+test('Test Subtitles', async(t) => {
```
kelson42, line 242: Don't write "Test"... we already know we are running tests.

src/util/const.ts
```diff
@@ -6,3 +6,4 @@ export const CONCURRENCY_LIMIT = 10;
 export const MIME_IMAGE_REGEX = /^image+[/-\w.]+$/;
 export const FIND_HTTP_REGEX = /^(?:https?:\/\/)?/i;
 export const DB_ERROR = 'internal_api_error_DBQueryError';
+export const WIKI_TO_HTML_API = 'api/rest_v1/transform/wikitext/to/html';
```
kelson42, line 9: Considering that this is used only in tests, this sould be in test modules.

src/util/saveArticles.ts
```diff
@@ -402,9 +403,33 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
     sourceEl.setAttribute('src', newUrl);
+
+    /* Scrape subtitle */
+    const trackEle = videoEl.querySelector('track');
+    if (!trackEle) {
+        return { mediaDependencies };
+    }
+
+    const { content, title } = await treatSubtitles(trackEle, downloader, webUrlHost, mw);
+    const article = new ZimArticle({ url: `${title}.vtt`, mimeType: 'text/vtt', data: content, ns: 'I' });
```
kelson42, line 414: Should be in `-` namespace IMO.

src/util/saveArticles.ts
```diff
@@ -402,9 +403,33 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+    trackEle.setAttribute('src', `${getRelativeFilePath(articleId, title, 'I')}.vtt`);
     return { mediaDependencies };
 }
 
+export async function treatSubtitles(trackEle: any, downloader: Downloader, webUrlHost: string, mw: MediaWiki): Promise<{ content: Buffer | string, title: string }> {
+    try {
+        const sourceUrl = getFullUrl(webUrlHost, trackEle.getAttribute('src'), mw.base);
+        const {trackformat, title} = QueryStringParser.parse(sourceUrl) as {title: string, trackformat: string};
+
+        const vttSourceUrl = sourceUrl.replace(trackformat, 'vtt');
```
kelson42, line 425: Super weak! what will happen if for some reason there is `srt` written somewhere else in the URL!

midik:
BTW, [here](https://github.com/nodejs/node/pull/33321)'s the fix for the bug that we're probably facing on v14

kelson42:
> BTW, [here](https://github.com/nodejs/node/pull/33321)'s the fix for the bug that we're probably facing on v14

Nice, happy if the bug is not on our side... and happy again to see this will be part of next v14 release (my guess).

test/unit/util.test.ts
```diff
@@ -60,3 +60,10 @@ test('Encoding ArticleId for Zim HTML Url', async(t) => {
         t.equal(encoded, encodeArticleIdForZimHtmlUrl(unencoded), `encodeArticleIdForZimHtmlUrl() encoding`);
     }
 });
+
+test('wikitext comparison', async(t) => {
+    const wikicode = `According to the principle of conservation of mechanical energy, the mechanical energy of an [[isolated system]] remains constant in time, as long as the system is free of [[friction]] and other non-conservative forces.`;
+    const html = `<p id="mwAQ">According to the principle of conservation of mechanical energy, the mechanical energy of an <a rel="mw:WikiLink" href="./Isolated_system" title="Isolated system" id="mwAg">isolated system</a> remains constant in time, as long as the system is free of <a rel="mw:WikiLink" href="./Friction" title="Friction" id="mwAw">friction</a> and other non-conservative forces.</p>`;
+    const resultHtml = await convertWikicodeToHtml(wikicode, 'https://en.wikipedia.org/');
+    t.equals(html, resultHtml, 'HTML matches')
```
midik, line 68: please follow the style with the tail

bakshiutkarsha: @midik I am not quite sure that I understood this correctly.

midik: @bakshiutkarsha sorry, I just mean a semicolon at the end of the line 68 ;)

bakshiutkarsha: Did that change already :)

src/util/saveArticles.ts
bakshiutkarsha: Reverted the change of dragging `downloader` and `zimcreator` to 3 level down and handled this in `processArticleHtml()` as suggested.

test/unit/s3.test.ts
```diff
@@ -30,7 +30,7 @@ _test('S3 checks', async(t) => {
     const imageExist = await s3.downloadIfPossible('bm.wikipedia.org/static/images/project-logos/bmwiki-2x.png', 'https://bm.wikipedia.org/static/images/project-logos/bmwiki-2x.png');
     t.assert(!!imageExist, 'Image exists in S3');
     // Checking the data related to image matches
-    t.equals(imageExist.headers.Metadata.etag, '"aeff-54a391a807034"', 'Etag matches');
+    t.equals(imageExist.headers.Metadata.etag, '"a740-5a6b0464619c2"', 'Etag matches');
```
kelson42, line 33: This is really weak... and consequently has been broken only a few months after been implemented. Automated test should not have the this ETAG hardcoded. If the whole lifecycle has been tested properly you should not need to do that. Should be pretty easy to fix.

test/unit/util.test.ts
```diff
@@ -60,3 +60,10 @@ test('Encoding ArticleId for Zim HTML Url', async(t) => {
         t.equal(encoded, encodeArticleIdForZimHtmlUrl(unencoded), `encodeArticleIdForZimHtmlUrl() encoding`);
     }
 });
+
+test('wikitext comparison', async(t) => {
```
kelson42, line 64: Good, but could you please abtract this with a function taking arguments `wikicode`, `html` and `comment`?

test/util.ts
```diff
@@ -64,3 +66,15 @@ export async function zimcheckAvailable() {
...
+
+export async function convertWikicodeToHtml(wikicode: string, baseUrl: string): Promise<string> {
+    try {
+        const resp = await axios.post(`${baseUrl}api/rest_v1/transform/wikitext/to/html`,  {
+            wikitext: wikicode,
+            body_only: true,
+        })
+        return resp.data;
+    } catch (err){
+        logger.log(`Got err during conversion of wikicode to HTML due to ${err}`);
```
kelson42, line 78: * `error` not `err`.
* the whole tst shoudl fail here if the HTML is not retrievable properly
* I don't think the temporary variable `resp` is really necessary just return directly the data.

src/util/saveArticles.ts
```diff
@@ -172,6 +174,10 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
                             }
                         }
 
+                        for (const dep of subtitles) {
```
kelson42, line 177: Here the variable name should be `subtitle`, not `dep`.

src/util/saveArticles.ts
```diff
@@ -402,7 +422,25 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
     }
 
     sourceEl.setAttribute('src', newUrl);
-    return { mediaDependencies };
+
+    /* Scrape subtitle */
+    const trackEle = videoEl.querySelector('track');
```
kelson42, line 427: It looks like you assume here there is only one subtitle?! It should obviously work (and be tested) with many.

src/util/saveArticles.ts
```diff
@@ -402,7 +422,25 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+    if (!trackEle) {
```
kelson42, line 428: The construct here is too complicated, just secure `subtitles` is an empty list if no subtitle and always to to the end of the function `treatVideo`.

test/unit/s3.test.ts
bakshiutkarsha: I have removed this case, because lifecycle is getting tested.

test/unit/util.test.ts
bakshiutkarsha: I have extracted the function, please let me know if that is how it intends to be?

src/util/saveArticles.ts
```diff
@@ -402,7 +421,24 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+        }
+    }
+    return { mediaDependencies, subtitles };
+}
+
+export async function treatSubtitles(trackEle: DominoElement, webUrlHost: string, mw: MediaWiki, articleId: string): Promise<string> {
+    const subtitleSourceUrl = getFullUrl(webUrlHost, trackEle.getAttribute('src'), mw.base);
+    const { title, lang } = QueryStringParser.parse(subtitleSourceUrl) as { title: string, lang: string };
+
+    const vttFormatUrl = subtitleSourceUrl.replace(/(trackformat=)[^\&]+/, `$1vtt`);
```
kelson42, line 439: Better use a javascript URL object here for the replacement of the argument value.
```
url.searchParams.set('trackformat', '....');
```

test/unit/util.test.ts
kelson42: Yes, in principe but:
* Don't use temporary variable, just put the strings in the fonction calls
* Reduce the size of your wikitext/html you test, keep only what is necessary to be tested. It does not have to be a senteence which is online available as well
* rename the fucntion to `testHtmlRewritingE2e()`

src/util/saveArticles.ts
```diff
@@ -402,7 +421,25 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
     }
 
     sourceEl.setAttribute('src', newUrl);
-    return { mediaDependencies };
+
+    /* Scrape subtitle */
+    const trackEle = videoEl.querySelectorAll('track');
+    if (trackEle) {
+        for (const track of Array.from(trackEle)) {
```
kelson42, line 428: This can be probably simplified in one line?!

src/util/saveArticles.ts
```diff
@@ -402,7 +421,25 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+            subtitles.push(await treatSubtitles(track, webUrlHost, mw, articleId));
+        }
+    }
+    return { mediaDependencies, subtitles };
+}
+
+export async function treatSubtitles(trackEle: DominoElement, webUrlHost: string, mw: MediaWiki, articleId: string): Promise<string> {
```
kelson42, line 435: I would like to have a clear differenciation in function/variable naming between `subtitle` = one track, and subtitles = multiple tracks. Here it seems you treat one track, so it should be singular form.

src/util/saveArticles.ts
```diff
@@ -402,7 +421,25 @@ async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
...
+    const subtitleSourceUrl = getFullUrl(webUrlHost, trackEle.getAttribute('src'), mw.base);
+    const { title, lang } = QueryStringParser.parse(subtitleSourceUrl) as { title: string, lang: string };
+
+    const vttFormatUrl =  new URL(subtitleSourceUrl);
```
kelson42, line 439: A small comment why we use vtt here would be helpful I believe.

test/unit/saveArticles.test.ts
```diff
@@ -238,3 +238,26 @@ test('--customFlavour', async (t) => {
...
+    const htmlDoc = domino.createDocument((await convertWikicodeToHtml(wikicode, dump.mwMetaData.base)).data);
+    const contentRes = await treatSubtitles(htmlDoc.querySelector('track'), 'en.wikipedia.org', mw, 'Mechanical energy');
+
+    t.ok(contentRes, 'Video with subtitle rewriting');
+});
+
+test('treat multiple subtitles in one video', async(t) => {
+    const { downloader, mw, dump } = await setupScrapeClasses({ format: '' });
+
+    // Wikicode is taken from article "User:Charliechlorine/sandbox" which has multiple(4) subtitles in this video
```
kelson42, line 256: Why you don't use `testHtmlRewritingE2e` here (and in other places)?

kelson42:
@bakshiutkarsha It needs a few changes and the CI shoudl pass.

test/unit/saveArticles.test.ts
bakshiutkarsha: Are you indicating to replace `convertWikicodeToHtml` to `testHtmlRewritingE2e`. It is bit unclear to me what exactly needs to be done here?

kelson42: Replace 'treat one subtitle' and 'treat multiple subtitles in one video' ... or add a new test.... which is only based on `testHtmlRewritingE2e` and test that the HTML for a video with multiple subtitles is OK.

src/util/saveArticles.ts
kelson42: @bakshiutkarsha Please close the discussion which problems have been resolved by you.

test/unit/util.test.ts
```diff
@@ -60,3 +60,12 @@ test('Encoding ArticleId for Zim HTML Url', async(t) => {
...
     }
 });
+
+async function testHtmlRewritingE2e(t: any, wikicode: string, html: string, comment: string){
+    const resultHtml = await convertWikicodeToHtml(wikicode, 'https://en.wikipedia.org/');
+    t.equal(html, resultHtml.data, comment);
+}
+
+test('wikitext comparison', async(t) => {
+    testHtmlRewritingE2e(t, `An [[isolated system]] remains the system is free.`, `<p id="mwAQ">An <a rel="mw:WikiLink" href="./Isolated_system" title="Isolated system" id="mwAg">isolated system</a> remains the system is free.</p>`, 'HTML and Wikitext match')
```
kelson42, line 70: Try to avoid lines of code with more than 80 characters. Here the readability would be improved with somethin like:
```
testHtmlRewritingE2e(
  t,
  `An [[isolated system]] remains the system is free.`,
  `<p id="mwAQ">An <a rel="mw:WikiLink" href="./Isolated_system" title="Isolated system" id="mwAg">isolated system</a> remains the system is free.</p>`,
  'HTML and Wikitext match')
```

test/unit/saveArticles.test.ts
kelson42: My last comment is confusing:
* 'treat one subtitle' should test exactly which URL
* 'treat multiple subtitles in one video' should do the same
* I would add an global test, testing the whole HTML testHtmlRewritingE2e

kelson42:
@bakshiutkarsha Can you please rebase on `master`?

bakshiutkarsha:
- 'treat one subtitle' should test exactly which URL - DONE
- 'treat multiple subtitles in one video' should do the same - DONE
- I would add an global test, testing the whole HTML testHtmlRewritingE2e - DONE, but can you confirm is this what you wanted?

bakshiutkarsha:
> You should have one test for `treatSubtitle()`, one test for `treatVideo()` and one E2E test for a video wiki text using `testHtmlRewritingE2e()`.
> 
> Yout current usage of `testHtmlRewritingE2e()` for subtitles is useless because you test it with the value of your own code... so this is probably always going to work. A unit test should test against hardcoded values (so here HTML)

The HTML I am getting from the API is not exactly same as the online version of it. For eg. API gives me HTML starting from <figure> tag and there is no such tag in online version of it. So, what should be done in this case?

PR #1118, opened 2020-05-15
Ubuntu Xenial is deprecated

kelson42:


kelson42:
Wrong Xenial is support until 2021

PR #1120, opened 2020-05-15
Add `latest` tag

### LINKED ISSUE #556: Publish automatically on npmjs & docker hub
I synchronise manually the Docker hub image, this is something I want to automatise
* For each merge to master, update a "dev" version
* For each release tag, publish a new version for "latest" and one with the tag.

Same should happen on npmjs. All of this should run/be trigerred on Travis-CI.

midik:
Fixes #556 

@kelson42 not sure how this tag is being handled on docker hub side. Is that we need to do here?

### DIFF
--- a/dev/publish.sh
+++ b/dev/publish.sh
@@ -10,7 +10,7 @@ if [ -n "$CI" ]; then
                 npm publish --tag rc
             else
                 echo "This is a production release"
-                npm publish
+                npm publish --tag latest
             fi
         else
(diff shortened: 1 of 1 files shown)


kelson42:
@midik I think you are mixing Docker with npmjs.org. Like written here https://github.com/openzim/mwoffliner/issues/556#issuecomment-615851932 nothing to do here for Docker. Not sure what we need for npmjs.org.

PR #1143, opened 2020-06-05
Use separate images for different node versions

bakshiutkarsha:
- Will solve the frequent crashing of S3 test in CI build


kelson42:
@bakshiutkarsha Please create a ticket explaining the problem and your solution. This is not the proper approach to my opinion.

kelson42:
@bakshiutkarsha This does not seem to be necessary anymore I guess? Can we close this PR?

bakshiutkarsha:
@kelson42 I have closed it.

PR #1146, opened 2020-06-06
Fix the flow for non-parsoid wikis (including archwiki)

### LINKED ISSUE #1139: Latest MWoffliner has a new sporadic error pattern when using local Parsoid
We did not had the problem before... but many wikis without Parsoid API are often failing with the following error:
```
[warn] [2020-05-26T14:11:39.182Z] Failed to get article [Mastodon_(Español)] using remote Parsoid, trying with local one TypeError: Cannot read property 'displaytitle' of undefined
    at renderMCSArticle (/tmp/mwoffliner/lib/util/articleRenderers.js:101:39)
    at /tmp/mwoffliner/lib/util/articleRenderers.js:54:26
    at Generator.next (<anonymous>)
    at fulfilled (/tmp/mwoffliner/lib/util/articleRenderers.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[warn] [2020-05-26T14:11:40.633Z] Failed to get article [Extra_keyboard_keys] using remote Parsoid, trying with local one TypeError: Cannot read property 'displaytitle' of undefined
    at renderMCSArticle (/tmp/mwoffliner/lib/util/articleRenderers.js:101:39)
    at /tmp/mwoffliner/lib/util/articleRenderers.js:54:26
    at Generator.next (<anonymous>)
    at fulfilled (/tmp/mwoffliner/lib/util/articleRenderers.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[warn] [2020-05-26T14:11:41.180Z] Failed to get article [Multihead] using remote Parsoid, trying with local one TypeError: Cannot read property 'displaytitle' of undefined
    at renderMCSArticle (/tmp/mwoffliner/lib/util/articleRenderers.js:101:39)
    at /tmp/mwoffliner/lib/util/articleRenderers.js:54:26
    at Generator.next (<anonymous>)
    at fulfilled (/tmp/mwoffliner/lib/util/articleRenderers.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

See for example:
* https://farm.openzim.org/pipeline/5ecd1e0aa15fcb47c20934d9/debug
* https://farm.openzim.org/pipeline/5ecf490218b11fb99e8bfa07/debug

It seems we have a problem in retrieving some kind of JSON.

midik:
Fixes #1139

### DIFF
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -16,7 +16,7 @@ import {getStrippedTitleFromHtml} from './misc';
 export const renderArticle = async (json: any, articleId: string, dump: Dump, forceParsoidFallback: boolean): Promise<RenderedArticle[]> => {
 
     const articleDetail = await articleDetailXId.get(articleId);
-    const useParsoidFallback = forceParsoidFallback || json.visualeditor?.result;
+    const useParsoidFallback = forceParsoidFallback || json.visualeditor?.result || json?.contentmodel === 'wikitext';
 
     if (useParsoidFallback) {
         const isMainPage = articleId === dump.mwMetaData.mainPage;
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -113,7 +113,7 @@ class Downloader {
     this.backoffOptions =  {
       strategy: new backoff.ExponentialStrategy(),
       failAfter: 7,
-      retryIf: (err: any) => err.code === 'ECONNABORTED' || err.response?.status !== 404,
+      retryIf: (err: any) => err.code === 'ECONNABORTED' || ![403, 404].includes(err.response?.status),
       backoffHandler: (number: number, delay: number) => {
         logger.info(`[backoff] #${number} after ${delay} ms`);
       },


midik:
This is quicker approach replacing the one taken in #1138.

src/util/articleRenderers.ts
```diff
@@ -16,7 +16,7 @@ import {getStrippedTitleFromHtml} from './misc';
 export const renderArticle = async (json: any, articleId: string, dump: Dump, forceParsoidFallback: boolean): Promise<RenderedArticle[]> => {
 
     const articleDetail = await articleDetailXId.get(articleId);
-    const useParsoidFallback = forceParsoidFallback || json.visualeditor?.result;
+    const useParsoidFallback = forceParsoidFallback || json.visualeditor?.result || json?.contentmodel === 'wikitext';
```
kelson42, line 19: This line and what is going one in this whole module needs an explanation because I don't get it. Our code never ever should deal with wikitext code. Why we need to "render" anything if we don't have any wikitext?

midik: > Our code never ever should deal with wikitext code

However, should we or not, we do get wikitext:
image
... that was the reason to `renderMCSArticle()` used to fail (there's no particular json properties). As designed `renderMCSArticle()` shouldn't been called in this case - we already have an html.

Moreover, there is the conditional branch in the source code, where we do handle that:
image

As by me, this is just an one of many things there that we have organized worst. That was 
another reason why I started to refactor the things around in that PR before.

kelson42: Which API call (an example) deliver wikitext?

kelson42: The whole purpose of the module is still not explained.

midik: > Which API call (an example) deliver wikitext?
We asking local parsoid for `pagebundle` as following:
```
http://localhost:8000/wiki.archlinux.org/v3/page/pagebundle/Downgrading_packages/606085
```
so it returns both `html` and `data-parsoid` attributes. I'm not sure what we need to ask for a bundle here because, according to the sources, we ignore everything but `html` at that point. Probably we need the urls as following?
```
http://localhost:8000/wiki.archlinux.org/v3/page/html/Downgrading_packages/606085
```

> The whole purpose of the module is still not explained.
Not sure what you mean asking this. There we assemble the content from sections etc.

kelson42:
This PR is outdated. This decision of taking a Parsoid fallback or not is now taken, once, at the start of the scraping process.

Issue #1145, opened 2020-06-05
Avoid hardcoded ETAG in automated tests


kelson42:
One of our S3 automated - relying on a Wikipedia image - test has an ETAG hardcoded. This is weak and within a few months the test has been broken becuase things has been changed on Wikipedia side.

We should remove the hardcoded ETAG in our test.

The test should be rewritten like this:
* Delete the image in S3
* Make a (download) run to test we download the image from Wikipedia (ETAG dynamically retrieved from online request).
* Test the image is properly uploaded to S3 with the proper ETAG
* Make an other run to secure the image comes from the S3 this time
* Make an other run and screwing the ETAG, check the image comes this time from Wikipedia
* Check that the image in S3 has been refreshed properly with proper ETAG

kelson42:
@bakshiutkarsha What is the status here? This is one of the blocker for the rest!

bakshiutkarsha:
> @bakshiutkarsha What is the status here? This is one of the blocker for the rest!
I took care of this in PR 1147.

PR #1147, opened 2020-06-08
optimisation of S3 flow

### LINKED ISSUE #1144: Reduce S3 exposure to race conditions
Race conditions happen around S3 because we have latencies between multiple steps of the optimization cache workflow.

This is a proposition to simplify/speed-up the process around S3 cache.

Currently with do typically 3 HTTP requests:
1 - Download a content for S3
2 - Check its etag value against online value (via a HEAD request)
3 - If not OK (rare) then issue a request to Mediawiki to downlaod the content

We would be quick by doing:
1 - Download a content for S3
2 - Downoad - or not - the content from Mediawiki by sending the get request with the ETAG coming from S3 and a a conditional header https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests

An example of GET request which will just do nothing:
```
curl -H 'If-None-Match: "3e7c-5a6b0464619c2"' https://bm.wikipedia.org/static/images/project-logos/bmwiki.png
```
But is the ETAG is different, then the content will be downloaded.

That approach reduces by 30% the typical latency and should help to avoid race conditions... beside the fact that it should seriously speed up things.

bakshiutkarsha:
should resolve #1144  and #1145


bakshiutkarsha:
The build is failing due to this intermittent error on Node JS 14:

```
#FailureMessage Object: 0x7ffc22044a50
 1: 0xa975a1  [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 2: 0x1972504 V8_Fatal(char const*, ...) [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 3: 0xe5c609 v8::internal::GlobalBackingStoreRegistry::Register(std::shared_ptr<v8::internal::BackingStore>) [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 4: 0xbc98d8 v8::ArrayBuffer::GetBackingStore() [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 5: 0xa037ef node::Buffer::Data(v8::Local<v8::Object>) [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 6: 0xaea939  [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 7: 0xc06c4b  [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 8: 0xc081f6  [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
 9: 0xc08876 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
10: 0x13a9fb9  [/home/travis/.nvm/versions/node/v14.4.0/bin/node]
Illegal instruction (core dumped)
```

src/S3.ts
```diff
@@ -62,21 +63,11 @@ class S3 {
         }
     }
 
-    public async downloadIfPossible(upstreamUrl: string, requestUrl: string): Promise<any> {
+    public async downloadIfPossible(upstreamUrl: string): Promise<any> {
```
kelson42, line 66: This should be:
```
public async downloadBlob(key: string): Promise<any>
```

kelson42: I don't really understand why `deleteBlob` has an other API than `downloadBlob`. I believe it should be:
```
public async deleteBlob(key: string): Promise<any>
```

src/Downloader.ts
```diff
@@ -677,6 +670,49 @@ class Downloader {
...
+    if (etag.substr(0, 2) !== 'W/') {
+        return etag;
+    }
+    const strongEtag =  etag.replace('W/', '').replace(/"/g, '');
+    return strongEtag;
+  }
+
+  private async checkStatusAndUploadtoS3(requestOptions: any, handler: any, imageResp: any) {
+    if (imageResp) {
+      requestOptions.headers['If-None-Match'] = await this.checkAndReplaceWeakEtag(imageResp.Metadata.etag);
```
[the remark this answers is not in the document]
kelson42: What is the story around weak etags, why we store them in S3 if this is to ignore them afterward?

src/Downloader.ts
```diff
@@ -677,6 +670,49 @@ class Downloader {
...
+      }
+
+      if (etag) {
+        this.s3.uploadBlob(this.stripHttpFromUrl(requestOptions.url), resp.data, etag);
+      }
+
+      handler(null, {
+        responseHeaders: resp.headers,
+        content: compressionWorked ? content : resp.data,
+      });
```
kelson42, line 702: This code seems OK, but is a bit unclear:
* First of all put a clear comment after the axios request to say that most of the time the code won't be executed becaue of the HTTP 304
* Then handle the compression stuff
* Then handle the etag/upload stuf
* Don't understand why you put `content` in `resp.data`

test/unit/downloader.test.ts
```diff
@@ -171,14 +171,41 @@ _test('Downloader class with optimisation', async (t) => {
     const httpOrHttpsRemoved = downloader.stripHttpFromUrl(testImage);
```
kelson42, line 171: Please move `stripHttpFromUrl()` to a a "utils' js module so we can use it as a function. Having it as a method of the Downloader is not proper.

src/Downloader.ts
```diff
@@ -649,15 +649,8 @@ class Downloader {
...
-          if (s3ImageResp) {
-            handler(null, {
-              responseHeaders: s3ImageResp.headers,
-              content: s3ImageResp.imgData,
-            });
-          } else {
-            await this.imageDownloadCompressAndUploadToS3(requestOptions, handler);
-          }
+        this.s3.downloadIfPossible(this.stripHttpFromUrl(requestOptions.url)).then(async (s3ImageResp) => {
+          this.checkStatusAndUploadtoS3(requestOptions, handler, s3ImageResp);
```
kelson42, line 653: Merge this two lines and with `checkStatusAndUploadtoS3()` and call the method `downloadImage()`.

kelson42:
@midik Could you please review this PR and put a particular attention on await/async and error catching to be sure that we handle this perfectly?

src/Downloader.ts
kelson42: Actually you should not treat status HTTP 304 as an error. You should tell Axios with `validateStatus` that 304 is OK and then treat it properly.

src/Downloader.ts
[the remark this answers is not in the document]
bakshiutkarsha: checkAndReplaceWeakEtag() can be advantageously replaced by a simple regex. All of this is synchronous, no reason to use await I believe [DONE]

What is the story around weak etags, why we store them in S3 if this is to ignore them afterward?
Now I am storing the etags in S3 after refining them.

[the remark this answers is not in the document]
kelson42: IMO, it would be advantageous to write a small migration script an run it only once.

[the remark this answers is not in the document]
kelson42: OK, lets's keep it like this...

PR #1155, opened 2020-06-09
Rename parameter

### LINKED ISSUE #1152: Rename a few variables related to local parsing
`--forceLocalParser` seems not documented, I don't get it displayed if I just call `$mwoffliner`. It should be `--forceLocalParser` and not `--forceLocalParsoid`. Subsequent variables should be renamed that way.

midik:
Fixes #1152.

Note: this flag had no effect and has no effect so far. We need to refactor parameters seriously (starting from how they are being used currently - we can get this info using `schedules` from zimfarm API - then remove odd/unused and simplify them) - not only rename them selectively.

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -48,7 +48,7 @@ interface DownloaderOpts {
   useDownloadCache: boolean;
   downloadCacheDirectory?: string;
   noLocalParserFallback: boolean;
-  forceLocalParsoid: boolean;
+  forceLocalParser: boolean;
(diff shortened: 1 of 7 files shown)


kelson42:
> this flag had no effect and has no effect so far.

Why not here ? https://github.com/openzim/mwoffliner/pull/1155/commits/8e51c49d0b33a2e7d1346d58e4b6766f5e26aeba#diff-f9a469579c2c80d21239ff3f38522ffdR159

PR #1157, opened 2020-06-13
Remove fallback logic + cleanup + small refactoring

### LINKED ISSUE #1154: Remove ParsoidFallback logic
I don't want to have for each `Download.getArticle()` a fallback logic. This is unecessary and makes the code difficult to read/execute. The backend - MWoffliner deals with - won't change during we run a scrape, so we should check/detect (see `checkCapabilities()`) whatever has to be detected at the start of the process and then stick to then and just fail is there is a problem (after applying the ususal retry strategies).

Therefore `Downloader.parsoidFallbackUrl` variable should be removed.

Therefore `Downloader.forceParsoidFallback` variable should be removed.

Therefore `Downloader.getArticle` should be revamped/simplified, the recursion in it avoided.

midik:
Fixes #1154

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -402,25 +402,8 @@ class Downloader {
 
     logger.info(`Getting article [${articleId}] from ${articleApiUrl}`);
 
-    try {
-      const json = await this.getJSON<any>(articleApiUrl);
-      if (json.type === 'api_error') {
-        this.forceParsoidFallback = true;
-        forceParsoidFallback = true;
-        logger.error(`Received an "api_error", forcing all article requests to use Parsoid fallback`);
-        throw new Error(`API Error when scraping [${articleApiUrl}]`);
-      }
-      return await renderArticle(json, articleId, dump, forceParsoidFallback);
-
-    } catch (err) {
-      if (forceParsoidFallback) throw err;
-      if (err?.response?.status === 404) throw err;
-
-      // falling back to local Parsoid
-      const errMsg = err.response ? JSON.stringify(err.response.data, null, '\t') : err;
-      logger.warn(`Failed to get article [${articleId}] using remote Parsoid, trying with local one`, errMsg);
-      return await this.getArticle(articleId, dump, true);
-    }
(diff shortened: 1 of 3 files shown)


src/Downloader.ts
```diff
@@ -171,13 +169,11 @@ class Downloader {
...
         await this.initLocalServices();
-        const domain = (urlParser.parse(this.mw.base)).host;
+        const domain = urlParser.parse(this.mw.base).host;
         this.restApiUrl = `http://localhost:6927/${domain}/v1/page/mobile-sections/`;
 
         if (!this.mwCapabilities.veApiAvailable) {
-          const webUrlHost = urlParser.parse(this.mw.webUrl).host;
-          // todo will be changed in #1154
-          this.parsoidFallbackUrl = `http://localhost:8000/${webUrlHost}/v3/page/pagebundle/`;
+          this.veApiUrl = `http://localhost:8000/${domain}/v3/page/pagebundle/`;
```
kelson42, line 176: Are you really sure that the local parsoid provides a veApiUrl? I believe this is a restApiUrl.

src/Downloader.ts
```diff
@@ -393,33 +389,16 @@ class Downloader {
...
-      if (forceParsoidFallback) throw err;
-      if (err?.response?.status === 404) throw err;
-
-      // falling back to local Parsoid
-      const errMsg = err.response ? JSON.stringify(err.response.data, null, '\t') : err;
-      logger.warn(`Failed to get article [${articleId}] using remote Parsoid, trying with local one`, errMsg);
-      return await this.getArticle(articleId, dump, true);
-    }
+    const json = await this.getJSON<any>(articleApiUrl);
+    return await renderArticle(json, articleId, dump, this.mwCapabilities);
```
kelson42, line 401: No try() catch() at all anymore. Hope you are sure about that.

midik: yes, that wrapped by outer try-catch and proven to not to fail unhandled

kelson42:
@midik Can you please rebase on `master` and sort out my last question/comment?

PR #1168, opened 2020-06-18
Add url host and protocol in mediawiki class

### LINKED ISSUE #1160: webUrlHost webUrlProtocol should be a property of Mediawiki class
Everywhere the code should rely on `mw.webUrlHost()`. It should not be (re-)computed each time (so lazy property) locally when we use it. It should return a value computed at the time of `mw.webUrl()` has been set.

Same with `urlParser.parse(mw.webUrl).protocol`

bakshiutkarsha:
should resolve #1160

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -15,6 +15,8 @@ class MediaWiki {
   public readonly modulePath: string;
   public readonly spaceDelimiter: string;
   public readonly webUrl: string;
+  public readonly webUrlHost: string;
+  public readonly webUrlProtocol: string;
   public readonly apiUrl: string;
   public readonly veApiUrl: string;
   public readonly restApiUrl: string;
@@ -43,6 +45,8 @@ class MediaWiki {
     this.wikiPath = config.wikiPath ?? 'wiki/';
 
     this.webUrl = urlParser.resolve(this.base, this.wikiPath);
+    this.webUrlHost = urlParser.parse(this.webUrl).host;
+    this.webUrlProtocol = urlParser.parse(this.webUrl).protocol;
 
     this.apiResolvedUrl = urlParser.resolve(this.base, this.apiPath);
     this.apiUrl = `${this.apiResolvedUrl}?`;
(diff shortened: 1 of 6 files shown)


src/MediaWiki.ts
```diff
@@ -15,6 +15,8 @@ class MediaWiki {
   public readonly modulePath: string;
   public readonly spaceDelimiter: string;
   public readonly webUrl: string;
+  public readonly webUrlHost: string;
+  public readonly protocol: string;
```
kelson42, line 19: This should be called `webUrlProtocol`.

src/util/rewriteUrls.ts
```diff
@@ -46,8 +45,7 @@ export async function rewriteUrl(articleId: string, mw: MediaWiki, dump: Dump, l
     }
 
     if (!hrefProtocol && href.slice(0, 2) === '//') {
-        const wikiProtocol = urlParser.parse(mw.webUrl).protocol;
-        href = `${wikiProtocol}${href}`;
+        href = `${mw.protocol}${href}`;
         linkNode.setAttribute('href', href);
```
kelson42, line 49: Remove `href` variable and just assign `${mw.protocol}${href}`

src/util/rewriteUrls.ts
```diff
@@ -46,8 +45,7 @@ export async function rewriteUrl(articleId: string, mw: MediaWiki, dump: Dump, l
...
         hrefProtocol = urlParser.parse(href).protocol;
```
kelson42, line 50: `hrefProtocol = mw.protocol` seems better to me.

src/util/saveArticles.ts
```diff
@@ -424,7 +423,7 @@ export async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolea
 
     /* Scrape subtitle */
     for (const track of Array.from(videoEl.querySelectorAll('track'))) {
-        subtitles.push(await treatSubtitle(track, webUrlHost, mw, articleId));
+        subtitles.push(await treatSubtitle(track, mw.webUrlHost, mw, articleId));
```
kelson42, line 426: I think now `treatSubtitle` would be advantageously changed to `treatSubtitle(track, mw, articleId)`?!

src/util/rewriteUrls.ts
bakshiutkarsha: The reason of having the `href` variable is it is getting used in other places as well.

PR #1170, opened 2020-06-19
Use normal default config for redis

### LINKED ISSUE #1163: Redis default configuration should be normal
I made the configuration special to us to avoid to specify it all the time. Now that the tool is more broadly use and that we have the ZIMfarm, we should come back to the "normal" configuration per default so user don't have to tune either the MWoffliner configuration or the Redis configuration to start to play with MWoffliner.

One time done, the configuration parameter to get it work on the ZIMfarm should be given to @rgaudin so it can be populated to all our MWoffliner recipes.

bakshiutkarsha:
should resolve #1163

### DIFF
--- a/src/Redis.ts
+++ b/src/Redis.ts
@@ -3,6 +3,7 @@ import logger from './Logger';
 
 class Redis {
   public client: any;
+
   constructor(argv: any, config: any) {
     this.client = redis.createClient(argv.redis || config.defaults.redisConfig);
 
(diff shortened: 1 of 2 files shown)


kelson42:
@bakshiutkarsha What would be the new configuration to setup in Zimfar to secure it continues to work?

kelson42:
> @bakshiutkarsha What would be the new configuration to setup in Zimfar to secure it continues to work?

Found out: `--redis=/dev/shm/redis.sock`

Issue #1178, opened 2020-06-21
All the files within one run should have the same date (in filename)


kelson42:
But this is not always the case for the moment. See for example: https://farm.openzim.org/pipeline/5ecd1e0da15fcb47c20934e3

PR #1181, opened 2020-06-22
Use same date within a single run

bakshiutkarsha:
should reolve #1178

### DIFF
--- a/src/Dump.ts
+++ b/src/Dump.ts
@@ -55,7 +55,7 @@ export class Dump {
 
     private formatFlavour: string;
 
-    constructor(format: string, opts: DumpOpts, mwMetaData: MWMetaData, customProcessor?: CustomProcessor) {
+    constructor(format: string, opts: DumpOpts, mwMetaData: MWMetaData, contentDate: string, customProcessor?: CustomProcessor) {
         this.mwMetaData = mwMetaData;
         this.opts = opts;
         this.customProcessor = customProcessor;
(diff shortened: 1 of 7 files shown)


bakshiutkarsha:
> * This does not fix the problem I believe because the date will be still different.

For the first point:
`This does not fix the problem I believe because the date will be still different.
`
I believe it should work as I am creating an instance of date before the `for` loop begins to create zims of different flavours.

For the last point:
`We don't need an extra getZimCreationDate for something which should be use only one time and which is a one line function.`

I am creating a method for this because it is being used in multiple places in test cases. Should I use hardcoded value of dates in test cases or should I duplicate the code?

kelson42:
> > * This does not fix the problem I believe because the date will be still different.
> 
> I believe it should work as I am creating an instance of date before the `for` loop begins to create zims of different flavours.

You call the method 3 times within the for loop, not before, see: https://github.com/openzim/mwoffliner/pull/1181/files#diff-4cc2b78cf5db017b0392c57c374d64dfR356. BTW, identation is wrong there.

> > * We don't need an extra getZimCreationDate for something which should be use only one time and which is a one line function
> 
> I am creating a method for this because it is being used in multiple places in test cases. Should I use hardcoded value of dates in test cases or should I duplicate the code?

Probably here it should be able to work (so not date in the filename) if not specified at all.

src/Dump.ts
```diff
@@ -33,7 +34,6 @@ export class Dump {
     public novid: boolean;
     public nopdf: boolean;
     public nodet: boolean;
-    public contentDate: string;
```
kelson42, line 36: `filenameDate`, not `contentDate` which is too vague/wrong.

src/mwoffliner.lib.ts
```diff
@@ -325,6 +325,9 @@ async function execute(argv: any) {
     // await trimUnmirroredPages(downloader); // TODO: improve simplify graph to remove the need for a second trim
   }
 
+  const date = new Date();
+  const contentDate = `${date.getFullYear()}-${(`0${date.getMonth() + 1}`).slice(-2)}`;
```
kelson42, line 329: Something like `const filenameDate = (new Date()).toISOString().slice(0, 7)` seems better to me. Only one line and simplier.

bakshiutkarsha: `(new Date()).toISOString().slice(0, 7)`
 will give "2020-06", which is only year and month. Just to confirm we do want the date as well, right? If yes, I just need to change the slice param.

PR #1195, opened 2020-06-26
Better redis deletion

kelson42:


kelson42:
At the end the PR is a lot simplier, so I merge it.

PR #1197, opened 2020-06-29
Removing concept of spaceDelimiter from MWOffliner

### LINKED ISSUE #1180: Remove the concept "spaceDelimiter"
This concept does not exist in Mediawiki and therefore should not exist in MWoffliner as well.

I believe it came in a first because of a misunderstanding around treatment to do with article lists.

As a matter of fact article lists shoudl contain the exact article id like known by Mediawiki API.

If our articles lists (which use to contain titles with _ in place of spaces) won't work anymore after removing this, then we will have to rerwite our articles lists.

Here are the occurences:
```
$ grep -r spaceDelimiter src/
src/Dump.ts:    spaceDelimiter: string;
src/Dump.ts:                radical += `_${pathParser.basename(this.opts.articleList).toLowerCase().replace(/ /g, this.opts.spaceDelimiter).replace(/\.\w{3}$/, '')}`;
src/types.d.ts:  spaceDelimiter?: string;
src/mwoffliner.lib.ts:    spaceDelimiter: '_',
src/mwoffliner.lib.ts:      spaceDelimiter: '_',
src/util/saveArticles.ts:        const namespace = id.indexOf(':') >= 0 ? id.substring(0, id.indexOf(':')).replace(/ /g, mw.spaceDelimiter) : '';
src/MediaWiki.ts:  public readonly spaceDelimiter: string;
src/MediaWiki.ts:    this.spaceDelimiter = config.spaceDelimiter || '_';
src/MediaWiki.ts:        const name = entry['*'].replace(/ /g, self.spaceDelimiter);
src/MediaWiki.ts:        const canonical = entry.canonical ? entry.canonical.replace(/ /g, self.spaceDelimiter) : '';
src/MediaWiki.ts:    const mainPage = entries.mainpage.replace(/ /g, self.spaceDelimiter);
```

bakshiutkarsha:
should resolve #1180

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -94,7 +94,7 @@ class MediaWiki {
       const entries = json.query[type];
       Object.keys(entries).forEach((key) => {
         const entry = entries[key];
-        const name = entry['*'].replace(/ /g);
+        const name = entry['*'];
         const num = entry.id;
         const allowedSubpages = ('subpages' in entry);
         const isContent = !!(entry.content !== undefined || util.contains(addNamespaces, num));
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -11,7 +11,7 @@ class MediaWiki {
   public metaData: MWMetaData;
   public readonly base: string;
   public readonly modulePath: string;
-  public readonly spaceDelimiter: string;
+  // public readonly spaceDelimiter: string;
   public readonly webUrl: Url;
   public readonly apiUrl: Url;
   public readonly veApiUrl: Url;
@@ -31,7 +31,7 @@ class MediaWiki {
     this.domain = config.domain || '';
     this.username = config.username;
     this.password = config.password;
-    this.spaceDelimiter = config.spaceDelimiter || '_';
+    // this.spaceDelimiter = config.spaceDelimiter || '_';
     this.getCategories = config.getCategories;
 
     this.base = ensureTrailingChar(config.base, '/');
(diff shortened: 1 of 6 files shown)


src/MediaWiki.ts
```diff
@@ -203,7 +201,7 @@ class MediaWiki {
     const body = await downloader.query(query);
     const entries = body.query.general;
 
-    const mainPage = entries.mainpage.replace(/ /g, self.spaceDelimiter);
+    const mainPage = entries.mainpage.replace(/ /g, '_');
```
kelson42, line 204: I don’t understand why we still have a replace() here.

bakshiutkarsha: This is what we call, in the very beginning to getSiteInfo, where in response we get a mainpage. Now, eg, if we are doing the scrape for the bm.wikipedia,

The query that we request looks like:
`action=query&meta=siteinfo&format=json&siprop=general|namespaces|statistics|variables|category|wikidesc
`

Response (only relevant part of response JSON):
```
{
  mainpage: 'Nyɛ fɔlɔ',
  base: 'https://bm.wikipedia.org/wiki/Ny%C9%9B_f%C9%94l%C9%94',
  sitename: 'Wikipedia',
  logo: '//bm.wikipedia.org/static/images/project-logos/bmwiki.png',
  generator: 'MediaWiki 1.35.0-wmf.39'
}
```

So, the mainpage is with space, so we need to replace that with '_'.

This is the sample article on wikipedia:
https://bm.wikipedia.org/wiki/Latino_Amerika

This is the sample API request to fetch the article:
https://bm.wikipedia.org/api/rest_v1/page/mobile-sections/Latino_Amerika

Here is the API documentation where it is mentioned that '_' should be used instead of spaces.
https://en.wikipedia.org/api/rest_v1/#/Mobile/getSections

kelson42: The Mediawiki documentation says at https://www.mediawiki.org/wiki/API:Siteinfo:
> mainpage: The title of the main page, as found in MediaWiki:Mainpage. 1.8+

So this is the title, not the pageid, and in fact it can be customized and it is, see for WPEN:
https://en.wikipedia.org/wiki/MediaWiki:Mainpage

So, here the problem is that you are not looking at the right place IMO. Considering that there is no proper entry in the json. I would recommend to extract it from the `base` URL.

bakshiutkarsha: But, the base will be encoded and look something like this:

`https://bm.wikipedia.org/wiki/Ny%C9%9B_f%C9%94l%C9%94`

So doing a operation to:
1. Extract the last segment from URl and then
2. Decoding it
is much overhead than a simple replace. Please suggest?

kelson42: you need of course to extract the id and url-decode it.

PR #1200, opened 2020-07-03
Fix for 404 pages on fr wiktionary

### LINKED ISSUE #989: Sample custom plugin for wiktionary

bakshiutkarsha:
should resolve #989

### DIFF
--- a/extensions/wiktionary_fr.js
+++ b/extensions/wiktionary_fr.js
@@ -14,9 +14,9 @@ module.exports = class WiktionaryFR { // implements CustomProcessor
 
         for (const link of sectionLinks) {
             if (!articleList.includes(link.innerHTML)) {
-                var span = doc.createElement("span");
+                const span = doc.createElement("span");
                 span.class = "new";
-                var content = doc.createTextNode(link.innerHTML);
+                const content = doc.createTextNode(link.innerHTML);
                 span.appendChild(content)
                 link.replaceWith(span)
             }
--- a/extensions/wiktionary_fr.js
+++ b/extensions/wiktionary_fr.js
@@ -3,10 +3,23 @@ module.exports = class WiktionaryFR { // implements CustomProcessor
         const frenchTitle = doc.querySelector(`#fr.sectionlangue`);
         return !!frenchTitle;
     }
-    async preProcessArticle(articleId, doc) {
+    async preProcessArticle(articleId, doc, articleList) {
         const nonFrenchTitles = Array.from(doc.querySelectorAll(`.sectionlangue:not(#fr)`));
         for (const title of nonFrenchTitles) {
             title.closest('details').remove();
+        }
+
+        const section = doc.querySelector(`.bandeau-voir`);
+        const sectionLinks = Array.from(section.querySelectorAll('a'));
+
+        for (const link of sectionLinks) {
+            if (!articleList.includes(link.innerHTML)) {
+                var span = doc.createElement("span");
+                span.class = "new";
+                var content = doc.createTextNode(link.innerHTML);
+                span.appendChild(content)
+                link.replaceWith(span)
+            }
         }
 
         const h4titles = Array.from(doc.querySelectorAll(`h4`));
--- a/extensions/wiktionary_fr.js
+++ b/extensions/wiktionary_fr.js
@@ -8,7 +8,7 @@ module.exports = class WiktionaryFR { // implements CustomProcessor
         for (const title of nonFrenchTitles) {
             title.closest('details').remove();
         }
-        
+
         const h4titles = Array.from(doc.querySelectorAll(`h4`));
         for (const h4title of h4titles) {
             h4title.closest('details').remove();
--- a/extensions/wiktionary_fr.js
+++ b/extensions/wiktionary_fr.js
@@ -3,25 +3,12 @@ module.exports = class WiktionaryFR { // implements CustomProcessor
         const frenchTitle = doc.querySelector(`#fr.sectionlangue`);
         return !!frenchTitle;
     }
-    async preProcessArticle(articleId, doc, articleList) {
+    async preProcessArticle(articleId, doc) {
         const nonFrenchTitles = Array.from(doc.querySelectorAll(`.sectionlangue:not(#fr)`));
         for (const title of nonFrenchTitles) {
             title.closest('details').remove();
         }
-
-        const section = doc.querySelector(`.bandeau-voir`);
-        const sectionLinks = Array.from(section.querySelectorAll('a'));
-
-        for (const link of sectionLinks) {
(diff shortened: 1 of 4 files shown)


bakshiutkarsha:
> The whole approach around `articleList` seems wrong. Why having an additional list? Articles which won't be mirrored should be removed in the Redis and all the rest should just work fine.

Thanks for the review, the code looks much simpler now.

bakshiutkarsha:
> > The whole approach around `articleList` seems wrong. Why having an additional list? Articles which won't be mirrored should be removed in the Redis and all the rest should just work fine.
> 
> Thanks for the comments, the code looks much simpler now.

extensions/wiktionary_fr.js
```diff
@@ -8,7 +8,7 @@ module.exports = class WiktionaryFR { // implements CustomProcessor
         for (const title of nonFrenchTitles) {
             title.closest('details').remove();
         }
-
+
```
kelson42, line 11: Pleae remove trailing spaces here.

src/util/saveArticles.ts
```diff
@@ -118,26 +118,62 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
...
-            {concurrency: CONCURRENCY_LIMIT}
+            { concurrency: CONCURRENCY_LIMIT }
         );
     } catch (err) {
         logger.log(`Not able download in bulk due to ${err}`);
     }
 }
 
+async function getAllArticlesToKeep(downloader: Downloader, mw: MediaWiki, dump: Dump) {
+    if (!dump.customProcessor) {
```
kelson42, line 129: I would like to see this specific test outside the `getAllArticlesToKeep()` call.
Here, in place, I would like to check if `dump.customProcessor.shouldKeepArticle` function exists at all. We can have a customflavour which does not implement that function.

src/util/saveArticles.ts
```diff
@@ -118,26 +118,62 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
...
+                try {
+                    const rets = await downloader.getArticle(articleId, dump);
+
+                    for (const { articleId, html: articleHtml } of rets) {
+                        if (!articleHtml) {
+                            continue;
+                        }
+
+                        const { articleDoc: _articleDoc } = await processArticleHtml(articleHtml, downloader, mw, dump, articleId);
+                        const articleDoc = _articleDoc;
```
kelson42, line 145: I don't believe this is necessary.

src/util/saveArticles.ts
```diff
@@ -118,26 +118,62 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
...
+            for (const [articleId] of Object.entries(articleKeyValuePairs)) {
+                try {
+                    const rets = await downloader.getArticle(articleId, dump);
+
+                    for (const { articleId, html: articleHtml } of rets) {
+                        if (!articleHtml) {
+                            continue;
+                        }
+
+                        const { articleDoc: _articleDoc } = await processArticleHtml(articleHtml, downloader, mw, dump, articleId);
```
kelson42, line 144: Why we do `processArticleHtml`, seems not necessary here to me.

src/util/saveArticles.ts
```diff
@@ -118,26 +118,62 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
...
+                        const articleDoc = _articleDoc;
+
+                        if (dump.customProcessor?.shouldKeepArticle) {
```
kelson42, line 147: Like I said earlier, this test should be done earlier (only one time).

src/util/saveArticles.ts
```diff
@@ -118,26 +118,62 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
...
+                            const shouldContinue = await dump.customProcessor.shouldKeepArticle(articleId, articleDoc);
+                            if (!shouldContinue) {
+                                await articleDetailXId.delete(articleId);
+                            }
```
kelson42, line 151: There is probably a more elegant manner to code this (in one line). Considering that we don't expect anything from the `delete()` I don't believe this is necessary to have an `await` here.

src/util/saveArticles.ts
bakshiutkarsha: We need _articleDoc to perform operations on it. For eg: `doc.querySelector(`#fr.sectionlangue`);`
 `articleHtml` is a type string, we need to process it in doc to use methods on it.

kelson42: What you need is a DOM. To get the DOM from the HTML, you just need to call `let doc = domino.createDocument(html);` so why calling the most complex function of the whole MWoffliner instead of just calling that line?

PR #1201, opened 2020-07-06
getFullUrl Rewritten

### LINKED ISSUE #1175: Rewrite getFullUrl()
`getFullUrl()` is a function which is called in many places and would benefit of a few improvments:
* Unit test the function properly
* There should not be a typechecking a start. The code should secure it is called properly. I believe we can secure that without additional overhead.
* It should be `mutateToFullUrl(url: URL, const baseUrl: URL)`
* The function should modify the object `url` and return void
* Path part of the URL should not contain `./` or `../`, use https://millermedeiros.github.io/mdoc/examples/node_api/doc/path.html#path.relative for that. (when necessary).
* If `baseUrl` is a complete URL, I don't think we need `webUrlHost` anymore.
* Create an other method which takes strings `getFullUrl(url: string, const baseUrl: string)`.

The goal is piece by piece to deal as much as possible with URL object to avoid parsing all the time the url string (which is a slow operation. So ideally we should tend to parse one time the urls at start to get the objects, then manipulate them, then format them again to write them in the DOM.

bakshiutkarsha:
should resolve #1175

### DIFF
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -46,32 +46,19 @@ export function touch(paths: string[] | string) {
   });
 }
 
-export function 
-getFullUrl(url: string, baseUrl: string) {
-  // if (typeof url !== 'string' || !url) {
-  //   throw new Error(`Expected url to be a string, got [${url}] instead`);
-  // }
+export function mutateToFullUrl(url: string, baseUrl: string) {
   const urlObject = urlParser.parse(url, false, true);
   if (!urlObject.protocol) {
     const baseUrlObject = baseUrl ? urlParser.parse(baseUrl, false, true) : {} as UrlWithStringQuery;
     urlObject.protocol = urlObject.protocol || baseUrlObject.protocol || 'http:';
     urlObject.host = urlObject.host || baseUrlObject.host;
-    
-    urlObject.pathname = "../-/j/js_modules/jsConfigVars.js"
-    // logger.log('BASE---', baseUrlObject.pathname);
-    // logger.log('URL0---', urlObject.pathname);
-    /* Relative path */
-    if (urlObject.pathname && urlObject.pathname.indexOf('/') !== 0 && baseUrlObject.pathname) {
-      // urlObject.pathname = `${pathParser.dirname(baseUrlObject.pathname)}/${urlObject.pathname}`;
-      logger.log('BASE---', baseUrlObject.pathname);
-      urlObject.pathname = pathParser.relative(baseUrlObject.pathname, urlObject.pathname)
-      logger.log('RELATICE', urlObject.pathname)
-    }
-    
-    url = urlParser.format(urlObject);
-    logger.log(url)
-  }
 
+  /* Relative path */
+    if (urlObject.pathname && urlObject.pathname.indexOf('/') !== 0 && baseUrlObject.pathname) {
+      urlObject.pathname = pathParser.relative(baseUrlObject.pathname, urlObject.pathname)
+    }
+    url = urlParser.format(urlObject);
+  }
   return url;
 }
 
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -46,22 +46,30 @@ export function touch(paths: string[] | string) {
   });
 }
 
-export function getFullUrl(webUrlHost: string, url: string, baseUrl: string) {
-  if (typeof url !== 'string' || !url) {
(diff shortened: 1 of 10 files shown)


src/util/misc.ts
```diff
@@ -46,27 +46,26 @@ export function touch(paths: string[] | string) {
   });
 }
 
-export function getFullUrl(webUrlHost: string, url: string, baseUrl: string) {
-  if (typeof url !== 'string' || !url) {
-    throw new Error(`Expected url to be a string, got [${url}] instead`);
-  }
+export function getFullUrl(url: string, baseUrl: string) {
   const urlObject = urlParser.parse(url, false, true);
   if (!urlObject.protocol) {
```
kelson42, line 51: We need a small commnet here to explain that there is nothing to do if we have a full URL already.

src/util/misc.ts
```diff
@@ -46,27 +46,26 @@ export function touch(paths: string[] | string) {
...
-
+    mutateToFullUrl(urlObject, baseUrl);
     url = urlParser.format(urlObject);
   }
-
   return url;
 }
 
+export function mutateToFullUrl(urlObject: any, baseUrl: string){
+  const baseUrlObject = baseUrl ? urlParser.parse(baseUrl, false, true) : {} as UrlWithStringQuery;
```
kelson42, line 59: We should make a small test here to test `urlObject.protocol` and just return if already test.

src/util/misc.ts
```diff
@@ -46,27 +46,26 @@ export function touch(paths: string[] | string) {
...
-    }
-
+    mutateToFullUrl(urlObject, baseUrl);
     url = urlParser.format(urlObject);
   }
-
   return url;
 }
 
+export function mutateToFullUrl(urlObject: any, baseUrl: string){
```
kelson42, line 58: * We should always have a non null `baseUrl`
* `baseUrl` should be a URL object
* Please keep the `url` and `baseUrl` naming. I'm not really in favour to put the type a variable in its name

src/util/misc.ts
```diff
@@ -46,27 +46,26 @@ export function touch(paths: string[] | string) {
...
+  const baseUrlObject = baseUrl ? urlParser.parse(baseUrl, false, true) : {} as UrlWithStringQuery;
+  urlObject.protocol = urlObject.protocol || baseUrlObject.protocol || 'http:';
+  urlObject.host = urlObject.host || baseUrlObject.host;
+
+  /* Relative path */
+  if (urlObject.pathname && urlObject.pathname.indexOf('/') !== 0 && baseUrlObject.pathname) {
+    urlObject.pathname = pathParser.relative(baseUrlObject.pathname, urlObject.pathname)
+  }
+}
+
```
kelson42, line 68: Thinking twice about it, I believe we could do even simplier with `url.resolve()`.

src/util/misc.ts
```diff
@@ -46,27 +46,26 @@ export function touch(paths: string[] | string) {
   });
 }
 
-export function getFullUrl(webUrlHost: string, url: string, baseUrl: string) {
-  if (typeof url !== 'string' || !url) {
-    throw new Error(`Expected url to be a string, got [${url}] instead`);
-  }
+export function getFullUrl(url: string, baseUrl: string) {
```
kelson42, line 49: `baseUrl` should be a URL object. Considering that is is almost always caleed with `mw.base` that should save us the parsing. But that mean you need to update `Mediawiki.base`. To me, `Mediawiki.base` should be called `Mediawiki.baseUrl` and should be an object. I don't really understand why we don't have moved in to URL object like the others?!

src/util/misc.ts
bakshiutkarsha: Now we are dealing with Url objects, it might be extra work to first convert it to string, as `url.resolve()` returns string(also accepts string parameter), and then convert it back to url object.

bakshiutkarsha:
> I discovered that you use a deprecated `Url` object in place of `URL`, see https://nodejs.org/api/url.html#url_legacy_url_api. Please do the replacement.

`Url` object is been used previously as well in the project as well, and if I will start changing the type, it will lead to a lot of changes because other url objects like(webUrl, apiUrl, veApiUrl, restApiUrl etc) has the same `Url `object type for now. So can we have a separate ticket for this change?

kelson42:
I believe this is a relatively easy fix and that it will be difficult to do this PR properly without doing that. Please try.

PR #1225, opened 2020-07-21
Fixing the broken links and test cases

### LINKED ISSUE #1203: Totally broken link in stupidepedia
See for example http://library.kiwix.org/stupidedia_de_all_maxi_2020-07/A/Stern, all the links which start with `/wiki/`.

bakshiutkarsha:
should resolve #1203

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -118,12 +118,10 @@ class MediaWiki {
 
   public extractPageTitleFromHref(href: any) {
     try {
-      const pathname = urlParser.parse(href, false, true).pathname || '';
-      if (pathname.indexOf('./') === 0) {
-        return util.decodeURIComponent(pathname.substr(2));
-      }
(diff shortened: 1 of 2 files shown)


bakshiutkarsha:
The `webUrl.pathname` is formed from `wikiPath`, and by default wikiPath is `/wiki`, but if we specify some other wikiPath, eg `index.php`, the condition `pathname.indexOf(this.webUrl.pathname) === 0` fails, due to which this error comes for all the wikis where wikiPath is specified.

In this PR, I have resolved by replacing the above condition and checking  whether `/wiki` comes in path or not. Please let me know if that is the right approach.

kelson42:
> the condition `pathname.indexOf(this.webUrl.pathname) === 0` fails

This is the bug. It should not.

kelson42:
Superseeded by #1291

PR #1237, opened 2020-07-26
Javascript API for toc

### LINKED ISSUE #596: Introduce the ZIM JS API
See https://wiki.openzim.org/wiki/Javascript_API (still a draft)

bakshiutkarsha:
should resolve #596


kelson42:
@Jaifroid @automactic @macgills This is the really early iteration to create a JS API. I have chosen the TOC API because it is I believe the most simple thing. I would like to have you to join this journey so we secure what is done here is helpful/usable to Kiwix reader' devevelopers. The first thing I would like you to secure with @bakshiutkarsha is that the implementation is correct and work at least as your one in your respective readers.
@mgautierfr From you I would like that you help to decide where to save this (or these) content in the ZIM file. The question of the API to access the content is as well still open.

Jaifroid:
@kelson42 @bakshiutkarsha Happy to troubleshoot/test from the Kiwix JS perspective. Just looking at the code, as this is intended to run in any supported browser/environment, the code should be re-cast as ES5. We need to support IE11 as a minimum, and 'async' won't work in older browsers.

I don't know the API spec, but as far as I can see, none of the code inside the async blocks actually runs asynchronously, so there is nothing to wait for. I think the same code can simply be written as a set of functions.

[Adding @mossroy to this thread.]

kelson42:
@Jaifroid we can define what ever we need as API, this is the coal of the PR discussion around TOC.

Jaifroid:
OK, thanks! In that case, we'll also need to polyfill [Element.closest()](https://dev.caniuse.com/element-closest).

Jaifroid:
Polyfill here: https://github.com/jonathantneal/closest 

Also, IE11 can't use forEach on a querySelector node list. Instead you have to use the `Array.prototype.slice.call(list).forEach()` syntax or else `[].slice.call(list).forEach()`.

Jaifroid:
Finally, I notice this uses `module.exports`, which could be an issue depending on whether you're expecting the browser to import the code, or you're going to ship it bundled with webpack or similar. And if you're going to ship it bundled, then babel will handle the polyfills needed for IE11 and other browsers, so my comments above wouldn't apply in that case.

mgautierfr:
I will say nothing on the js code, I'm not a expert here. (Except that I'm stuck in a js world where `async` and `module` were not existing, old browsers are probably stuck with me so we must be careful about the js we use)

To be complete here, do we agree that :
- The js code will be loaded "somehow" (to define) in the page.
- Client applications (android, ios, desktop) will call functions to get the toc or move to the right position.

I have few questions :

Should we use a js method to change the location ? If a client can get the whole toc of the archive, it could change the location of the page directly (a probably better than relying on a embedded js).

The following script has the advantage to be article independent, the TOC itself is inferred from the html content. But we could also standardize the html content to allow the client to found the TOC itself (not sure it is a good idea, but for brainstorming purpose, it is good to have it in mind).
We could simply add a json dict in each article with the TOC, client would simply have to get it.
(This json dict could be generated with a independent script, but the "api" would be the dict, not the script).

It worth it to compare this with the comment https://github.com/kiwix/kiwix-lib/issues/322#issuecomment-584225014

kelson42:
>     * The js code will be loaded "somehow" (to define) in the page.

Yes, this is the idea as the reader has control over the HTML widget/DOM. The operation should be as easy as possible and we will need a solution at libzim/libkiwix to make the retrieval of this JS API code as easy/clean as possible.

>     * Client applications (android, ios, desktop) will call functions to get the toc or move to the right position.

Yes, anything after retrieval of JS API code is full in hand of ZIM reader.

> Should we use a js method to change the location ? If a client can get the whole toc of the archive, it could change the location of the page directly (a probably better than relying on a embedded js).

I have not clear opinion on that.

> The following script has the advantage to be article independent, the TOC itself is inferred from the html content. But we could also standardize the html content to allow the client to found the TOC itself (not sure it is a good idea, but for brainstorming purpose, it is good to have it in mind).

This is what happens for the moment.

The JS API allows to:
* Not have to reimplement again & again the same logics in readers
* Allow ZIM creator to superseed default implementation and do things differently

> We could simply add a json dict in each article with the TOC, client would simply have to get it.
> (This json dict could be generated with a independent script, but the "api" would be the dict, not the script).

This is how EPUB works. The JS API is a more flexible/leaner approach to achieve the same result.

mgautierfr:
> The JS API allows to:
> - Not have to reimplement again & again the same logics in readers
> - Allow ZIM creator to superseed default implementation and do things differently

Yes, I agree. I would put the js code in the zim file itself. And make article add it in the "normal" way.
(Or not have js at all and simply hard code the json dict in the article).

> This is how EPUB works. The JS API is a more flexible/leaner approach to achieve the same result.

How it is better ?

kelson42:
> > This is how EPUB works. The JS API is a more flexible/leaner approach to achieve the same result.
> 
> How it is better ?

* Integrates within the standard JS API (no other concept)
* No additional storage in 99% case
* Nothing to do in normal case fore the scraper
* Ability to handle special cases with only a little storage cost

kelson42:
> Yes, I agree. I would put the js code in the zim file itself. And make article add it in the "normal" way.
> (Or not have js at all and simply hard code the json dict in the article).

The JS API is in each/the ZIM.

mgautierfr:
If it is js that generate the dict :

> Integrates within the standard JS API (no other concept)

Which one ? We are currently defining it. We could define that a dict have to be present.

> No additional storage in 99% case

The js could generate the dict at runtime (read time).

> Nothing to do in normal case fore the scraper

If this is to the js to generate it, no more work for the scraper, at least not more than add a js with a specific api in the zim file wi


> Ability to handle special cases with only a little storage cost

Same if it is the js creating the dict.
For really specific case, we could store directly the dict without have to create a one time script.

Jaifroid:
I agree generating the TOC dynamically from the article headings, rather than having it as an extra json item is the simplest generic approach that doesn't require any extra storage.

Apart from that, I think it's important to decide now how the script gets attached to the page. Since MWoffliner is written in Typescript, I believe, you probably have a toolchain that compiles TS down to JS for Node. A similar toolchain can be used to compile ES6 down to ES5 and below, and automatically bundle the needed polyfills for all API scripts as a single optimized JS file - e.g. `kiwixAPI.js`. Webpack + babel will do this.

Is this what's intended? It would seem the most sensible approach, as it will support even IE11.

(It would be great not to have to bundle jQuery in the ZIM as at present! It's redundant with babel.)

bakshiutkarsha:
@kelson42 @Jaifroid  For now, I have exposed the class in the `window` object, so anyone can access it from there.

Also, I am bundling the file after Zim gets created, we can separate that out as well. Let me know your thoughts about that?

bakshiutkarsha:
> @kelson42 @Jaifroid For now, I have exposed the `TableOfContent` class in the `window` object, so anyone can access it from there.
> 
> Also, I am bundling the file after Zim gets created, we can separate that out as well. Let me know your thoughts about that?

kelson42:
> @kelson42 @Jaifroid For now, I have exposed the `TableOfContent` class in the `window` object, so anyone can access it from there.
> 
> Also, I am bundling the `kiwixAPI.js` file after Zim gets created, we can separate that out as well. Let me know your thoughts on that?

@bakshiutkarsha There is many things which I would like to change with your PR:
* Your PR makes many assumptions about the content. I understand you make this PR based on a Mediawiki Content, but what you are building here is a generic JS API which should as far as possible with any HTML. Relying on `data-level` attribute is not possible for example
* I don't think we want to have an object here, except if there is an excellent to make it statefull. It should be stateless. The array returns should contained the CSS selector to find again the node. That way it should be possible to have a stateless solution.
* I want all the function in the namespace "zim::toc"
* Your code should work will all kinds of `<H>` nodes.. not only `<h2>`
* I hardly see how `import` could work in that context. @Jaifroid Do you have any idea to have a good system to handle dependencies? The user should typically load within the DOM the needed API only with one call!

kelson42:
I would like to have the TOC JS API in `/-/_api/toc,js`

res/kiwixAPI.js
```diff
@@ -0,0 +1,43 @@
...
+        if (headerObject.id === "") {
+            headerObject.id = "documentparserid" + i;
+        }
+        sectionList.push(headerObject);
+    }
+    return sectionList;
+}
+
+zim.toc.hasTableOfContent = function() {
+    return getSectionList().length > 0 ? true : false;
```
mgautierfr, line 19: This call (complex) `getSectionList` just to know if a `h*` tag is present.
You can directly test the value of `document.querySelectorAll('h1, h2, h3, h4, h5, h6').length`. Or even better test that on header tag is present without getting all header tags with specific js api I don't know.

res/kiwixAPI.js
```diff
@@ -0,0 +1,43 @@
...
+function getSectionList(){
+    let sectionList = [];
+    for (i = 0; i < document.querySelectorAll('h1, h2, h3, h4, h5, h6').length; i++) {
+        let headerObject = document.querySelectorAll('h1, h2, h3, h4, h5, h6')[i];
+        if (headerObject.id === "") {
+            headerObject.id = "documentparserid" + i;
+        }
+        sectionList.push(headerObject);
+    }
+    return sectionList;
```
mgautierfr, line 15: This function modify the content of the dom in the same time it construct a list.
And the final list is equal to `document.querySelectorAll('h1, h2, h3, h4, h5, h6')`.

We can split this in two functions, one run once at document loading to add the `id`, and one just returning `document.querySelectorAll('h1, h2, h3, h4, h5, h6')` when needed.

res/kiwixAPI.js
```diff
@@ -0,0 +1,43 @@
...
+            "section_id": section.id,
+            "section_name": section.innerText,
+        })
+    });
+    return respArrOfSections;
+}
+
+zim.toc.scrollToSection = function(index){
+    const sectionList = getSectionList();
+    const sectionIdElem = document.getElementById(sectionList[index].id);
```
mgautierfr, line 37: We can probably generate the `id` to query (`"documentparserid" + index`) and directly use it to get the `sectionIdElem` (`const sectionIdElem = document.getElementById("documentparserid"+index);`)

This way we don't need to get the section list.

---

And by the way, `sectionList[index]` is already the elem returned by `document.getElementById(secttion[index].id)`.

res/kiwixAPI.js
```diff
@@ -0,0 +1,43 @@
...
+    let respArrOfSections = [];
+    const sectionList = getSectionList()
+    sectionList.forEach(section => {
+        respArrOfSections.push({
+            "toc_level": section.tagName,
+            "section_id": section.id,
+            "section_name": section.innerText,
+        })
+    });
+    return respArrOfSections;
```
mgautierfr, line 32: We may want to store this result in the `zim` (or `zim.toc`) object.
This way we don't have to generate it every time.

mgautierfr (changes requested):

res/kiwixAPI.js
bakshiutkarsha: But then it won't remain stateless I think. I made it like that on purpose after review comments. We don't want that dependency of one run to save the state of the sections for any API method, each API method could be called independently with the use of `getSections()`

Also, what I understood from the discussions, this file should be having no control over what is happening during the load of the DOM. It is just there to serve as the API purpose once any other application uses it.

res/kiwixAPI.js
bakshiutkarsha: `const sectionIdElem = document.getElementById("documentparserid"+index)` won't work in this case as we only give the `documentparserid` in the cases where id is not already present.

So, I used `sectionList[index]`

mgautierfr: Indeed.
So we can simply do `const sectionIdElem = getSectionList()[index];` I think.

res/kiwixAPI.js
mgautierfr: > But then it won't remain stateless I think.

Well this is already not stateless as you are modifying the dom.

I think that what we want is that user (kiwix readers) don't need to call one method to setup and another ones to use. Just methods to use the toc and don't care about setup.
This is what is done currently (and I agree with the principle). But we don't need to (internally) do the setup all the time. We can do it only once. Splitting the code into one function doing the setup (modify the dom, maybe create the array of sections  and put it in `zim.toc` object) and other method accessing the object (internally calling the setup function if needed) is better.

Jaifroid:
Can I also reiterate that the API needs to support older browsers (concretely, IE11), which means that it needs to be re-cast as ES5, or transpiled by Babel in the toolchain. Is this to be addressed at compile time?

bakshiutkarsha:
> Can I also reiterate that the API needs to support older browsers (concretely, IE11), which means that it needs to be re-cast as ES5, or transpiled by Babel in the toolchain. Is this to be addressed at compile time?

Right now, it is getting compiled after the MWOffliner run is completed. I still have to set that bundle file path to `/-/_api/toc.js`, will be done in my next commit.

Do you suggest some other time to compile?

kelson42:
@mgautierfr I can not comment on your comment for some reason I ignore, but I specifically ask to not store the TOC in its class. So no setup. Just set an `id` if missing one. This is the simpliest an most robust solution.

bakshiutkarsha:
@kelson42 The file path is coming on the desired location now:

toc

Jaifroid:
> * I hardly see how `import` could work in that context. @Jaifroid Do you have any idea to have a good system to handle dependencies? The user should typically load within the DOM the needed API only with one call!

Sorry for very late reply to this. `Import` can work if the module is transpiled and bundled by [webpack + babel](https://monkelite.com/what-are-babel-and-webpack/). It then handles attaching dependencies internally, when they are required by the user. TypeScript may also have its own tools to do something similar, but I'm not an expert in that.

Jaifroid:
> * I understand you make this PR based on a Mediawiki Content, but what you are building here is a generic JS API which should as far as possible with any HTML. Relying on `data-level` attribute is not possible for example

@kelson42 Perhaps you can clarify one thing here. This API is being built as part of mwoffliner, which is MediaWiki-specific. Other scrapers have very different codebases, I understand. A ToC will only be relevant for Encyclopaedia-like articles, or other reference works, so I wonder if it is not inevitable that the ToC API will have to be specific to each ZIM format. Headings are not always produced via H1 or H2 tags any more (unfortunately), but in some contexts may be simply wrapped in `<div>`  tags with a `data-level` or `heading` class, making it nigh-on impossible to write generic code to extract the ToC. What is your view on this?

Jaifroid:
@bakshiutkarsha Is it possible to make a test ZIM with a small number of articles in, so that we can test the implementation and help troubleshoot with older browsers?

mgautierfr:
> I specifically ask to not store the TOC in its class. So no setup.

I accept that :)

> Just set an id if missing one.

This is a kind of setup.

---

The real point (for me) is not the code but the public api :
- hasTableOfContent() -> bool
- getSections() -> list of sections {"toc_level", "section_id", "ection_name"}
- scrollToSection(index)

I wonder if the scrollToSection should take a "id" instead of a "index". It would be more useful as the section id is already in the returned information of `getSections`.
If we need a index, the `section_id` is useless and user code has to keep a index with the returned section dict to be able to scroll to the section.

bakshiutkarsha:
> @bakshiutkarsha Is it possible to make a test ZIM with a small number of articles in, so that we can test the implementation and help troubleshoot with older browsers?

Yes, we can run a particular wiki with a parameter of  `--articlelist` where we can specify some articles from that wiki, and it will create the zim with those particular articles.

I tried to attach the file here, but it seems zim format file type is not supported by github.

Jaifroid:
> I tried to attach the file here, but it seems zim format file type is not supported by github.

Thank you, @bakshiutkarsha . To make the ZIM file available, assuming it's not too large, you could create a new branch, add the file to the repo somewhere, push it, and let me know the branch name. Alternatively, you could attach it to a new [release](https://github.com/openzim/mwoffliner/releases) -- the release can be deleted after we've finished with it. Or else feel free to use onedrive / gdrive / dropbox and post the link. In the past we've used [Framadrop](https://covid19.framadrop.org/en/) for this, but it looks like it is closing down 😥.

bakshiutkarsha:
> > I tried to attach the file here, but it seems zim format file type is not supported by github.
> 
> Thank you, @bakshiutkarsha . To make the ZIM file available, assuming it's not too large, you could create a new branch, add the file to the repo somewhere, push it, and let me know the branch name. Alternatively, you could attach it to a new [release](https://github.com/openzim/mwoffliner/releases) -- the release can be deleted after we've finished with it. Or else feel free to use onedrive / gdrive / dropbox and post the link. In the past we've used [Framadrop](https://covid19.framadrop.org/en/) for this, but it looks like it is closing down 😥.

Here is the link, I uploaded it in drive: [ZIM FILE](https://drive.google.com/file/d/1IP4PiY_Wp96nPTwrnDlqQTctE-Ywj85O/view?usp=sharing)

Jaifroid:
Thank you, @bakshiutkarsha . I've downloaded that and have run it through Kiwix JS.
The console dev tools tell me that the following JS scripts are attached to the one article in this ZIM (see list on left of screenshot):

image
 
I'm having trouble locating which of these scripts loads the `kiwixAPI.js` module. They mostly look like mediawiki scripts, except for a few legacy mwoffliner scripts that open and close all sections, etc. Can you see the relevant script for the API (or a module loader script) here?

bakshiutkarsha:
> @kelson42 The file path is coming on the desired location now:
> 
> toc

@Jaifroid  This is the location for the file as suggested by @kelson42

Jaifroid:
OK, I see -- so the idea is that the client software (reader) must look for and load/run the script, it is not attached to articles like the old API you see here (e.g. details-element polyfill listed here)?

In that case, I'll need to write a bit of code to attach the script in order to test it properly. I'll try to do that in the next day or two and get back to you, as it's getting late now in UK!

.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
bakshiutkarsha: It's needed.

Jaifroid:
@bakshiutkarsha I'm trying to test your ZIM but cannot find the relevant file. I am using this code to load the script:
```
           // Insert Kiwix API script
            var script = innerDoc.createElement('script');
            script.type = 'text/javascript';
            script.src = '../-/_api/toc.js';    
            innerDoc.getElementsByTagName('head')[0].appendChild(script);
```
However this now gives me the console.log error:

image

Am I looking for the right file? Or is it bundled in the final ZIM under a different name? (We mention above `kiwixAPI.js` as a possibility for the full API bundled with webpack.)

bakshiutkarsha:
> @bakshiutkarsha I'm trying to test your ZIM but cannot find the relevant file. I am using this code to load the script:
> 
> ```
>            // Insert Kiwix API script
>             var script = innerDoc.createElement('script');
>             script.type = 'text/javascript';
>             script.src = '../-/_api/toc.js';    
>             innerDoc.getElementsByTagName('head')[0].appendChild(script);
> ```
> 
> However this now gives me the console.log error:
> 
> image
> 
> Am I looking for the right file? Or is it bundled in the final ZIM under a different name? (We mention above `kiwixAPI.js` as a possibility for the full API bundled with webpack.)
> 
> EDIT: I also can't find `-/kiwixAPI.js` or `-/_api/kiwixAPI.js` in the ZIM you attached above.

Sorry for the delayed response. I double-checked on my side and the file is getting generated on the mentioned location of `-/_api/toc.js`. 

I know we previously decided on `kiwixAPI` name, but after recent comments by @kelson42 I have changed the names and file location accordingly. So, in Mwoffliner the file is created under `zim` folder named `toc.js` and when final zim is created it comes under the above mentioned path.

Jaifroid:
Thanks @bakshiutkarsha . I'll test with your latest ZIM and get back to you.

Jaifroid:
@bakshiutkarsha I confirm the file is now in the new ZIM you uploaded, and I can attach it using the code below. However, once loaded, I have no clue how to use it. Where should I look for the defined functions? Are they under a global object? In which scope? To be clear, I can't find, e.g. `zim.hasTableOfContents()`, but I may be looking in the wrong place.

image

bakshiutkarsha:
> @bakshiutkarsha I confirm the file is now in the new ZIM you uploaded, and I can attach it using the code below. However, once loaded, I have no clue how to use it. Where should I look for the defined functions? Are they under a global object? In which scope? To be clear, I can't find, e.g. `zim.hasTableOfContents()`, but I may be looking in the wrong place.
> 
> image

Everything in the namespace of `zim.toc`, so the function call would be `zim.toc.hasTableOfContents()`, and I have exported the zim object like [this](https://github.com/openzim/mwoffliner/pull/1237/files#diff-97552c81824c240c1f61478d5572ae76R41)

bakshiutkarsha:
> > @bakshiutkarsha I confirm the file is now in the new ZIM you uploaded, and I can attach it using the code below. However, once loaded, I have no clue how to use it. Where should I look for the defined functions? Are they under a global object? In which scope? To be clear, I can't find, e.g. `zim.hasTableOfContents()`, but I may be looking in the wrong place.
> > image
> 
> Everything is in the namespace of `zim.toc`, so the function call would be `zim.toc.hasTableOfContents()`, and I have exported the zim object like [this](https://github.com/openzim/mwoffliner/pull/1237/files#diff-97552c81824c240c1f61478d5572ae76R41). Can you confirm that you are able to find the `zim` object only?

kelson42:
@Jaifroid Have you been able to get it working? @bakshiutkarsha is finishing the her GSoC and we would like to merge her work (if it works for you).

Jaifroid:
> Everything is in the namespace of `zim.toc`, so the function call would be `zim.toc.hasTableOfContents()`, and I have exported the zim object like [this](https://github.com/openzim/mwoffliner/pull/1237/files#diff-97552c81824c240c1f61478d5572ae76R41). Can you confirm that you are able to find the `zim` object only?

OK I'll check now.

Jaifroid:
Hmm. @bakshiutkarsha it doesn't seem to be getting exported at least on load of the script. See underlined / circled code and console log output in image. Maybe we need to wait for the export to happen? I could try with a timeout. Or do you have other ideas? This code is running in Chromium 85.0, so it shouldn't be a browser incompatibility.

image

Jaifroid:
@bakshiutkarsha If you would like to try the code, please use this branch of Kiwix JS Windows:

https://github.com/kiwix/kiwix-js-windows/tree/Test-ToC-API 

Or suggest modifications to the loading code, and I can try.

To test for yourself, you just need to spin up a localhost web server to run the code. Load up app.js and set breakpoint on line 3671 NB because it's a PWA, it caches all code for offline use, so when you make a change you must go to the Application tab in dev tools, unregister pwabuilder-sw.js and then force reload with Ctrl-Shift-R (in Chrome / Edge Chromium).

bakshiutkarsha:
> Hmm. @bakshiutkarsha it doesn't seem to be getting exported at least on load of the script. See underlined / circled code and console log output in image. Maybe we need to wait for the export to happen? I could try with a timeout. Or do you have other ideas? This code is running in Chromium 85.0, so it shouldn't be a browser incompatibility.
> 
> image

Let me do some changes around the exporting of zim object on my side of the code.

Jaifroid:
Just to say I tried with code below (timeout of 500ms), but no luck.
```
            script.addEventListener('load', function() {
                setTimeout(function() {
                    console.log('Testing for presence of ToC API:');
                    console.log(zim);
                    console.log(zim.toc);
                    console.log(zim.toc.hasTableOfContents());
                }, 500);
            });
```

bakshiutkarsha:
> Just to say I tried with code below (timeout of 500ms), but no luck.
> 
> ```
>             script.addEventListener('load', function() {
>                 setTimeout(function() {
>                     console.log('Testing for presence of ToC API:');
>                     console.log(zim);
>                     console.log(zim.toc);
>                     console.log(zim.toc.hasTableOfContents());
>                 }, 500);
>             });
> ```

I have exported the object globally now, you will be able to find it now. I checked from my side as well. 
Here is the zim [ZIM](https://drive.google.com/file/d/1rb-Vg5BiTgwpsRiviYHTXpWjxMOQZA0o/view?usp=sharing)

Jaifroid:
@bakshiutkarsha I've just downloaded and tested the new ZIM, and I'm still getting the same error. Did you test it in the browser? It's quite easy: just sync the repo and switch to branch `Test-ToC API`, and run index.html from localhost.

image

bakshiutkarsha:
> @bakshiutkarsha I've just downloaded and tested the new ZIM, and I'm still getting the same error. Did you test it in the browser? It's quite easy: just sync the repo and switch to branch `Test-ToC API`, and run index.html from localhost.
> 
> image

For me in that project, it is not even finding the file, how do I run it properly to get the file?
Screenshot 2020-09-03 at 12 24 46 PM

bakshiutkarsha:
@Jaifroid  That is how I have tested in my local, by putting the whole file on console and trying to get the `zim` obj from it. I know it is not the same how things are working in the kiwix-js, but **earlier** the `zim` Object was somehow not getting exposed from the bundle file for me as well, after exposing it to window Object, it started working.
Screenshot 2020-09-03 at 12 41 53 PM

Jaifroid:
> For me in that project, it is not even finding the file, is there something wrong the way I am running it?
> Screenshot 2020-09-03 at 12 24 46 PM

@bakshiutkarsha Did you switch to the `Test-ToC-API` branch? In that branch I wrote a bit of code to attach the file. The file is defnitiely attached, as you see from my console.log, but when onload event fires (it wouldn't fire if the script were not loaded!) no JS appears to do any of the object exporting. NB If you do switch to that branch, be sure to go to Applications tab of devtools, unregister pwabuilder-sw.js and force refresh.

Jaifroid:
> That is how I have tested in my local, by putting the whole file on console and trying to get the zim obj from it. I know it is not the same how things are working in the kiwix-js, but earlier the zim Object was somehow not getting exposed from the bundle file for me as well, after exposing it to window Object, it started working.

What are the precise steps to reproduce this with Kiwix JS (ideally programmatically)? I don't really understand "putting the whole file on console and trying to get the zim obj from it". It's good progress if you've managed to make it run, but we need to be able to do it in code once the script is found and attached.

Jaifroid:
Also, what precisely do you mean by "exposing it to the window Object"? Do you mean attaching it like: `window.zim = something`? I was attaching to the document, becuase that is how scripts are usually loaded by HTML, but we can probably load it in the DOM using another method than creating a script element in the html head.

bakshiutkarsha:
> > For me in that project, it is not even finding the file, is there something wrong the way I am running it?
> > Screenshot 2020-09-03 at 12 24 46 PM
> 
> @bakshiutkarsha Did you switch to the `Test-ToC-API` branch? In that branch I wrote a bit of code to attach the file. The file is defnitiely attached, as you see from my console.log, but when onload event fires (it wouldn't fire if the script were not loaded!) no JS appears to do any of the object exporting. NB If you do switch to that branch, be sure to go to Applications tab of devtools, unregister pwabuilder-sw.js and force refresh.

Yes my branch is `Test-ToC-API`, but no luck in loading the file after unregistering the force refresh as well

Jaifroid:
Very strange? You are using the latest ZIM with the file in it?
But no problem, we can easily write a bit of code on a branch in Kiwix JS (which doesn't have the SW caching, so easier to be sure we're both running the same code). What do you suggest? This is what I used:

```
            // Insert kiwxAPI.js
            var script = innerDoc.createElement('script');
            script.type = 'text/javascript';
            script.src = '../-/_api/toc.js';
            script.addEventListener('load', function() {
                console.log('Testing for presence of ToC API:');
                console.log(zim);
                console.log(zim.toc);
                console.log(zim.toc.hasTableOfContents());
            });
            innerDoc.getElementsByTagName('head')[0].appendChild(script);

```

bakshiutkarsha:
> Very strange? You are using the latest ZIM with the file in it?
> But no problem, we can easily write a bit of code on a branch in Kiwix JS (which doesn't have the SW caching, so easier to be sure we're both running the same code). What do you suggest? This is what I used:
> 
> ```
>             // Insert kiwxAPI.js
>             var script = innerDoc.createElement('script');
>             script.type = 'text/javascript';
>             script.src = '../-/_api/toc.js';
>             script.addEventListener('load', function() {
>                 console.log('Testing for presence of ToC API:');
>                 console.log(zim);
>                 console.log(zim.toc);
>                 console.log(zim.toc.hasTableOfContents());
>             });
>             innerDoc.getElementsByTagName('head')[0].appendChild(script);
> ```

Yes, I can see this code in my branch, even that's the line where the error occurs:

`innerDoc.getElementsByTagName('head')[0].appendChild(script);`

It would be great if you can do that I can switch to that and debug.

Jaifroid:
@bakshiutkarsha I think I know what's going wrong. I should have explained you need to switch to Service Worker mode in the Configuration. Apologies! If you are running it in jQuery mode then no js at all gets fetched from the ZIM. In Kiwix JS Windows, please go to Configuration and you'll see the switch at the bottom of the page. (Same for Kiwix JS, by the way, but I'll make a branch there shortly.)

bakshiutkarsha:
> @bakshiutkarsha I think I know what's going wrong. I should have explained you need to switch to Service Worker mode in the Configuration. Apologies! If you are running it in jQuery mode then no js at all gets fetched from the ZIM. In Kiwix JS Windows, please go to Configuration and you'll see the switch at the bottom of the page. (Same for Kiwix JS, by the way, but I'll make a branch there shortly.)

It is working now, thanks :)
I don't think now there is a need for another branch.

Jaifroid:
OK!

bakshiutkarsha:
Screenshot 2020-09-03 at 5 42 06 PM

I was able to get the zim obj now with this piece of code:

```
script.addEventListener('load', function() {
                const articleIframeWindow = document.getElementById("articleContent").contentWindow;
                console.log(articleIframeWindow.zim);
                console.log(articleIframeWindow.zim.toc);
                console.log(articleIframeWindow.zim.toc.hasTableOfContent());
            });
```
Let me know your thoughts on this @Jaifroid

Jaifroid:
That's excellent news, @bakshiutkarsha! I'll add it -- or similar -- to that branch and give it a test.

Jaifroid:
@bakshiutkarsha I confirm that the object can be found - well done! 😊

Using my original code (quoted above), it can also be found on 

`window.frames[0].frameElement.contentWindow.zim`

or using the code you supplied. We should probably document this somewhere (in the code?). @kelson42 How do you want to document the API's usage? The API is available even after page load has finished, not only within the onload function.

Here is the result of running one of the functions on a page without a ToC:

image

(it correctly returns `false`).

Here is the result of running functions on the article with a ToC:

image

As you see, here the `getSections()` function fails with the error message `i is not defined`. As I assume i is an iterator, you probably forgot `var` or `let` in front of it? (We use strict mode, so all variables must be defined.)

I also have a suggestion to rename `hasTableOfContent()` to `hasTableOfContents()` (plural) because that is a the conventional name for such a table: a Table of Contents.

Finally, I need to test the code on IE11, but that involves extracting the files specifically because IE11 cannot run in Service Worker mode, so JS is not **automatically** extracted. We can easily solve that for this specific file / API on our end, but I would like to test that the babel/corejs transpiling is working well.

Jaifroid:
@bakshiutkarsha Success with IE11! I've pushed code to the `Test-ToC-API` branch of Kiwix JS Windows that attaches the API in any browser running Kiwix JS, in jQuery as well as SW modes. Here it is running in IE11. IE11 DevTools also confirms the nature of the error with the `getSections()` function (the variable `i` ineeds to be defined in strict mode).

image

But it's great news that the API is working (apart from that strict mode error) in IE11. Congratulations -- I know it's tricky code to work with webpack / corejs / babel!

kelson42:
@bakshiutkarsha @Jaifroid So fixing this last small problem and then merging?

Jaifroid:
@kelson42 Nearly there: once that problem is fixed we need to test getting the section headings and the scroll to heading function, but that will hopefully be a quick test.

Where should the usage of the API be documented? Should there be some documentation in the code (though it will probably never be seen by those who would use it if only in mwoflliner code)? Or would it go in the openzim documentation? It would be helpful to developers to document a generic method of attaching the API file.

@bakshiutkarsha When you do another ZIM with the corrections, maybe choose a different article or selection of articles (that Nazi general makes me feel a bit queasy... 😮 ^_^).

bakshiutkarsha:
@Jaifroid  That's great news, I am glad it is working for old browsers as well. I have done the changes in this new zim file [Zim](https://drive.google.com/file/d/1GtLnKMcivdzYpyXePdxnULzTbIrHltFU/view?usp=sharing), after you are done testing I can push my final changes too. 

Also, I have changed the Nazi general article to someone exactly opposite, hope he won't make you queasy :P  :)

Jaifroid:
Thanks @bakshiutkarsha. The link for the new ZIM won't let me in -- says I need to request access from you (which I've done).

bakshiutkarsha:
> Thanks @bakshiutkarsha. The link for the new ZIM won't let me in -- says I need to request access from you (which I've done).

I have given access now.

Jaifroid:
Ah, Nelson! Much better!! 😊

`toc.getSections()` is now working in Chromium:

image

Jaifroid:
But I'm having a bit of trouble using scrollToSection():

image

bakshiutkarsha:
> But I'm having a bit of trouble using scrollToSection():
> 
> image

I think I know what the problem is, the article is opening in an iframe, so `location.href` is not working how it is supposed to work, let me check the solution for this.

Jaifroid:
@bakshiutkarsha I would just use JavaScript scrollTo or scrollIntoView() (I can't quite remember now), so that the element comes into view. We can't use a reload for this purpose (location.href), as it will disrupt many things. But a reference *within* an iframe document to document will reference the iframe, and outside an iframe it will reference the top level, so using scrollIntoView shoudl work. Apologies for the rush, I can't look into this till tomorrow.

Jaifroid:
Try: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView 

But remember that if the element to be closed to is in a hidden/collapsed section, it may need to be opened first (though that could be a call to another API dedicated to opening and closing sections, depending on the API overall design.

Jaifroid:
Another thing to be aware of is that IE11 doesn't support .closest(), and I'm seeing this (image below). It should already be polyfilled with corejs, so I'm not quite sure why that's appearing. Possibly the corejs /babel config needs looking at? (I'm not experienced in using them.)

image

kelson42:
Could we do following renaming:
* `toc_level` -> `level`
* `section_id` -> `section_id`
* `section_name` -> `label`

I believe as well that the `toc_level` should be only the number.

Jaifroid:
@bakshiutkarsha @kelson42  I've added a little demo of usage of the API to the Kiwix JS Widnows `Test ToC API` branch. Code here:

https://github.com/kiwix/kiwix-js-windows/blob/Test-ToC-API/www/js/app.js#L3666 

The usage code is pasted below for convenience. @bakshiutkarsha to run it from localhost remember you have to update branch, delete the pwabuilder service worker and force reload first. This demonstrates the use of `scrollIntoView()` which I think is a better way of scrolling to an element without changing location hash (it's also a smooth scroll on some browsers). That particular usage could be incorporated into `scrollToSection()` (which this demo doesn't currently use due to errors noted above).

```
            // Use Kiwix API
            var useTocAPI = function() {
                console.log("Kiwix API is available and attached");
                // Demonstrate API usage
                var _toc = window.frames[0].frameElement.contentWindow.zim.toc;
                if (_toc) console.log("The toc API is available");
                if (_toc && _toc.hasTableOfContent()) {
                    console.log("Article has a table of contents");
                    var sections =_toc.getSections();
                    // Scroll to Section (change variable for testing)
                    var secNum = 10;
                    console.log("Scrolling to " + secNum);
                    innerDoc.getElementById(sections[secNum].section_id).scrollIntoView();
                } else {
                    console.log("Article has not table of contents");
                }
            };
```

Jaifroid:
@bakshiutkarsha I forgot to say that to see the demo working, if you wan to try or experiment with the code, switch to **jQuery** mode in Configuration. Reason: in SW mode, the sections are closed, and this demo doesn't deal with opening a perent section before scrolling to a collapsed section.

Jaifroid:
@bakshiutkarsha Have added a very basic open section bit of code to the demo so it now works in SW mode, but not yet on IE11, which doesn't support closest. I'm not sure if the [`closest()` polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) is being exported, and if so where it could be found, but it may only be working inside the functions provided by the API?

zim/api/toc.js
```diff
@@ -0,0 +1,45 @@
...
+            "section_id": section.id,
+            "section_name": section.innerText,
+        })
+    });
+    return respArrOfSections;
+}
+
+zim.toc.scrollToSection = function(index){
+    const sectionIdElem = getSectionList()[index];
+    sectionIdElem.closest('details').setAttribute('open', '');
```
Jaifroid, line 37: @bakshiutkarsha I think this line has to be changed to:
```
const closestClosed = sectionIdElem.closest('details:not([open])');
if (closestClosed) closestClosed.setAttribute('open', '');
```
The reason is that often `h3` headings are open in MediaWiki ZIMS, while the parent `h2` is closed, so we have to search for the nearest ancestor that is closed and open that one.

zim/api/toc.js
```diff
@@ -0,0 +1,45 @@
...
+    location.href = `#${sectionIdElem.id}`;
```
Jaifroid, line 38: @bakshiutkarsha This can be changed to:

`sectionIdElem.scrollIntoView();`

mgautierfr:
There are a lot of new things here I will not comment but I would like to raise my previous comment I think it has not been address/answered :

> The real point (for me) is not the code but the public api :
>
> -    hasTableOfContent() -> bool
> -    getSections() -> list of sections {"toc_level", "section_id", "ection_name"}
> -    scrollToSection(index)
>
>I wonder if the scrollToSection should take a "id" instead of a "index". It would be more useful as the section id is already in the returned information of getSections.
> If we need a index, the section_id is useless and user code has to keep a index with the returned section dict to be able to scroll to the section

Jaifroid:
@mgautierfr The API functions as you describe it except for `scrollToSection`. I agree it might make more sense to have that take an id rather than an index. Also, since getSections returns an array, it is already numerically indexed and there is no point replicating the array index in another field.

bakshiutkarsha:
@Jaifroid  I have done the mentioned changes and verified as well, the scroll is now working. Here is the latest zim you can verify with: [Zim](https://drive.google.com/file/d/13u93Y1xMc31PQUXyhUs4c-QaVJ38PXU_/view?usp=sharing)

Jaifroid:
@bakshiutkarsha I've updated the demo at [Test ToC API](https://github.com/kiwix/kiwix-js-windows/tree/Test-ToC-API). It's working well on Chromium now, including scrolling to a section within a closed section, and it now opens the parent.

Just one more issue I'm running into is with IE11 (which will also affect other supported older browsers). See console log output below. The problem here is that `corejs` is not exporting the [`closest` polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill). Maybe you need to tweak config? (It's not a complicated polyfill, just a few lines of code, but it's much better for corejs / babel to provide it rather than to use custom code in the API). 

image

Jaifroid:
Other than that, what do you and @kelson42 think of @mgautierfr's suggestion to have scrollToSection(id) instead of scrollToSection(index)?

.babelrc
```diff
@@ -0,0 +1,8 @@
...
+    "presets": [
+        ["@babel/preset-env", {
+            "useBuiltIns": "entry",
+            "corejs": 3
+        }]
```
Jaifroid, line 6: @bakshiutkarsha Looking at the [Babel usage guide](https://babeljs.io/docs/en/usage), I think the issue with missing polyfills might be because you haven't specified any browsers in this setup file. E.g. the setup has these entries (NB don't use these, they need to be modified for our needs):
```
  "presets": [
    [
      "@babel/env",
      {
        "targets": {
          "edge": "17",
          "firefox": "60",
          "chrome": "67",
          "safari": "11.1",
        },
        "useBuiltIns": "usage",
        "corejs": "3.6.4",
      }
    ]
  ]
```
As a minimum we need to support:
```
"edge": "15",
"firefox": "52",
"ie": "11"
```
I'm not sure what we support for Chrome (e.g. minimum Android version) and Safari for iOS app), but with IE11, most other things are probably covered.

bakshiutkarsha: @Jaifroid  I have done that in package.json [here](https://github.com/openzim/mwoffliner/pull/1237/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R145)

Jaifroid: Hmm. So any thoughts on why it's not working?

Two random guesses:

1. (I don't really know babel config, so correct me if I'm wrong): maybe adding the browser targets in `package.json` only affects the environment in which **mwoffliner** runs, i.e. `node.js` (I assume)? Whereas we need a config that will load corejs polyfills according to the environment in which `-/_api/toc.js` is running, i.e. the detection logic needs to run inside toc.js. That may be completely wrong...
2. You're using corejs 3 with `babel-preset-env`. I wonder if the breaking change mentioned in https://github.com/babel/babel/pull/7646 affects this setup:

> Updating @babel/polyfill to core-js@3 will break @babel/preset-env. @babel/preset-env with useBuiltIns transforms import of @babel/polyfill to import of required core-js modules. Most users don't add core-js dependency directly. In case of a minor update, even if @babel/polyfill dependency will be updated with @babel/preset-env, in the top level of node_modules we will have core-js@2, not core-js@3 because a serious part of dependencies depends on packages with core-js@2 in dependencies.

A not great alternative would be to provide our own closest function if `Element.closest` is undefined:
```
        // Create a closest function alternative because IE11 and others do not support closest
        function closest (ele, s) {
            var cele = ele;
            var cmatches = Element.prototype.matches || Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
            do {
                if (cmatches.call(cele, s)) return cele;
                cele = cele.parentElement || cele.parentNode;
            } while (cele !== null && cele.nodeType === 1);
            return null;
        };
```

kelson42: @bakshiutkarsha Can you help here or have you left this PR to someone else?

bakshiutkarsha: Yeah, I can close this. Will implement the changes suggested by @Jaifroid .

Jaifroid:
@kelson42 @bakshiutkarsha  Is there any way we can get this finished? We were very very close!

kelson42:
@Jaifroid I hope so soon. This is part of 1.12 milestone (so in November) with a few other ticket. Hopefully @MananJethwani will be able to help here to complete.

kelson42:
I have rebased the project on git `master`.

PR #1238, opened 2020-07-26
Broken API URL fixed for some wikis

bakshiutkarsha:


bakshiutkarsha:
> Please can you give me one or two wikis failing currently because of this?

gbf.wiki and stupidedia

Issue #990, opened 2019-09-21
Input arguments should be tested properly at the start of the process


kelson42:
Many times, if the URLs given (or any another command line arguments) are wrong, then the process starts and dies later. All the arguments should be properly tested before anything else.

kelson42:
See #1069 and #491

MananJethwani:
@kelson42 I would like to work on this if no one is working on it

MananJethwani:
the solution I think would be good is to make another test.ts file and make individual as well as combined test for all arguments so that this file can be re used at other places as well.

kelson42:
@MananJethwani Your are welcome to make a PR. Start small, just test one or two things first.

PR #1249, opened 2020-08-17
Sanitize command line arguments

MananJethwani:
Adding tests for arguments in the beginning of the process. for issue #990

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -50,6 +50,16 @@ if (argv.osTmpDir) {
   }
 }
 
+/* ***********************/
+/* TESTING ALL ARGUMENTS */
+/* ***********************/
+
+import { sanitize_all } from './sanitize-argument';
+sanitize_all(argv).catch((err) => {
+  console.error(`\n\n**********\n\n${err.message}\n\n**********\n\n`);
+  process.exit(2);
+});
+
 /* ***********************************/
 /* GO THROUGH ENTRY POINT            */
 /* ***********************************/
--- a/src/sanitize-argument.ts
+++ b/src/sanitize-argument.ts
@@ -1,0 +1,118 @@
+import os from 'os';
+import S3 from './S3';
+import axios from 'axios';
+import Redis from './Redis';
+import urlParser from 'url';
+import logger from './Logger';
+import { config } from './config';
+import fs, { readFileSync } from 'fs';
+import * as QueryStringParser from 'querystring';
+import { isValidEmail, ensureTrailingChar } from './util';
+
+export async function sanitize_all(argv: any) {
+  // extracting all arguments
+  const {
+    speed: _speed,
+    adminEmail,
+    mwUrl,
+    customZimFavicon,
+    optimisationCacheUrl,
+  } = argv;
+
+  const cpuCount = os.cpus().length;
+
+  // sanitizing speed
+  sanitize_speed(_speed);
+
+  // sanitizing s3
+  try {
+    if (optimisationCacheUrl) {
+      // Decompose the url with path and other S3 creds
+      const s3UrlObj = urlParser.parse(optimisationCacheUrl);
+      const queryReader = QueryStringParser.parse(s3UrlObj.query);
+      const s3Url = (s3UrlObj.host || '') + (s3UrlObj.pathname || '');
+      this.s3Obj = new S3(s3Url, queryReader);
+      await this.s3Obj.initialise().then(() => {
+        logger.log('Successfully logged in S3');
+      });
+    }
+  } catch (err) {
+    throw err;
+  }
+
+  // sanitizing mwUrl
+  await sanitize_mwUrl(mwUrl).catch((err)=>{
+    throw err;
+  });
+
+  // sanitizing adminEmail
+  sanitize_adminEmail(adminEmail);
+
+  // redis client sanitization
+  // created a redis client and then closed it.
+  sanitize_redis(argv);
+
+  // sanitizing custom zim favicon
+  if (customZimFavicon) {
+    await sanitize_customZimFavicon(customZimFavicon)
+    .catch(err=>{
+      throw err;
+    });
+  }
+}
+export function sanitize_speed(_speed:any)
+{
+  if (_speed && isNaN(_speed)) {
+    throw new Error(
+      `speed is not a number, please give a number value to --speed.`
+    );
+  }
+}
+
+export async function sanitize_mwUrl(mwUrl:string)
+{
+    await axios
+    .get(mwUrl)
+    .catch((err)=>{
+      throw new Error(`mwUrl [${mwUrl}] is not valid.`)
+    })
+}
+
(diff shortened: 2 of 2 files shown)


MananJethwani:
@kelson42 I wil keep on making changes and pushing here.

kelson42:
@MananJethwani I have put your PR in draft. Please change it when you have finished and PR is ready for review.

MananJethwani:
@kelson42 for testing redis argument I need to create a redis client should I test that as well for errors by creating it?

kelson42:
@MananJethwani Sorry for being so late with your review, I will give you a feedback soon.

MananJethwani:
no problem I understand You might have work load due to GSOC 2020. I will start working on other issues as well you take your time :)

src/cli.ts
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 // tslint:disable-next-line
-':' //# -*- mode: js -*-; exec /usr/bin/env node --max-old-space-size=9000 --stack-size=42000 "$0" "$@"
+':'; //# -*- mode: js -*-; exec /usr/bin/env node --max-old-space-size=9000 --stack-size=42000 "$0" "$@"
```
kelson42, line 3: I'm not sure about that? What the reason behind?

src/test.ts
```diff
@@ -0,0 +1,176 @@
+import os from 'os';
```
kelson42, line 1: Maybe we could rename `test.ts` to `sanitize-arguments.ts` this might be nearer to what we want and avoid any confusion (for example with automated tests)?

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+import urlParser from 'url';
+import logger from './Logger';
+import { config } from './config';
+import MediaWiki from './MediaWiki';
+import Downloader from './Downloader';
+import fs, { readFileSync } from 'fs';
+import * as QueryStringParser from 'querystring';
+import { isValidEmail, ensureTrailingChar } from './util';
+
+function IsURL(url: string) {
```
kelson42, line 14: I'm quite sure we can do that with a standard module?

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (customZimFavicon) {
+    await test_customZimFavicon(customZimFavicon,downloader)
+    .catch(err=>{
+      throw err;
+    });
+  }
+}
+
+export function test_speed(_speed:any)
+{
```
kelson42, line 124: I would prefer to have this on the line above.

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+  try {
+    const test_redis = new Redis(argv, config);
+    logger.log('closing test redis DB');
+    test_redis.client.quit();
+  } catch (err) {
+    throw err;
+  }
+}
+
+export async function test_customZimFavicon(customZimFavicon:any,downloader:Downloader)
```
kelson42, line 157: space after `,`

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+    noLocalParserFallback,
+    forceLocalParser,
+    customFlavour: customProcessorPath,
+  } = argv;
+
+  const cpuCount = os.cpus().length;
+
+  // checks for speed
+  test_speed(_speed);
+  const speed = Math.max(1, Math.round(cpuCount * (_speed || 1) * 3));
```
kelson42, line 61: This line should not be there I believe.

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+    format,
+    filenamePrefix,
+    publisher: _publisher,
+    outputDirectory: _outputDirectory,
+    addNamespaces: _addNamespaces,
+    articleList: _articleList,
+    customZimFavicon,
+    optimisationCacheUrl,
+    noLocalParserFallback,
+    forceLocalParser,
```
kelson42, line 53: This is a duplicate of code which should avoid to define the argument line parsing twice. We should have shared code.

src/test.ts
```diff
@@ -0,0 +1,176 @@
...
+export function test_speed(_speed:any)
+{
+  if (_speed && isNaN(_speed)) {
+    throw new Error(
+      `speed is not a number, please give a number value to --speed.`
+    );
+  }
+}
+
+export function test_url(mwUrl:string)
```
kelson42, line 132: In this function and many other one, there is an incoherency between the name of the function and what is done. This fonction should be called `test_mwUrl` (then the error msg makes sense). Or we should have a `test_url` and then the error msg should be outside.

kelson42 (changes requested):
In the principle looks good, but a few details to be fixed. I believe as well that it would be beneficial to consider this more as a "sanitize_arguments". This would:
* Parse the command line
* Verify the argument
* Fix them if necessary
* Returns an error if necessary
* Returns the sanitized values so we can continue with the process.

@MananJethwani Does that makes sense to you?

src/cli.ts
MananJethwani: sorry might be done by the js extension will revert it back

src/test.ts
MananJethwani: yes that is a better name I will change it

src/test.ts
MananJethwani: @kelson42 I was earlier using standard module(valid-Uri) but these modules only check if URI is formatted properly or not they don't check if these URI make sense for example these will let pass URI like - https://en.wikip where as this URI doesn't make any sense. I don't know any other module or could find it any where if you know please let me know

MananJethwani: or is it fine to let these types of URI to pass?

src/test.ts
MananJethwani: @kelson42 I needed speed to declare downloader class

src/test.ts
kelson42: Maybe you could use the module you talk about and check that you get HTTP 200 OK with a simple Axios get request? Or maybe just even the Axios simple get request?

src/test.ts
MananJethwani: should I rename the functions as well to like sanitize-mwUrl etc.

src/test.ts
MananJethwani: @kelson42 I don't understand how to do this part....can you help?

src/test.ts
kelson42: Yes, but ok but please keep using `_` as separator

src/test.ts
kelson42: That might be the problem. You should not instanciate the Downloader... it seems you just need it to download the favicon. Why not using Axios directly?

src/test.ts
kelson42: OK, maybe we should leave that for the moment.

MananJethwani:
I don't understand what is wrong here I checked the logs and then reverted back those changes and tried running the same test it shows the same error.

kelson42 (approved):
Thx, looks better now.

PR #1250, opened 2020-08-17
Fix for display issues

### LINKED ISSUE #1177: Major display issues with Russian wiktionary
*@ErinPo commented on Jun 21, 2020, 7:39 AM UTC:*

All search result pages are 90% overlapped with a Creative Commons disclaimer which cannot be dismissed. Only opening the main page works without overlays, but then it fetches the desktop version which causes lots of horizontal scrolling. It seems the offline browser is not properly sending the mobile user agent.

[Screenshot](https://i.imgur.com/vNzRpKY.jpg)

*This issue was moved by [kelson42](https://github.com/kelson42) from [kiwix/apple#280](https://github.com/kiwix/apple/issues/280).*

bakshiutkarsha:
fixes #1177

### DIFF
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -111,7 +111,7 @@ const renderMCSArticle = (json: any, dump: Dump, articleId: string, articleDetai
         json.remaining.sections
             .forEach((oneSection: any, i: number) => {
                 // if below is to test if we need to nest a subsections into a section
-                if (oneSection.toclevel === 1) {
+                if (oneSection.toclevel === 1 || oneSection.toclevel === 0) {
                     html = html.replace(`__SUB_LEVEL_SECTION_${i}__`, ''); // remove unused anchor for subsection
                     html += sectionTemplate({
                         section_index: i + 1,


src/util/articleRenderers.ts
```diff
@@ -111,7 +111,7 @@ const renderMCSArticle = (json: any, dump: Dump, articleId: string, articleDetai
         json.remaining.sections
             .forEach((oneSection: any, i: number) => {
                 // if below is to test if we need to nest a subsections into a section
-                if (oneSection.toclevel === 1) {
+                if (oneSection.toclevel === 1 || oneSection.toclevel === 0) {
```
kelson42, line 114: * Why we keep `toclevel === 1`?
* What are the boundaries of tolevel values?
* What is this test doing here exactly, the comment above does not make it clear to me, can you please rewrite it?

bakshiutkarsha: Other Mediawikis that I have come across have their TOC sections starting from level 1 and sub-sections in these wiktionary start with level 2. For eg. en.wiktionary.

https://en.wiktionary.org/api/rest_v1/page/mobile-sections/Latin

On the other hand, ru.wiktionary has TOC section starting from level 0 and its sub-sections start from 1.

https://ru.wiktionary.org/api/rest_v1/page/mobile-sections/%D0%92%D0%B8%D0%BA%D0%B8%D1%81%D0%BB%D0%BE%D0%B2%D0%B0%D1%80%D1%8C

I checked both kind of Mediawikis and adding the check wouldn't affect existing wikis and would help us handle sections for wikis having TOC sections starting with 0.

kelson42: @bakshiutkarsha Why some are starting with toclvel == 1 and other tolevel == 2? maybe the wiki code is not the same?

PR #1251, opened 2020-08-17
Style of navbar fixed for wikitonary

### LINKED ISSUE #1033: Wiktionary translations can't be shown
*@danielzgtg commented on Feb 23, 2020, 3:21 AM UTC:*

**Describe the bug**

The translations in Wiktionary cannot be viewed anymore on the latest ZIMs. They start collapsed and do not expand when tapped. This applies to translations, conjugations, and collapsible tables in general.

This happens with:

*   `wiktionary_en_all_maxi_2020-01.zim`
*   `wiktionary_en_all_nopic_2019-12.zim`
*   `wiktionary_en_all_novid_2018-06.zim`

This does not happen with and things worked fine with `wiktionary_en_all_nopic_2017-08.zim`

**Expected behavior**

To be able to see the translations.

**Steps to reproduce the behavior:**

1.  Tap on "Library" in the overflow menu
2.  Open `wiktionary_en_all_maxi_2020-01.zim`
3.  Open the definition for "wiki"

**Screenshots**

Actual behavior:

[Screenshot_20200222-191642](https://user-images.githubusercontent.com/25646384/75102845-7f892e80-55c0-11ea-9108-f4f05e4f8571.jpg)

Example of expected behavior in older ZIM:

[Screenshot_20200222-191715](https://user-images.githubusercontent.com/25646384/75102846-81eb8880-55c0-11ea-948b-4de2354b5911.jpg)

**Environment**

*   Version of Kiwix Android : 3.1.3
*   Device : OnePlus 7 Pro
*   OS version : OxygenOS Open Beta 9

*This issue was moved by [kelson42](https://github.com/kelson42) from [kiwix/kiwix-android#1796](https://github.com/kiwix/kiwix-android/issues/1796).*

bakshiutkarsha:
Fixes #1033

### DIFF
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -112,7 +112,7 @@ const renderMCSArticle = (json: any, dump: Dump, articleId: string, articleDetai
         const firstTocLevel = json.remaining.sections[0].toclevel;
         json.remaining.sections
             .forEach((oneSection: any, i: number) => {
-                if (oneSection.toclevel === firstTocLevel) {
+                if (oneSection.toclevel === firstTocLevel || oneSection.text.indexOf('class="NavFrame"')) {
                     html = html.replace(`__SUB_LEVEL_SECTION_${i}__`, ''); // remove unused anchor for subsection
                     html += sectionTemplate({
                         section_index: i + 1,
--- a/res/inserted_style.css
+++ b/res/inserted_style.css
@@ -42,5 +42,3 @@ sup,sub{line-height:0.75em}dfn{font-style:normal;font-weight:bold}i > dfn{font-s
 
 
 .client-js .toc-mobile{background-color:#f8f9fa;border:solid 1px #eaecf0}.client-js .toc-mobile .toc-button{float:left;display:block;font-size:0.8em}.client-js .toc-mobile .collapsible-heading{visibility:visible;background-position:right center}.client-js .toc-mobile .collapsible-heading .indicator{margin:0.2em 0.4em 0 0;float:right;font-size:0.6em}.client-js .toc-mobile .collapsible-block{border-bottom:0;margin-left:52px;margin-right:24px;font-size:0.7em;width:auto}.client-js .toc-mobile .collapsible-block > ul{padding-bottom:1.3em}.client-js .toc-mobile .collapsible-block ul{list-style:none;margin-bottom:1px}.client-js .toc-mobile .collapsible-block ul li{padding-top:0.8em;line-height:1.2}.client-js .toc-mobile .collapsible-block ul li a{white-space:pre-wrap;word-break:break-word}
-
-.NavContent:not([open]) {display: block !important}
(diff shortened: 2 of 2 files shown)


bakshiutkarsha:
> This need an in detail explanation why we need to insert custom CSS. Why the CSS of the online version is not sufficient?

As per the discussion on the issue thread and based on my investigation, it seems that either the event handler for toggling the visibility of the navframe is not getting attached or the custom CSS inserted by mwoffliner is causing problems with the visibility. To fix tthe issues, I tried both the approaches mentioned on the issue thread:

- Using javascript: I added a custom JS script to be bundled and loaded in the article page to remove `client-js` class from `body` and add the `client-nojs` class to it. This solution doesn't seem to fix the visibility issue.

- So finally, I had to retort to adding custom CSS that overrides the existing CSS for the `NavFrame`. This fixes the issue and the translations are visible as the custom CSS is marked `important` and would always override conflicting CSS property for this CSS class. The chances of it breaking something else is very low as the CSS is applied on just 1 property of 1 class.

kelson42:
@bakshiutkarsha We should retrieve any css/js from the online version. If not we need to understand exactly why. This explanation is missing here... you fix a problem which is not even understodd priperly. This should be avoided.

bakshiutkarsha:
> @bakshiutkarsha We should retrieve any css/js from the online version. If not we need to understand exactly why. This explanation is missing here... you fix a problem which is not even understodd priperly. This should be avoided.

https://github.com/openzim/mwoffliner/issues/1033#issuecomment-596182865
@kelson42  Now, I carefully scrutinized all the comments by @Jaifroid on this ticket and this last comment helped me to resolve it. His previous comments have explained exactly the question of `why` is it happening and `how` to resolve it. I made the same fix suggested by him. Let me know if you think this fix is not proper, I will try to dig deeper.

Jaifroid:
I'd be happy to test / troubleshoot the implementation here, but it may take a few days.

kelson42 (changes requested):
No, like I said many times. I don’t want dependence to the wiki content in mwoffliner code.

src/util/articleRenderers.ts
```diff
@@ -112,7 +112,7 @@ const renderMCSArticle = (json: any, dump: Dump, articleId: string, articleDetai
         const firstTocLevel = json.remaining.sections[0].toclevel;
         json.remaining.sections
             .forEach((oneSection: any, i: number) => {
-                if (oneSection.toclevel === firstTocLevel) {
+                if (oneSection.toclevel === firstTocLevel || oneSection.text.indexOf('class="NavFrame"')) {
```
Jaifroid, line 115: @bakshiutkarsha I'm not commenting on the general idea here, but only on your use of `indexOf`. The `indexOf` method returns -1 if the string you are searching does **not** contain the search string, not 0. Therefore, unless I've misunderstood what you're trying to do, I think you need:

`~oneSection.text.indexOf('class="NavFrame"')
`
**Note the tilde at the start.** This converts -1 to 0, and 0 to -1, etc. Therefore, if the text appears at position 0, the tilde converts 0 to -1 and the test will pass. But if `indexOf` returns -1 (not found), the tilde converts it to 0, and the test fails.

Jaifroid: PS It would probably be better to use the `classList.contains()` method for this test. It is supported on IE10+. See https://dev.caniuse.com/classlist .

Jaifroid: Also, your test would fail if an editor had written `class='NavFrame'` with single quotes or `class="dark NavFrame"`, hence why using `classList.contains()` is safer.

kelson42:
@bakshiutkarsha Any news here? If our section mgmt code css+js is not able to make the difference between our section code and the content code, then we need to specify our DOM nodes properly, like for example by putting `class=_mwoffliner`?

PR #1262, opened 2020-09-21
fixed automated test failing due to lack of offline.css

MananJethwani:
@kelson42 I just added some catch blocks another way was to add status code 404 as valid status code but then the response received in offlineCSSContent and finallCss would be different as the response has a time stamp embeded in HTML part of data

another problem was if we add status code 404 as a valid status code then we have to make it valid in array buffer option which can cause problems while downloading actual css content during scraping.


MananJethwani:
the test is now failing on another file the issue I mentioned is solved here actually

PR #1263, opened 2020-09-22
Fix automated tests

kelson42:


MananJethwani:
@kelson42 I guess all the URL used during testing should be hosted under wiki.kiwix only using someone else's URL is not good as in future if some of those URL become invalid like in case of offline.css tests will start failing again

MananJethwani:
Yes I know and I am suggesting that the change you have applied must be done with all URL as if we are using someone else's URL in future they might not exist creating new error.....and can I work on e2e test if no one is working on them?....the redundancy check is failing in one and in other URL test is failing

kelson42:
@MananJethwani Can you please comment on the PR? I don't know about which other URLS you are talking about. I had a look as well to the e2e test problem... and so far it looks like a bug (a regression) in the backend, have reported it at https://phabricator.wikimedia.org/T263528 (so wait and see).

Issue #1254, opened 2020-09-07
Add --webp option


kelson42:
With this option, GIF, PNG and JPEG should be converted to WEBP images.

Jaifroid:
But this would leave a number of platforms behind. See https://caniuse.com/webp . I don't think browser support is ready for this?

kelson42:
@Jaifroid Few clarifications:
* For the moment this would be an option (so not run everywhere)
* You probably need to read https://github.com/openzim/python_scraperlib/issues/45
* First candidates would be the ZIM to be put in the Android custom apps

Jaifroid:
Ah, OK, thanks! I see there is a polyfill that supports old browsers, including IE11, so it's not the breaking change I thought it was at first.

arjungarg07:
Hey @kelson42, I am willing to work on this issue. Can you assign this to me?

arjungarg07:
Hey @kelson42, I am willing to work on this as my first issue. Can you assign this to me? and Can you brief me more about this issue, please!

arjungarg07:
Hey @kelson42, I am willing to work on this as my first issue. Can you assign this to me?

kelson42:
@arjungarg07 Thx, good to me.

arjungarg07:
Do i have to add this option in the parameters of mwoffliner API?
Can you please brief me more about this.

kelson42:
@arjungarg07 yes, this is a new option.

PR #1265, opened 2020-09-23
Add support to Webp bitmap image conversion

MananJethwani:
@kelson42 I am using sharp to convert buffer from images to webp format, resolving #1254 should I add a test for this option?


MananJethwani:
@kelson42 just made animation on webp working and files will have `.webp` after conversion but there is a problem if I change file name for gif animations are removed can you look at it if possible

src/Downloader.ts
```diff
@@ -586,6 +590,10 @@ class Downloader {
   }
 
   private async getCompressedBody(resp: any): Promise<any> {
+    const content_type = resp.headers['content-type'];
+    if (this.webp && (content_type.includes('jpeg') || content_type.includes('png') || content_type.includes('jpg') || content_type.includes('gif'))) {
+      imageminOptions.plugins.push(imageminWebp({ quality: 75 }))
+    }
     return this.isMimeTypeImage(resp.headers['content-type']) ? await imagemin.buffer(resp.data, imageminOptions) : resp.data;
```
kelson42, line 597: This part and `imageminOptions` should be rewritten based on our discussion on Slack. Should be a one line method.

src/util/saveArticles.ts
```diff
@@ -28,6 +28,11 @@ type FileStore = RedisKvs<{
     width?: number;
 }>;
 
+function isImageUrl(url: string) {
+    url.toLowerCase();
+    return (url.substr(url.length-4) === 'jpeg' || url.substr(url.length-3) === 'jpg' || url.substr(url.length-3) === 'png' || url.substr(url.length-3) === 'gif')
+}
+
```
kelson42, line 35: No:
* Same name than an other method of the same software which does something similar but not identical
* Use regex for this kind of work

Just remove this, more infos inthe next comment

src/util/saveArticles.ts
```diff
@@ -49,6 +54,9 @@ export async function downloadFiles(fileStore: FileStore, zimCreator: ZimCreator
             let isFailed = false;
             try {
                 if (resp.result && resp.result.content) {
+                    if (isImageUrl(resp.path) && downloader.webp) {
+                        resp.path += '.webp'
+                    }
```
kelson42, line 59: I don't think this is the right approach because:
* It duplicates a test which is already done in `getCompressedBody()`
* It is based on the filename (which is less trustable than the `mime-type`)

I belive we should avoid that here totally by:
* allowing `getCompressedBody` to update the filename in the `resp`
* secure this new resp is returned in `getContentCb()`

Than way we test only one time, optimize content and change filename together. Less code, quicker and less error prone.

src/util/saveArticles.ts
```diff
@@ -517,7 +525,12 @@ async function treatImage(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
         }
 
         /* Change image source attribute to point to the local image */
-        img.setAttribute('src', newSrc);
+        if (downloader.webp && isImageUrl(newSrc)) {
+            img.setAttribute('src', newSrc + '.webp');
+        }
+        else {
+            img.setAttribute('src', newSrc);
+        }
```
kelson42, line 533: Use a ternary operator. Concerning `isImageUrl` (which I have asked to delete), you will have to create a new method called `isWebpImageUrl()` and please implement it like `isImageUrl()` ie. with a regular expression.

kelson42 (changes requested):
We are almost there.

Jaifroid:
For us at Kiwix JS, the most compatible method to insert the two polyfills needed would be the HTML Install method (Option A on the [Webp Hero readme](https://github.com/chase-moskal/webp-hero)), referencing local copies of both polyfills in the ZIM. Note we would need both because we need to support IE11. Could they be included only if an article has images, to prevent overhead in case of ZIMs (or individual articles) with no images?

src/util/saveArticles.ts
MananJethwani: I will use a regular expression here but other than that it should be called isImageUrl only i guess because `newSrc` won't be having .webp in it(as it has been taken out of html content directly) we have to concatenate webp in it.

kelson42: shouldConvertImageUrlToWebp() ?

MananJethwani: ok

src/util/saveArticles.ts
MananJethwani: @kelson42 the response being retruned here by getContentCb is not having the property resp.path, resp.path is set in DownloadBulk function. should I then also try to implement these changes?

MananJethwani: another point is `isMimeTypeimage()` passes true for svg as well and we are not converting svg to webp so we shouldn't concatenate anything here I can apply it in `downloadBulk()` instead of here that would be better

MananJethwani:
@kelson42 for now I have applied all the changes as you asked except the suggestion of allowing `getCompressedBody` to update the filename in the `resp` as I have mentioned above this resp is finally being passed to resp.result in `BulkDownload()` and resp.path is set there seperately the resp argument given to `getCompressedBody()` has no property path.

src/util/saveArticles.ts
```diff
@@ -104,7 +108,7 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
             argList,
             async (arg) => {
                 const resp: any = {};
-                resp.path = arg.val.path;
+                resp.path = shouldConvertImageUrlToWebp(arg.val.path, downloader.webp) ? arg.val.path + '.webp' : arg.val.path;
```
kelson42, line 111: This should happen in the optimization part: only one test and then (1) make the optim (2) change the mime-type and filename in the `resp`.

src/util/saveArticles.ts
```diff
@@ -517,7 +521,7 @@ async function treatImage(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, art
         }
 
         /* Change image source attribute to point to the local image */
-        img.setAttribute('src', newSrc);
+        shouldConvertImageUrlToWebp(newSrc, downloader.webp) ? img.setAttribute('src', newSrc + '.webp') : img.setAttribute('src', newSrc);
```
kelson42, line 524: Crraige return after `?`, long line should ideally be avoided.

kelson42 (changes requested):
One time these remarks are taken in account, we sould then look:
* At testing
* S3 support.

src/Downloader.ts
```diff
@@ -585,20 +616,29 @@ class Downloader {
...
     logger.info(`Downloading [${url}]`);
     try {
-      if (this.optimisationCacheUrl && this.isImageUrl(url)) {
-        this.downloadImage(url, handler);
+      if (this.optimisationCacheUrl && isImageUrl(url)) {
+        this.downloadImage(url, handler, path);
       } else {
-        const resp = await axios(url, this.arrayBufferRequestOptions);
+        let resp = await axios(url, this.arrayBufferRequestOptions);
+        resp = await this.getCompressedBody(resp, path)
```
kelson42, line 638: You should not have to give the `path` through all this methods. This should be avoided.

Deal with `resp.headers.path_postfix`.

kelson42 (changes requested):

src/util/misc.ts
```diff
@@ -334,3 +334,11 @@ export function stripHttpFromUrl(url: string): string {
...
 }
 
+
+export function isImageUrl(url: string): boolean {
+  return !!URL_IMAGE_REGEX.exec(url);
+}
+
+export function isImageMimeType(mimetype: string): boolean {
+  return !!BITMAP_IMAGE_MIME_REGEX.exec(mimetype);
+}
```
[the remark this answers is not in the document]
MananJethwani: ok will make 4 functions and !! were not my doing.....it would be better to use .test instead

src/util/const.ts
```diff
@@ -3,7 +3,9 @@ export const MEDIA_REGEX = /^(.*\/)([^/]+)(\/)(\d+px-|)(.+?)(\.[A-Za-z0-9]{2,6}|
 export const MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE = 10;
 export const URL_IMAGE_REGEX = /(https?:\/\/.*\.(?:jpg|gif|png|svg|jpeg))/i;
```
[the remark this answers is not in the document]
MananJethwani: sorry might have missed it will take care of that as well

src/Downloader.ts
```diff
@@ -109,6 +104,49 @@ class Downloader {
...
+          imageminAdvPng({ optimizationLevel: 4, iterations: 5 }),
+          imageminWebp({ quality: 75 })
+        ]
+      })
+      imageminOptions.set('image/gif', {
+        plugins: [
+          imageminGifsicle({ optimizationLevel: 3, colors: 64 }),
+          imageminWebp({ quality: 75 })
+        ]
+      })
```
kelson42, line 128: I don't have seen that before, but I don't see any value to to imageminJpegoptim` and all the traditional optims if we do the webp conversion. We should to only the webp convertion. Make a small loop over the the three mime-types and just make it working with `imageminWebp`

src/Downloader.ts
```diff
@@ -109,6 +104,49 @@ class Downloader {
...
+    } else {
+      imageminOptions.set('image/jpeg', {
+        plugins: [
+          imageminJpegoptim({ max: 60, stripAll: true }),
+          // imageminJpegtran(),
```
kelson42, line 133: Delete the comment please

src/Downloader.ts
```diff
@@ -109,6 +104,49 @@ class Downloader {
...
+        ]
+      })
+      imageminOptions.set('image/png', {
+        plugins: [
+          // imageminOptiPng(),
```
kelson42, line 138: Delete the comment please

src/Downloader.ts
```diff
@@ -586,19 +616,27 @@ class Downloader {
   }
 
   private async getCompressedBody(resp: any): Promise<any> {
-    return this.isMimeTypeImage(resp.headers['content-type']) ? await imagemin.buffer(resp.data, imageminOptions) : resp.data;
+    if (isBitmapImageMimeType(resp.headers['content-type'])) {
+      resp.data = await imagemin.buffer(resp.data, imageminOptions.get(resp.headers['content-type']));
+      if (this.webp) {
+        resp.headers.path_postfix = '.webp'
```
kelson42, line 622: Semi-colon missing here... how can it work?

src/Downloader.ts
```diff
@@ -633,9 +671,10 @@ class Downloader {
           this.s3.uploadBlob(stripHttpFromUrl(url), resp.data, etag);
         }
 
+        resp = await this.getCompressedBody(resp)
```
kelson42, line 674: semi-colon missing here again I believe.

src/util/const.ts
```diff
@@ -1,9 +1,11 @@
...
 export const MIN_IMAGE_THRESHOLD_ARTICLELIST_PAGE = 10;
-export const URL_IMAGE_REGEX = /(https?:\/\/.*\.(?:jpg|gif|png|svg|jpeg))/i;
+export const IMAGE_URL_REGEX = /(https?:\/\/.*\.(?:jpg|gif|png|svg|jpeg))/i;
 export const CONCURRENCY_LIMIT = 10;
-export const MIME_IMAGE_REGEX = /^image+[/-\w.]+$/;
+export const IMAGE_MIME_REGEX = /^image+[/-\w.]+$/;
 export const FIND_HTTP_REGEX = /^(?:https?:\/\/)?/i;
 export const DB_ERROR = 'internal_api_error_DBQueryError';
 export const WEAK_ETAG_REGEX = /^(W\/)/;
+export const BITMAP_IMAGE_URL_REGEX = /(jpeg|gif|png|jpg)$/i;
```
kelson42, line 10: It should be `/(https?:\/\/.*\.(?:jpg|gif|png|jpeg))/i`

src/util/saveArticles.ts
```diff
@@ -112,6 +116,7 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
 
                 return downloader.downloadContent(arg.val.url).then((r) => {
                     resp.result = r;
+                    resp.path += resp.result.responseHeaders.path_postfix ? resp.result.responseHeaders.path_postfix : '';
```
kelson42, line 119: `resp.path += resp.result.responseHeaders.path_postfix || ''` would be better I believe

kelson42 (changes requested):

src/util/misc.ts
```diff
@@ -334,3 +334,19 @@ export function stripHttpFromUrl(url: string): string {
...
+  return BITMAP_IMAGE_URL_REGEX.test(url);
+}
+
+export function isImageMimeType(mimetype: string): boolean {
+  return IMAGE_MIME_REGEX.test(mimetype);
+}
+
+export function isBitmapImageMimeType(mimetype: string): boolean {
+  return BITMAP_IMAGE_MIME_REGEX.test(mimetype);
+}
```
kelson42, line 352: `mimeType`, not `mimetype`.

kelson42 (changes requested):

src/Downloader.ts
```diff
@@ -586,19 +600,27 @@ class Downloader {
...
+      if (this.optimisationCacheUrl && isImageUrl(url)) {
         this.downloadImage(url, handler);
       } else {
-        const resp = await axios(url, this.arrayBufferRequestOptions);
+        let resp = await axios(url, this.arrayBufferRequestOptions);
+        resp = await this.getCompressedBody(resp);
         handler(null, {
           responseHeaders: resp.headers,
-          content: await this.getCompressedBody(resp),
+          content: resp.data,
```
kelson42, line 623: I don't think you need that change. `resp` being and object, it should be passed by reference, the modifications you do in `getCompressedBody` should automatically apply and `responseHeaders: resp.headers` should take them in account.

src/Downloader.ts
```diff
@@ -633,9 +655,10 @@ class Downloader {
           this.s3.uploadBlob(stripHttpFromUrl(url), resp.data, etag);
         }
 
+        resp = await this.getCompressedBody(resp);
         handler(null, {
           responseHeaders: resp.headers,
-          content: await this.getCompressedBody(resp),
+          content: resp.data
```
kelson42, line 661: same here, you should not need to change anthing.

src/util/dump.ts
```diff
@@ -42,7 +43,10 @@ export async function getAndProcessStylesheets(downloader: Downloader, links: Ar
                     if (!url.match('^data')) {
                         const filePathname = urlParser.parse(url, false, true).pathname;
                         if (filePathname) {
-                            const filename = pathParser.basename(filePathname).replace(/-.*x./, '.');
+                            const filename = pathParser.basename(filePathname).replace(/-.*x./, '.')
+                            + shouldConvertImageFilenameToWebp(pathParser.basename(filePathname).replace(/-.*x./, '.'), downloader.webp)
```
kelson42, line 47: Do that in two steps, to avoid calling `pathParser.basename(filePathname).replace(/-.*x./, '.')` two times.

kelson42, line 47: @Jaifroid Sorry in advance for the bit silly question, but do you know if the polyfill will apply ton `webp` pictures used in CSS?

kelson42 (changes requested):
* we need to introduce the polyfill
* we need to introduce automated test.

src/util/dump.ts
kelson42: @MananJethwani Like faired, the polyfill can not apply to the pictures in the CSS. We have to secure we don't apply webp optimisation to the pictures the CSS relies on. I don't know exactly how, but considering that these pictures have to be saved in an other namespace, we could try to rely on this to make the filtering.

kelson42 (changes requested):
Concerning the S3, it is currently not taken in account at all and will consequently fail if we use this option with webp.

The solution should be to introduce a `webp` tag to the S3 entries. That way by upload/download, we will know if the content is webp or not. This means we can have in the S3 only one version (webp or gif/png/jpg) at the same time. But that should be good enough for the transitory period of time.

src/util/misc.ts
```diff
@@ -334,3 +335,27 @@ export function stripHttpFromUrl(url: string): string {
...
+  return BITMAP_IMAGE_MIME_REGEX.test(mimeType);
+}
+
+export function shouldConvertImageFilenameToWebp(url: string, webp: boolean) {
+  return webp && WEBP_CANDIDATE_IMAGE_FILENAME_REGEX.test(url);
+}
+
+export function isWebpCandidateImage(webp: boolean, cssDependency: boolean, content_type: string){
+  return (webp && !cssDependency && content_type !== 'image/gif')
+}
```
kelson42, line 361: * Name should be `isWebpCandidateImageMimeType`
* This shoudl use a revex
* Not handling of `cssDependency` at this level

src/S3.ts
```diff
@@ -42,11 +42,11 @@ class S3 {
         });
     }
 
-    public async uploadBlob(key: string, data: any, eTag: string) {
+    public async uploadBlob(key: string, data: any, eTag: string, Webp: boolean) {
         const params = {
             Bucket: this.bucketName,
             Key: key,
-            Metadata: {etag: eTag },
+            Metadata: {etag: eTag ,webp: Webp},
```
kelson42, line 49: S3 should ideally not known anything about the webp problematic. No need to create dependencies to this here. Just replace with a `version` parameter taking a string value and put this as additional Metdata.

Call the function then with `webp` value in case of `webp` conversion, and `1` otherwise.

src/types.d.ts
```diff
@@ -12,6 +12,8 @@ declare module 'mkdirp';
 declare module 'country-language';
 declare module 'imagemin-advpng';
 declare module 'imagemin-jpegoptim';
+declare module 'imagemin-webp';
+declare module 'file-type';
```
kelson42, line 17: We shoudl not need this here, this is a test dependence.

src/Downloader.ts
```diff
@@ -586,19 +591,34 @@ class Downloader {
   }
 
   private async getCompressedBody(resp: any): Promise<any> {
-    return this.isMimeTypeImage(resp.headers['content-type']) ? await imagemin.buffer(resp.data, imageminOptions) : resp.data;
+    if (isBitmapImageMimeType(resp.headers['content-type'])) {
```
kelson42, line 594: I would remove this line and just `return true` in each optim conditional block (that will be quicker to execute.

src/Downloader.ts
```diff
@@ -586,19 +591,34 @@ class Downloader {
...
+      if (isWebpCandidateImage(this.webp, this.cssDependenceUrls.hasOwnProperty(resp.config.url), resp.headers['content-type'])) {
```
kelson42, line 595: `this.cssDependenceUrls.hasOwnProperty(resp.config.url)` should not be given as argument, but just part of the test:
```
isWebpCandidateImage(this.webp, resp.headers['content-type']) &&
  !this.cssDependenceUrls.hasOwnProperty(resp.config.url)
```

src/Downloader.ts
```diff
@@ -11,35 +11,44 @@ import axios, { AxiosRequestConfig } from 'axios';
...
-    // imageminJpegtran(),
+  ]
+})
+imageminOptions.set('image/gif', {
+  plugins: [
     imageminGifsicle({ optimizationLevel: 3, colors: 64 }),
-  ],
-};
+  ]
+})
```
kelson42, line 51: All of this is wrong, it was proper in last review. Please bring it back and just make small adaptation for imagemin-gif.

kelson42 (changes requested):

src/Downloader.ts
```diff
@@ -586,19 +591,34 @@ class Downloader {
...
 
   private async getCompressedBody(resp: any): Promise<any> {
-    return this.isMimeTypeImage(resp.headers['content-type']) ? await imagemin.buffer(resp.data, imageminOptions) : resp.data;
+    if (isBitmapImageMimeType(resp.headers['content-type'])) {
+      if (isWebpCandidateImage(this.webp, this.cssDependenceUrls.hasOwnProperty(resp.config.url), resp.headers['content-type'])) {
+        resp.data = await imagemin.buffer(resp.data, {
+          plugins: [
+            imageminWebp({ quality: 75 })
+          ]
+        });
```
kelson42, line 600: Instead of having one or the other imagemin options, you will probably have to create profiles. But all of this needs to be setup earlier.

kelson42 (changes requested):

src/util/dump.ts
```diff
@@ -42,11 +43,12 @@ export async function getAndProcessStylesheets(downloader: Downloader, links: Ar
                     if (!url.match('^data')) {
                         const filePathname = urlParser.parse(url, false, true).pathname;
                         if (filePathname) {
-                            const filename = pathParser.basename(filePathname).replace(/-.*x./, '.');
+                            const filename = pathParser.basename(filePathname).replace(/-.*x./, '.')
```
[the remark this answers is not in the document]
MananJethwani: that's actually outdated code corrected it afterward

src/util/misc.ts
MananJethwani: @kelson42 should I also remove webp argument from here?

MananJethwani: actually if we are using 2 profiles then there is no need at all for this function

src/Downloader.ts
MananJethwani: @kelson42 I don't get this part

src/util/misc.ts
kelson42: Probably, indeed, a bit unsure

src/Downloader.ts
kelson42: lets keep it like this

src/Downloader.ts
```diff
@@ -613,7 +643,7 @@ class Downloader {
   private async downloadImage(url: string, handler: any) {
     try {
       this.s3.downloadBlob(stripHttpFromUrl(url)).then(async (s3Resp) => {
-        if (s3Resp?.Metadata?.etag) {
+        if (s3Resp?.Metadata?.etag && s3Resp.Metadata.webp === (!this.cssDependenceUrls.hasOwnProperty(url) && this.webp ? 'webp' : '1')) {
```
kelson42, line 646: `&& s3Resp.Metadata.webp === (!this.cssDependenceUrls.hasOwnProperty(url) && this.webp ? 'webp' : '1'` should be removed. Within a run, we write everything with the same version tag.

src/Downloader.ts
```diff
@@ -635,7 +665,9 @@ class Downloader {
         // Check for the etag and upload
         const etag = this.removeEtagWeakPrefix(mwResp.headers.etag);
         if (etag) {
-          this.s3.uploadBlob(stripHttpFromUrl(url), mwResp.data, etag);
+          this.s3.uploadBlob(stripHttpFromUrl(url), mwResp.data, etag,
+            !this.cssDependenceUrls.hasOwnProperty(mwResp.config.url) && this.webp ? 'webp' : '1'
```
kelson42, line 669: Same here, `!this.cssDependenceUrls.hasOwnProperty(mwResp.config.url)` shoudl be removed. Everything should be written with the same tag.

src/S3.ts
```diff
@@ -42,11 +42,11 @@ class S3 {
         });
     }
 
-    public async uploadBlob(key: string, data: any, eTag: string) {
+    public async uploadBlob(key: string, data: any, eTag: string, version: string) {
         const params = {
             Bucket: this.bucketName,
             Key: key,
-            Metadata: {etag: eTag },
+            Metadata: {etag: eTag , version},
```
kelson42, line 49: This is a key value system, so `{etag: eTag , version: version}`

src/mwoffliner.lib.ts
```diff
@@ -443,6 +446,12 @@ async function execute(argv: any) {
       { type: 'css', moduleList: Array.from(cssModuleDependencies) },
     ];
 
+    if(downloader.webp)
```
kelson42, line 449: Wrong identifation, wrong sapce

src/util/dump.ts
```diff
@@ -142,3 +142,17 @@ export async function downloadAndSaveModule(zimCreator: ZimCreator, mw: MediaWik
         throw e;
     }
 }
+
+export async function downloadPolyfillModules(zimCreator: ZimCreator, downloader: Downloader) {
+    const { content } = await downloader.downloadContent(encodeURI('https://unpkg.com/webp-hero@0.0.0-dev.26/dist-cjs/polyfills.js'));
```
kelson42, line 147: In place of downloading we should have this as an `npm` dependence and read it. Will be more robust.

src/util/saveArticles.ts
```diff
@@ -796,14 +800,20 @@ async function templateArticle(parsoidDoc: DominoElement, moduleDependencies: an
         styleDependenciesList: string[],
     };
 
+    if (webp) {
+        jsDependenciesList.push('webpHeroPolyfill');
+        jsDependenciesList.push('webpHeroBundle');
```
kelson42, line 805: Do we really need both? I doubt a bit that will work... but don't have tested.

src/util/saveArticles.ts
```diff
@@ -915,6 +925,15 @@ function isSubpage(id: string, mw: MediaWiki) {
     return false;
 }
 
+function addWebpScript(webp: boolean) {
```
kelson42, line 928: It is not OK to have a script called `addWebpScript()` which add empty string. In general you should try to not add, change function calls. The less the functions change, the less arguments they have, the better.

kelson42 (changes requested):

src/S3.ts
MananJethwani: lint and codefactor were showing issue with it that's why I did it this way

MananJethwani: and js support if we write same name for key and value we can just write it like this

src/util/const.ts
```diff
@@ -1,9 +1,12 @@
...
+export const IMAGE_URL_REGEX = /(https?:\/\/.*\.(?:jpg|gif|png|svg|jpeg))/i;
 export const CONCURRENCY_LIMIT = 10;
-export const MIME_IMAGE_REGEX = /^image+[/-\w.]+$/;
+export const IMAGE_MIME_REGEX = /^image+[/-\w.]+$/;
 export const FIND_HTTP_REGEX = /^(?:https?:\/\/)?/i;
 export const DB_ERROR = 'internal_api_error_DBQueryError';
 export const WEAK_ETAG_REGEX = /^(W\/)/;
+export const WEBP_CANDIDATE_IMAGE_FILENAME_REGEX = /(jpeg|png|jpg)$/i;
+export const BITMAP_IMAGE_MIME_REGEX = /^image+[/-\w.]+(jpeg|png|gif)$/;
+export const WEBP_CANDIDATE_IMAGE_MIME_TYPE = /image+[/]+gif/;
```
kelson42: Shoudl be positive like `WEBP_CANDIDATE_IMAGE_FILENAME_REGEX`.

src/S3.ts
kelson42: ok...

kelson42 (changes requested):

kelson42:
@MananJethwani Small other problem, video poster URLs are not rewritten. In your ZIM, we still point to the `jpg` picture.

kelson42:
@Jaifroid We struggle to test with a browser which does not support wepb. Does the following one works well for the older one on Windows you use to test again?
[Uploading test_wpeb_1.zim.zip…]()

src/util/saveArticles.ts
MananJethwani: @kelson42 yes the function calling takes place polyfill.js but webpHeroBundle has something to do with conversions

src/util/dump.ts
```diff
@@ -142,3 +142,17 @@ export async function downloadAndSaveModule(zimCreator: ZimCreator, mw: MediaWik
         throw e;
     }
 }
+
+export async function importPolyfillModules(zimCreator: ZimCreator, downloader: Downloader) {
```
kelson42, line 146: `downloader` should be removed.

src/util/dump.ts
```diff
@@ -142,3 +142,17 @@ export async function downloadAndSaveModule(zimCreator: ZimCreator, mw: MediaWik
...
+    let content  = fs.readFileSync(process.cwd()+'/node_modules/webp-hero/dist-cjs/polyfills.js','utf8');
+    let text = content.toString();
+    let articleId = jsPath(config, 'webpHeroPolyfill');
+    let article = new ZimArticle({ url: articleId, data: text, ns: '-' });
```
kelson42, line 150: You can merge this in one line, I see no reason to use intermediary variables:
```
    let article = new ZimArticle({
        url: jsPath(config, 'webpHeroPolyfill'),
        data: fs.readFileSync(process.cwd()+'/node_modules/webp-hero/dist-cjs/polyfills.js','utf8').toString(),
        ns: '-'
    });
```

src/util/saveArticles.ts
kelson42: OK

src/util/saveArticles.ts
```diff
@@ -796,14 +802,20 @@ async function templateArticle(parsoidDoc: DominoElement, moduleDependencies: an
...
     const htmlTemplateDoc = domino.createDocument(
         htmlTemplateCode(articleId)
             .replace('__ARTICLE_CANONICAL_LINK__', genCanonicalLink(config, mw.webUrl.href, articleId))
             .replace('__ARTICLE_CONFIGVARS_LIST__', jsConfigVars !== '' ? genHeaderScript(config, 'jsConfigVars', articleId) : '')
             .replace(
                 '__ARTICLE_JS_LIST__',
                 jsDependenciesList.length !== 0
-                    ? jsDependenciesList.map((oneJsDep) => genHeaderScript(config, oneJsDep, articleId)).join('\n')
+                    ? `${jsDependenciesList.map((oneJsDep) => genHeaderScript(config, oneJsDep, articleId)).join('\n')} \n
+                        ${webp && addWebpScript()}`
```
kelson42, line 818: Seems to be an indentation problem here.

src/util/saveArticles.ts
```diff
@@ -915,6 +927,13 @@ function isSubpage(id: string, mw: MediaWiki) {
     return false;
 }
 
+function addWebpScript() {
+    return `<script>
+	        var webpMachine = new webpHero.WebpMachine()
+	        webpMachine.polyfillDocument()
+        </script>`
```
kelson42, line 934: Make a one line DOM not and indent it properly so this is aligned with the rest. `<script>var webpMachine = new webpHero.WebpMachine(); webpMachine.polyfillDocument();</script>`

kelson42 (changes requested):
@MananJethwani Looks good to me, could you please close this PR and open one directly in a new branch of this repository (you should have the permission now)?

Jaifroid:
> We struggle to test with a browser which does not support wepb. Does the following one works well for the older one on Windows you use to test again?

@kelson42 In the case of Kiwix JS, this effectively means Internet Explorer, Firefox OS, and Edge <18 (e.g. Windows mobile). However, none of these support Service Worker mode, which would be needed to use the in-builtnpolyfill automatically. Therefore, I would need to write a small routine to extract the polyfill from the ZIM and attach it to the DOM, which I'll try to do tomorrow. However, it would not be a true test of whether the polyfill would work automaticlally on, e.g., a webview. It would simply tell us that the polyfill included in the ZIM does work. It would not be able to tell us that it is correctly attached by code inside the ZIM, since I'd be writing my own code to attach. (I hope that makes sense!)

kelson42:
@Jaifroid I was absolutly ignorant about the fact that this polyfill was based on SW. Meanwhile we have tested with a current Safari on macOS (which does not have the wepb support either) and webp images are displayed properly. Therefore we have our answer. Concerning browsers missing support of both webp & SW, your proposal of Kiwix JS patch is really welcome. If you believe the patch could be successfully applicable in MWoffliner directly, I'm happy to do it as well.

@MananJethwani Please put a comment in the PR beside the `<script>` node introduction to secure we keep the path in the ZIM stable.

Jaifroid:
@kelson42 It's just that to run *any* JavaScript-from-the-ZIM automatically in Kiwix JS, we have to load it in Service Worker mode. Due to technical difficulties, it is very hard to load JS-from-the-ZIM by parsing and attaching each included JS file. However, it is easy if we know what we are looking for, and I think supporting webp is important enough to add some code that checks if the ZIM contains the polyfill and attaches it if it is found. This is not difficult to add, and will help us to continue supporting old browsers. I'll do this in https://github.com/kiwix/kiwix-js/issues/650 .

kelson42:
Superseeded by #1271

PR #1271, opened 2020-10-08
Add support to Webp bitmap image conversion

MananJethwani:
this solves the issue #1254


kelson42 (changes requested):
Images files don't have `.webp` added properly at their end if they come from S3.

Jaifroid:
Could I just check the file location we should look for this manually (in our jQuery mode?). I saw some hint in the PR [here](https://github.com/openzim/mwoffliner/pull/1271/files#diff-86a21593fa572f8bea0f0ffd3620cdc2R145), which seems to suggest - `-/webpHeroPolyfill/polyfills.js` and `-/webpHeroBundle/webp-hero.bundle.js` (though this is just a guess). If that is remotely correct, I wonder why we don't use:

`-/_api/webp/polyfills.js` and
`-/_api/webp/webp-hero.bundle.js`

to be consistent with #1237 ? Then I could make a generic `useKiwixAPI` function that would include attaching these scripts.

kelson42:
@Jaifroid This is not part of the API (we still don't have merged the version version). Paths are (and hopefully be stable at):
* `-/j/js_modules/webpHeroBundle.js`
* `-/j/js_modules/webpHeroPolyfill.js`

Jaifroid:
@kelson42 I've had a go at adding the patch to Kiwix JS, but the ZIM you provided `test_wpeb_1.zim` does not appear to have a landing page, and no articles in the title index. Since there is no article, there is no opportunity to attach the scripts to the article DOM. Do you have a test ZIM with at least one article containing webp images?

kelson42:
@Jaifroid Here a webp version of the Wikimed ZIM we have in our custom apps: http://mirror.download.kiwix.org/zim/.hidden/custom_apps/wikipedia_en_medicine-app_maxi_2020-10.zim

Issue #1026, opened 2019-12-17
Transparent PNG are not always displayed properly in dark mode


kelson42:
In dark mode, the bakground is black. If the PNG is a black design on transparent background, then nothing will be visible for the user. This is what happen with this article http://library.kiwix.org/wikipedia_fr_mathematics_maxi/A/Parabole on Android for example.

Jaifroid:
But how can mwoffliner help? The scraper should always scrape the file in the default mode, which is light. It would then be up to the OS or the reader to invert the image if it is either an SVG or a PNG based on an SVG. In Kiwix JS Windows, I have a rule (or two!) in the dark mode stylesheet that inverts the image in this situation, ensuring that equations and other images are visible in dark mode. Most dark modes avoid inverting the image, because it looks horrible to the user, but unfortunately some images must be inverted and I think the decision has to be made by the reader, not the scraper.

kelson42:
@Jaifroid By surrending the IMG in a `<div>` of the exact same size and with a white (or same colour like the background) background. In light mode, no difference, and in darmode, we will have a square with white bacgkround (so no transparency here).

Jaifroid:
@kelson42 IMHO this would be a dirty implementation because it would hard-code a light background into the source, negate the transparency in the image, and defeat the purpose of a dark mode (to avoid being blinded by white background at night, for example). It would be better to mark these images as needing special treatment by the reader (in the same way as SVG equations as stated in https://github.com/kiwix/kiwix-android/issues/1649 ) but not hard-code any specific colour background into them.

kelson42:
@Jaifroid I'm against any kind of hack at the reader level for obvious. As long as we don't have JS API, what I have proposed seems to be the best compromise to me between short and long term goals.

Jaifroid:
@kelson42 OK, but please be very careful not to surround equations (which are SVG) with a hard-coded white background. These are inline and are read as part of the text of the page. I give an example of equations rendered in dark mode below (from a ZIM without the MathML). Just imagine how horrible this would look if each of the equations on this page had a white background...

![image](https://user-images.githubusercontent.com/4304337/71160491-91a4c480-223f-11ea-9976-2b22787f2933.png)

kelson42:
@Jaifroid This is this problem or you can have content which is black on black... what is the best to your opinion?

Jaifroid:
[...]

PR #1279, opened 2020-10-18
added white stroke for transparent png in dark mode

MananJethwani:
this solves the issue #1026 
@kelson42 here is the zim file after applying white stroke

img2


kelson42 (changes requested):
Really elegant hacky solution. Hopefully we won’t discover any bad side effects. Have made a few remarks directly on Slack to polish the solution.

Ping @Jaifroid and @automactic as wll as they might be interestto have a look as well to this PR.

kelson42 (changes requested):
@MananJethwani This looks good from the principle, please provide a demo ZIM file to we can all test if it works fine. Please provide one before the fix and one with the fix, so we can see the difference.

MananJethwani:
@kelson42 give me some time I found a bug solving it first

MananJethwani:
[issue_1026.zip](https://github.com/openzim/mwoffliner/files/5402620/issue_1026.zip)
here are the files before and after applying white stroke to images
@kelson42  I had to add an inversion to cancel out the inversion applied by drop-shadow effect which inverted few images

MananJethwani:
@Jaifroid and @automactic did you have a look at zip file mentioned above is it fine?

kelson42 (changes requested):
Firefox normal mode:
image

Firefox dark mode:
image

This PR does not work.

kelson42:
@MananJethwani I don't believe this problem is solveable easily. Was an error to tag itt as "good-first-issue", sorry.

Issue #1040, opened 2020-03-05
Custom flavour extensions should be in the npm package


kelson42:
And before trying to load an external `js` file with `--customFlavour`, it should be checked in the internal `extensions` directory.

So:
* `--customFlavour=WiktionaryFR` should load `WikitionaryFR.js` in this extensions directory.
* `--customFlavour=foobar` should load `foobar.js` in the current directory (from where `mwoffliner` is launched by the user).
* `--customFlavour=foobar.js` should load `foobar.js` in the current directory (from where `mwoffliner` is launched by the user).
* `--customFlavour=/foobar.js` should load `foobar.js` in the root directory directory

LakmaNeha:
@kelson42 can I take up this issue?

LakmaNeha:
@kelson42 can I work on this?

kelson42:
@LakmaNeha yes

LakmaNeha:
@kelson42  If I understand correctly, except for the absolute path case(4th one), we always search for the extension first in the extensions directory and then, If we don't find in the extensions directory, then we are going to search in the current directory. Is that prioritisation right? or shall we search in the current directory first and then search in extensions directory. IMO searching for the file, first in the current directory and If not found, search in extensions directory makes sense, as this is a custom option provided to developer, we should prioritise developer's relative path first rather than our extension's path. What do you think?

kelson42:
@LakmaNeha You are right, local first.

LakmaNeha:
@kelson42 Also, if we can't locate the file, we are going to abort throwing an error message right? which is something like `couldn't locate the file in the current directory or extensions directory. Try giving the absolute path for the file.`

Secondly, as this may lead to file name conflict, I think showing some message like
`Using customProcessor - /some/absolute_path/to/customFlavour` would be informative and helps in debugging if something is wrong. What do you think?

kelson42:
@LakmaNeha yes

MananJethwani:
@kelson42 is this issue still open?.....I would like to work on it

PR #1283, opened 2020-10-22
resolve custom flavor paths

MananJethwani:
this solve #1040 
@kelson42 what things need to be done for the testing part do let me know.


kelson42 (changes requested):
* There is no automated tests
* I believe none of the scenarios described in the ticket works because of multiple errors
* Proper spaces are not always there
* Argument checking/sanitizing (and proper path setting) should happen in `src/sanitize-argument.ts`
* Dedicated `src/sanitize-argument.ts` seems not necessary

To conclude: put everything in `src/sanitize-argument.ts`, read the requirements accurately, test all the scenarios properly/automaticaly and read reread 3 times PR code before asking for review... and that should work :)

src/sanitize-argument.ts
```diff
@@ -5,11 +5,12 @@ import Redis from './Redis';
 import urlParser from 'url';
 import logger from './Logger';
 import { config } from './config';
-import fs, { readFileSync } from 'fs';
+import fs from 'fs';
 import * as QueryStringParser from 'querystring';
-import { isValidEmail, ensureTrailingChar } from './util';
+import { isValidEmail, getCustomFlavorPath } from './util';
```
kelson42, line 10: Please move the `getCustomFlavorPath` code here and rename it to `sanitize_custon_flavour`, You can then remove `src/util/customFlavor.ts`,

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
             'https://wikimedia.org/api/rest_v1/media/math/render/svg/34cbb1e27dae0c04fc794a91f2aa001aca7054c1',
             'Full Url when base and url both strtas with http/s');
 })
+
+test('Custom flavour path', async(t) => {
+
+    // checks in current working directory.
+    let createStream = fs.createWriteStream(pathParser.resolve(process.cwd(), 'testCustomFlavour.js'));
```
kelson42, line 74: In general, you should avoid to write randomly in the current directory or in the code directory. You never know what kind of side effect you could have (like the file you want to create exists already, people works during the test and will see file appearing/disappearing). I fyou have to create tmp files. You create them in the OS temporary directory and you secure that even if the test for some reason is stopped in the middle then the files disappear by using something like https://www.npmjs.com/package/tmp. The good news is that I don't believe you need that at all here.

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
...
+
+    t.equal(getCustomFlavorPath('testCustomFlavour.js'), pathParser.resolve(process.cwd(), 'testCustomFlavour.js'),
```
kelson42, line 76: Why not moving in `extensions` and testing with `wiktionary_fr.js`? Seems lighter!?

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
...
+        'Custom flavour in working directory.');
+
+    // checks when current and working directory have file with same name(preference to working directory).
+    createStream = fs.createWriteStream(pathParser.resolve(process.cwd(), 'extensions/testCustomFlavour.js'));
```
kelson42, line 80: Same here, but moving one directory level higher.

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
...
+
+    t.equal(getCustomFlavorPath('testCustomFlavour'), pathParser.resolve(process.cwd(), 'testCustomFlavour.js'),
+        'When file with same name exist in extensions directory.');
```
kelson42, line 83: etc... You have a module file which already exists, use it instead of creating one only for the test.

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
...
+
+    // checks in extension directory.
+    t.equal(getCustomFlavorPath('wiktionary_fr.js'), pathParser.resolve(process.cwd(), 'extensions/wiktionary_fr.js'),
+        'Custom flavour in extensions directory.');
+
+    t.equal(getCustomFlavorPath('wiktionary_fr'), pathParser.resolve(process.cwd(), 'extensions/wiktionary_fr.js'),
+        'Custom flavour in extension directory without js extension.');
+
+    // checks in absolute path.
+    createStream = fs.createWriteStream(pathParser.resolve(process.cwd(), 'extensions/testCustomFlavour.js'));
```
kelson42, line 97: You should not use `process.cwd()` here because you have no clue where the test is started from. Secure the path by changing the current directory for the test which test a local/same_dir `*.js` file... Otherwise you need to build the paths related to something which is fix, like the root of the codebase. You made a similar error by reading the files for the Hero polyfill.

test/unit/util.test.ts
```diff
@@ -65,3 +67,45 @@ test('Get full URL', async(t) => {
...
+        'Positive check with absolute path.');
+
+    t.equal(getCustomFlavorPath(pathParser.resolve(process.cwd(), 'extensions/negativeTest.js')),
+        null, 'Negative test for absolute path.');
+
+    createStream.end();
+    fs.unlinkSync(pathParser.resolve(process.cwd(), 'extensions/testCustomFlavour.js'));
+
+    // negative scenario
+    t.equal(getCustomFlavorPath('wrongCustomFlavour.js'), null, 'Returning null when file doesnt exist.')
```
kelson42, line 110: To conclude, you should never have to create any temporary file and you should not used `process.cwd()`.

kelson42 (changes requested):

test/unit/util.test.ts
MananJethwani: @kelson42 this test is for case one(searching for file in current wokring directory so I guess this is the right place isn't it?)

test/unit/util.test.ts
MananJethwani: yes here I will move directly instead of moving from current working directory

test/unit/util.test.ts
MananJethwani: @kelson42 this is to test if correct module is picked when a file with same name exist in both current working directory and extension directory I shouldn't have used current working directory part here but I guess the file is still needed.

test/unit/util.test.ts
MananJethwani: @kelson42 this test is for case one(searching for file in current wokring directory) so I guess this file is needed and we would need tmp folder as well then.

kelson42: OK, then `tmp` package to create a directory in OS temporary directory and inside put your fake extension (with `tmp`)... and do whatever is necessary.

test/unit/util.test.ts
kelson42: ok, see my previous comment.

test/unit/util.test.ts
kelson42: ok

package.json
```diff
@@ -102,6 +103,7 @@
     "service-runner": "^2.7.8",
     "sharp": "^0.26.2",
     "swig-templates": "^2.0.3",
+    "tmp": "^0.2.1",
```
kelson42, line 106: AFAIK, this should be a dev dependence, not a runtime?!

src/sanitize-argument.ts
```diff
@@ -24,7 +26,15 @@ export async function sanitize_all(argv: any) {
   // sanitizing speed
   sanitize_speed(_speed);
 
-  // sanitizing s3
+  // sanitizing custom flavour
+  if (argv.customFlavour) {
+    argv.customFlavour = sanitize_customFlavour(argv.customFlavour);
+    if (!argv.customFlavour) {
+      throw new Error('Custom Flavour not found');
+    }
```
kelson42, line 34: I believe this can be done in one line

src/sanitize-argument.ts
```diff
@@ -115,4 +126,31 @@ export async function sanitize_customZimFavicon(customZimFavicon:any)
...
+ * 2. mwoffliner's extensions directory
+ * 3. absolute path(for root folder)
+ *
+ * Note: CustomFlavour doesn't necessarily need be given with extension(.js)
+ * like --customFlavour=wiktionary_fr. Hence, .js is explicitly added for
+ * path resolution.
+ */
+
+export function sanitize_customFlavour(customFlavour: string): string {
+  customFlavour += customFlavour.substr(customFlavour.length - 3) !== '.js' ? '.js' : '';
```
kelson42, line 144: Better use `path.extname()`

src/sanitize-argument.ts
```diff
@@ -115,4 +126,31 @@ export async function sanitize_customZimFavicon(customZimFavicon:any)
...
+  const possiblePaths = [
```
kelson42, line 145: I have no problem to avoid the temporary `possiblePaths`.

test/unit/util.test.ts
```diff
@@ -65,3 +71,32 @@ test('Get full URL', async(t) => {
             'https://wikimedia.org/api/rest_v1/media/math/render/svg/34cbb1e27dae0c04fc794a91f2aa001aca7054c1',
             'Full Url when base and url both strtas with http/s');
 })
+
+test('Custom flavour path', async(t) => {
+
+    // checks in current working directory.
+    const tmpObj = tmp.fileSync({ postfix: '.js' });
+    process.chdir('/tmp');
```
kelson42, line 79: No, you should change directory based on values of `tmpObj`. You have no garanty that this is in `/tmp`.

test/unit/util.test.ts
```diff
@@ -65,3 +71,32 @@ test('Get full URL', async(t) => {
...
+
+    // checks in extension directory.
+    t.equal(sanitize_customFlavour('wiktionary_fr.js'), pathParser.resolve(__dirname, '../../extensions/wiktionary_fr.js'),
+        'Custom flavour in extensions directory.');
+
+    t.equal(sanitize_customFlavour('wiktionary_fr'), pathParser.resolve(__dirname, '../../extensions/wiktionary_fr.js'),
+        'Custom flavour in extension directory without js extension.');
+
+    // checks in absolute path.
+
```
kelson42, line 92: Why an empty line?

kelson42 (changes requested):

src/sanitize-argument.ts
MananJethwani: exceptions can't be included in inline conditions using `||`

kelson42: I really wonder about that... but ok

kelson42 (approved):
GOOD

Issue #1289, opened 2020-10-28
On WikiMed ZIM it looks likes redirects are scraped but links to redirect are removed


kelson42:
For example:
in https://en.wikipedia.org/wiki/Book:Cancer, the link to "Colon cancer" exists

In the ZIM:
http://library.kiwix.org/wikipedia_en_medicine_maxi/A/Book%3ACancer it is black,

although:
http://library.kiwix.org/wikipedia_en_medicine_maxi/A/Colon%20cancer exists.

I believe there is a problem of article id comparison (with or without space/underscore).

devangmukherjee:
Hi, I would like to work on this. Can you please tell me where can I find the code related to this?

PR #1292, opened 2020-10-31
corrected redirection

MananJethwani:
this solves #1289


kelson42 (changes requested):
The code looks good but one of test needs to be extended so next such a problem occurs, we can detect it.

kelson42 (approved):

PR #1296, opened 2020-11-05
Mime type check

### LINKED ISSUE #1183: Correctly detect mimetype of article.
Few zim files (as wikipedia_ar_history_maxi_2020-05.zim) have complex url for image (`m/osm-intl,15,33.333055555556,69.916944444444,300x300.png?lang=ar&domain=ar.wikipedia.org&title=حصار+خوست&groups=_41d389aefe5f67866d9b480e5c44559d54fb9683.svg`).
Those images are png but as the url ends with svg, it is detected as svg and :
- added to a compressed cluster.
- the wrong mimetype is returned to the reader.

Mwoffliner should use a better method to detect the mimetype of the data:
- remove what after `?` before detect the mimetype
- check the content of the data and not just the url
- Use other method (As we are on web world, there is no guaranty at all that an url have a extension we can base on to detect the mimetype)

MananJethwani:
this resolves #1183

### DIFF
--- a/src/util/const.ts
+++ b/src/util/const.ts
@@ -7,6 +7,6 @@ export const IMAGE_MIME_REGEX = /^image+[/-\w.]+$/;
 export const FIND_HTTP_REGEX = /^(?:https?:\/\/)?/i;
 export const DB_ERROR = 'internal_api_error_DBQueryError';
 export const WEAK_ETAG_REGEX = /^(W\/)/;
-export const WEBP_CANDIDATE_IMAGE_FILENAME_REGEX = /((jpeg|png|jpg)$|(jpeg%3F|png%3F|jpg%3F))/i;
+export const WEBP_CANDIDATE_IMAGE_FILENAME_REGEX = /((\.jpeg|\.png|\.jpg)$|(\.jpeg%3F|\.png%3F|\.jpg%3F))/i;
 export const BITMAP_IMAGE_MIME_REGEX = /^image+[/-\w.]+(jpeg|png|gif)$/;
 export const WEBP_CANDIDATE_IMAGE_MIME_TYPE = /image+[/]+(jpeg|png)/;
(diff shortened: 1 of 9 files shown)


src/util/const.ts
```diff
@@ -7,6 +7,6 @@ export const IMAGE_MIME_REGEX = /^image+[/-\w.]+$/;
 export const FIND_HTTP_REGEX = /^(?:https?:\/\/)?/i;
 export const DB_ERROR = 'internal_api_error_DBQueryError';
 export const WEAK_ETAG_REGEX = /^(W\/)/;
-export const WEBP_CANDIDATE_IMAGE_FILENAME_REGEX = /(jpeg|png|jpg)$/i;
+export const WEBP_CANDIDATE_IMAGE_FILENAME_REGEX = /((\.jpeg|\.png|\.jpg)$|(\.jpeg%3F|\.png%3F|\.jpg%3F))/i;
```
kelson42, line 10: I'm sure there is a more elegant regex than this.

kelson42: It should be `WEBP_CANDIDATE_IMAGE_URL_REGEX` as it test a regex right?

kelson42: I guess `IMAGE_URL_REGEX` should be fixed as well?

test/unit/webpAndRedirection.test.ts
```diff
@@ -42,6 +43,18 @@ Real-time computer graphics`;
 
     const downloader = new Downloader({ mw, uaString: '', speed: 1, reqTimeout: 1000 * 60, noLocalParserFallback: false, forceLocalParser: false, webp: false, optimisationCacheUrl: '' });
 
+    t.assert(shouldConvertImageFilenameToWebp('../I/m/osm-intl%2C9%2C52.2789%2C8.0431%2C300x300.png%3Flang.svg', true),
+        'adding webp to fileName having png before arguments');
+    t.assert(shouldConvertImageFilenameToWebp('../I/m/osm-intl%2C9%2C52.2789%2C8.0431%2C300x300.jpg%3Flang.svg', true),
+        'adding webp to fileName having jpg before arguments');
+    t.assert(shouldConvertImageFilenameToWebp('../I/m/osm-intl%2C9%2C52.2789%2C8.0431%2C300x300.jpeg%3Flang.svg', true),
+        'adding webp to fileName having jpeg before arguments');
```
kelson42, line 51: We should have test with a real question mark?!

kelson42 (changes requested):

src/util/const.ts
kelson42: I have fixed all of this.

PR #1299, opened 2020-11-06
changes to solve rendering pb of license in footer

### LINKED ISSUE #1233: The table is not displayed correctly + there are no licenses at the bottom
**wikipedia_en_computer_maxi_2020-07.zim**
The table in the article about the [**Ubuntu**](https://en.wikipedia.org/wiki/Ubuntu#Releases) operating system is displayed incorrectly:
Ubuntu releases-Kiwix

Here's what the table should look like ([**Ubuntu releases**](https://en.wikipedia.org/wiki/Ubuntu#Releases)):
Ubuntu releases

In addition, some articles from the above-mentioned **ZIM** file do not contain information about licenses at the bottom. This label is missing in the articles: **[Ubuntu](https://en.wikipedia.org/wiki/Ubuntu)**, **[Ubuntu version history](https://en.wikipedia.org/wiki/Ubuntu_version_history)**, etc.:
bottom

MananJethwani:
this resolves #1233

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -867,48 +867,45 @@ async function templateArticle(parsoidDoc: DominoElement, moduleDependencies: an
 
     /* Set footer */
     const div = htmlTemplateDoc.createElement('div');
-    try {
-        if (articleDetail.revisionId) {
-            /* Revision date */
-            const date = new Date(articleDetail.timestamp);
-            const lastEditedOnString = date ?
-                interpolateTranslationString(dump.strings.LAST_EDITED_ON, {
-                    date: date.toISOString().substring(0, 10),
-                }) : null;
 
-            const creatorLink =
-                `<a class="external text" ` +
-                `${lastEditedOnString ? `title="${lastEditedOnString}"` : ''} ` +
-                `href="${mw.webUrl.href}?title=${encodeURIComponent(articleId)}&oldid=${articleDetail.revisionId}">` +
-                `${dump.mwMetaData.creator}</a>`;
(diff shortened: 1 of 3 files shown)


kelson42 (changes requested):
Could we please have an explanation in the commit message about what is exactly going on here? I guess such a bug had other impacts? I would like as well to remove the `try{} catch{}` at the place where the foother has been used because it has masked the error and I don't see any legitimate use case why we should have it.

kelson42 (approved):
Looks good, I have in addition:
* Rebased on git master
* Fix a test which was a bit too strict
* Removed a test which seems useless

src/util/saveArticles.ts
```diff
@@ -868,33 +868,31 @@ async function templateArticle(parsoidDoc: DominoElement, moduleDependencies: an
...
+    const date = new Date(articleDetail.timestamp);
+    const lastEditedOnString = date ?
+        interpolateTranslationString(dump.strings.LAST_EDITED_ON, {
+            date: date.toISOString().substring(0, 10),
+        }) : null;
+
+    const creatorLink =
+        `<a class="external text" ` +
+        `${lastEditedOnString ? `title="${lastEditedOnString}"` : ''} ` +
+        `href="${mw.webUrl.href}?title=${encodeURIComponent(articleId)}&oldid=${articleDetail.revisionId}">` +
```
MananJethwani, line 881: if revision Id is not present shouldn't we add a link to main wikipedia page......should this be done here as well?

kelson42: What is a scenario for not having a revisionId?

PR #1306, opened 2020-11-14
added condition for only one para

### LINKED ISSUE #838: If only one section and smal screen width, then the section should not be collapsed
Uh just to be clear - a user just complained that top-level section titles were collapsed by default (and it's hard to disagree with him). Am I right to understand that this will be corrected to a wikipedia-like mobile view?
comparaison
@kelson42 's to be honest I'm not sure I even understand the issue you described, as expanded top-level sections and collapsed everything below that seems to be the correct approach :-/

https://github.com/openzim/mwoffliner/issues/786#issuecomment-502667450

MananJethwani:
this resolves #838

### DIFF
--- a/res/script.js
+++ b/res/script.js
@@ -10,13 +10,16 @@ window.onload = function () {
         }
     });
 
-    /* If small screen size and contains section */
+    /* If small screen size and contains section(s) */
     if (window.innerWidth < 720 && $('details')) {
+
         /* Find the highest level section in window */
-        const HighestLevel = Math.min(...$('details').map( function() {return $(this).attr('data-level');}).get());
+        const sectionTopLevel = Math.min(...$('details').
+            map( function() { return $(this).attr('data-level'); }).get());
+
         /* Collapse all highest level section if more than one */
-        if($(`details[data-level=${HighestLevel}]`).length !== 1){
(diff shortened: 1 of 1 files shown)


res/script.js
```diff
@@ -14,6 +14,13 @@ window.onload = function () {
         $('details[data-level=2]').attr('open', false);
     }
 
+    if (window.innerWidth < 720 && $('details')) {
+        var HighestLevel = Math.min(...$('details').map( function() {return $(this).attr('data-level');}).get());
+        if($(`details[data-level=${HighestLevel}]`).length == 1){
+            $(`details[data-level=${HighestLevel}]`).attr('open', true);
+        }
+    }
+
```
kelson42, line 23: The approach is probably OK, but I believe we could do better:
* Put everything under the `if (window.innerWidth < 720)` test to avoid duplicate test
* Put comments for each scenario ( we have two scnearios here)
* `$('details[data-level=2]')` is wrong as well, as we need to apply to the toplevel sections (could be different level than 2).
* We should probably have: if only one top section then do nothing (otherwise open everything).

kelson42 (changes requested):

res/script.js
MananJethwani: ok

kelson42 (approved):

PR #1308, opened 2020-11-16
added correct no of ../ and removed parentPath in subpages

### LINKED ISSUE #1094: Navigation bar URLs are wrong (with subpages)
Navigation bar URLs are not properly rewritten in the navigation bar, see for example at the top of the page:
http://library.kiwix.org/edutechwiki_fr_all_maxi/A/B%C3%A9haviorisme/l'apprentissage_par_essai_et_erreur_d'Edward_Lee_Thorndike

Remark: This is one new ticket in long tail of ticket regarding URL rewritting. I can only repeat mysel: "This has to be solved extremly rigoursly and automated tested in the same manner".

MananJethwani:
this resolves #1094 
@kelson42 I removed parentPath as it was not being updated and for each iteration it was reset to empty string(as mapping is different from loops)

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -855,19 +855,17 @@ async function templateArticle(parsoidDoc: DominoElement, moduleDependencies: an
         const parents = articleId.split('/');
         parents.pop();
         let subpages = '';
-        let parentPath = '';
         await Promise.all(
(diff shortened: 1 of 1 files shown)


kelson42 (approved):
Seems to work.

PR #1310, opened 2020-11-19
added ci.yml for github action CI

### LINKED ISSUE #1174: Move our CI to Github Actions
This is an ongoing move in the whole organization. Reasons are simply that Github Actions is better than Travis CI.

MananJethwani:
@kelson42  this resolves #1174


kelson42:
@MananJethwani Can you please remove travis and fix codefactor?

MananJethwani:
@kelson42 done

kelson42:
@MananJethwani we should as well make the local paraoid true and stopping him stops the whole process, as long as th real problem is not fixed.

kelson42:
What about nodejs 10?

MananJethwani:
@kelson42 should I remove that file from testing by changing file name?
and for nodejs 10 should I include version 10,12 and 14 instead of 12,13 and 14

kelson42:
* No keep, the test, but makes the code just returning without running through all the tests
* Just add nodejs 10

MananJethwani:
@kelson42 actually all tests passes in that scenario......the test doesn't terminate as parsoid doesn't shutdown......if we go through the scrapping process it wouldn't terminate I could return before scrapping in that case if you want

MananJethwani:
@kelson42 any idea why this fails in nodejs 10?

kelson42:
@MananJethwani No, you are the javascript/typescraipe expert :)

MananJethwani:
@kelson42 done:sweat_smile:.....please review

MananJethwani:
@kelson42 please have a look when free

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
+name: Node.js CI
+
+on:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 0'
```
kelson42, line 6: Why running on cron? I don't think we need that. What we need is to secure that each time we merge something to master the CI is ran. This is to secure the codecov page (see Readme.md).

kelson42 (changes requested):

.github/workflows/ci.yml
MananJethwani: @kelson42 it's a weekly check in case any of our test develops a problem it could be noticed and solved(and not noticed when someone else  makes a PR).....but if you want I can remove it

PR #1317, opened 2020-12-03
added proper support for logging in for mediawiki 1.28+

### LINKED ISSUE #42: Login does not work with Mediawiki 1.28+
It seams as since Mediawiki 1.27 there are new api's for login. Could you please add support for those new api's?

MananJethwani:
this resolves #42

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -67,7 +67,7 @@ class MediaWiki {
       await axios(this.apiUrl.href, {
         data: qs.stringify({
           action: 'login',
-		      format: 'json',
+          format: 'json',
           lgname: this.username,
           lgpassword: this.password,
           lgtoken: JSON.parse(content.toString()).query.tokens.logintoken,
@@ -82,10 +82,11 @@ class MediaWiki {
         if (resp.data.login.result !== 'Success') {
           throw new Error('Login Failed');
         }
-        downloader.arrayBufferRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
-        downloader.jsonRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
-        downloader.streamRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
+
         downloader.loginCookie = resp.headers['set-cookie'].join(';');
+        downloader.arrayBufferRequestOptions.headers.cookie = downloader.loginCookie;
+        downloader.jsonRequestOptions.headers.cookie = downloader.loginCookie;
+        downloader.streamRequestOptions.headers.cookie = downloader.loginCookie;
       })
       .catch((err) => {
         throw err;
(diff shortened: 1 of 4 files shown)


kelson42 (changes requested):
I believe:
* If this does not work with all user/password, then the documentation should be updated
* I see no code to secure that all requests done later to the api are using the cookie.

src/MediaWiki.ts
```diff
@@ -54,24 +56,37 @@ class MediaWiki {
...
+        },
+        method: 'POST',
+      })
+      .then((resp) => {
+        if (resp.data.login.result !== 'Success') {
           throw new Error('Login Failed');
         }
-        downloader.loginCookie = `${jsonResponse.cookieprefix}_session=${jsonResponse.sessionid}`;
-      }
+        downloader.arrayBufferRequestOptions.headers.Cookie = resp.headers['set-cookie'].join(';');
```
MananJethwani, line 85: @kelson42 I set cookies here so that all request to api are being made properly was I supposed to do anything else?

kelson42: @MananJethwani Which piece o code secures this cookie is use for all requests?

MananJethwani: ` downloader.arrayBufferRequestOptions.headers.Cookie = resp.headers['set-cookie'].join(';');`
I am not sure that this works yet still checking on private wiki but mostprobably should

MananJethwani: sorry my bad should have changed `downloader.loginCookie`

src/MediaWiki.ts
```diff
@@ -54,24 +56,40 @@ class MediaWiki {
...
+      })
+      .then((resp) => {
+        if (resp.data.login.result !== 'Success') {
           throw new Error('Login Failed');
         }
-        downloader.loginCookie = `${jsonResponse.cookieprefix}_session=${jsonResponse.sessionid}`;
-      }
+        downloader.arrayBufferRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
+        downloader.jsonRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
+        downloader.streamRequestOptions.headers.cookie = resp.headers['set-cookie'].join(';');
```
MananJethwani, line 87: had to add this all for securing cookies in all requests after login

src/mwoffliner.lib.ts
```diff
@@ -187,6 +187,9 @@ async function execute(argv: any) {
     webp,
   });
 
+  /* perform login */
+  await mw.login(downloader);
```
MananJethwani, line 191: shifted this here as in many cases site info might be restricted as well without login and login should be done before making most basic requests

kelson42: Good

kelson42 (approved):

PR #1321, opened 2020-12-06
Create dedicated Docker 'scraper' user

### LINKED ISSUE #995: Docker user should not be root
This makes it otherwise impossible to use NFS on Wikimedia VPS as cache. On the top of this, this is a security recommendation.

kelson42:
Fixes #995

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,4 +31,11 @@ ENV REDIS /dev/shm/redis.sock
 RUN printf '#!/bin/bash\n/usr/local/bin/mwoffliner --redis=$REDIS "$@"' > /usr/local/sbin/mwoffliner
 RUN chmod +x /usr/local/sbin/mwoffliner
 
+RUN useradd scraper
+RUN usermod -u 4242 scraper
+RUN mkdir -p /data
+WORKDIR /data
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin This is an attempt to make our Docker a bit more secure. But this has for direct consequence that the mounted volumed should be writeable by the new user `scraper`. Not sure as well about the Redis daemon... which user runs it? Should we change it?

rgaudin (changes requested):
- most rootless images defaults to user `1001` as this is what a regular distro would create so it has more chances to work out of the box.
- rootless implies choosing a user so uid and gid might be anything
- `/data` is gonna be mounted and there's no guarantee you'll be able to write on it. You may want to check that in entrypoint to provide a quick and clear feedback should permissions not be adequate. I don't know how quick and clear mwoffliner is gonna fail in this case.
- README may need to be updated to inform about write requirements on mounted volume source.

benoit74:
This PR is stalled for too long, and it makes too naive assumptions about the consequence (especially in the Zimfarm, where user will probably not have sufficient rights to write to output directory).

Closing for the time being, I do not expect to see this being merged soon.

Issue #1318, opened 2020-12-04
Wpes should not use parsoid 


kelson42:
https://farm.openzim.org/pipeline/506d67c207825269cd729cf5/debug

Why api is not detected?

kelson42:
This was a configuration problem.

kelson42:
@MananJethwani Would you be able to check if the `--customMainpage` does not point to a HTTP 404 or even a redirection? We should in `sanitizeArguments` be able to verify that it gets a proper HTTP 200 response.

MananJethwani:
@kelson42 sure

PR #1324, opened 2020-12-08
fixing url check for mobile page

### LINKED ISSUE #1322: Missing h1 article heading if customMainPage is not the exactly articleId
For example, with the `wikipedia_en_ray_charles_maxi_2020-12.zim`, none of the articles have the h1 header at the top of the page (see screenshot of article from December ZIM [top] compared with article for November ZIM [bottom]).

With `wikipedia_en_computer_maxi_2020-12.zim` some articles have h1 and others don't.

**Ray Charles December 2020:**

image

**Ray Charles November 2020:**

image

MananJethwani:
this temporarily resolves #1322

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -197,7 +197,7 @@ async function execute(argv: any) {
   }
   const mainPage = customMainPage || (articleList ? '' : mwMetaData.mainPage);
 
-  await downloader.checkCapabilities(mainPage);
+  await downloader.checkCapabilities(mwMetaData.mainPage);
   await downloader.setBaseUrls();
 
   const redis = new Redis(argv, config);
--- a/src/sanitize-argument.ts
+++ b/src/sanitize-argument.ts
@@ -8,7 +8,7 @@ import logger from './Logger';
 import { config } from './config';
 import fs from 'fs';
 import * as QueryStringParser from 'querystring';
-import { isValidEmail } from './util';
+import { isValidEmail, DEFAULT_WIKI_PATH } from './util';
 
 export async function sanitize_all(argv: any) {
 
@@ -19,6 +19,7 @@ export async function sanitize_all(argv: any) {
     mwUrl,
     customZimFavicon,
     optimisationCacheUrl,
+    mwWikiPath,
   } = argv;
 
   const cpuCount = os.cpus().length;
@@ -54,6 +55,18 @@ export async function sanitize_all(argv: any) {
   await sanitize_mwUrl(mwUrl).catch((err)=>{
     throw err;
   });
+
+  // sanitize Custom Main Page
+  if (argv.customMainPage) {
+    argv.customMainPage = argv.customMainPage.replace(/ /g,'_');
+    const customMainPageUrl = `${mwUrl}/${mwWikiPath || DEFAULT_WIKI_PATH}${argv.customMainPage}`;
+    await axios.get(`${customMainPageUrl}`, { maxRedirects: 0 })
+    .then((res) => {
+      if (res.status !== 200) {
+        throw new Error(`customMainPage doesn't return 200 status code for url ${customMainPageUrl}`);
+      }
+    });
+  }
 
   // sanitizing adminEmail
   sanitize_adminEmail(adminEmail);
--- a/src/util/const.ts
+++ b/src/util/const.ts
@@ -11,3 +11,4 @@ export const WEAK_ETAG_REGEX = /^(W\/)/;
 export const WEBP_CANDIDATE_IMAGE_URL_REGEX = /(\.jpeg|\.png|\.jpg)(\?.*)?$/i;
 export const BITMAP_IMAGE_MIME_REGEX = /^image+[/-\w.]+(jpeg|png|gif)$/;
 export const WEBP_CANDIDATE_IMAGE_MIME_TYPE = /image+[/]+(jpeg|png)/;
+export const DEFAULT_WIKI_PATH = 'wiki/'
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -4,7 +4,7 @@ import logger from './Logger';
 import * as util from './util';
 import * as domino from 'domino';
 import type Downloader from './Downloader';
-import { ensureTrailingChar } from './util';
+import { ensureTrailingChar, DEFAULT_WIKI_PATH } from './util';
 
 
 class MediaWiki {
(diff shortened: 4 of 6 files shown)


kelson42 (changes requested):
Thank you for proposing quickly a fix, but would that not be healthier to just check/fix things a the sanitize step?

src/mwoffliner.lib.ts
```diff
@@ -197,7 +197,7 @@ async function execute(argv: any) {
   }
   const mainPage = customMainPage || (articleList ? '' : mwMetaData.mainPage);
 
-  await downloader.checkCapabilities(mainPage);
+  await downloader.checkCapabilities(mainPage, mwMetaData.mainPage);
```
kelson42, line 200: Why adding a new argument? I thought we agre we should always test with `mwMetaData.mainPage`?

src/sanitize-argument.ts
```diff
@@ -107,6 +122,30 @@ export function sanitize_redis(argv:any)
...
 }
 
+export async function checkApiAvailabilty(url: string): Promise<boolean>{
+  try {
+    const resp = await axios.get(url, { maxRedirects: 0 });
+    return resp.status === 200 && !resp.headers['mediawiki-api-error'];
+  } catch (err) {
+    return false;
+  }
+}
```
kelson42, line 132: Full code duplicate, this should never happen.

src/sanitize-argument.ts
```diff
@@ -107,6 +122,30 @@ export function sanitize_redis(argv:any)
...
+  const veApiUrl = new URL(`${apiUrl.href}action=visualeditor&mobileformat=html&format=json&paction=parse&page=`);
+  const desktopRestApiUrl = new URL(ensureTrailingChar(new URL(restApiPath ?? 'api/rest_v1/page/html', baseUrl.href).toString(), '/'));
+  const baseUrlForMainPage = checkApiAvailabilty(desktopRestApiUrl.href + encodeURIComponent(customMainPage)) ? desktopRestApiUrl.href :
+                             checkApiAvailabilty(veApiUrl.href) ? veApiUrl.href :
+                             undefined;
+  if (!checkApiAvailabilty(baseUrlForMainPage + encodeURIComponent(customMainPage)) && noLocalParserFallback) {
+    throw new Error(`Failed to verify status code 200 on custom main page`)
+  }
+  return customMainPage.replace(/ /g,'_')
+}
```
kelson42, line 147: In  #1318, I just ask to test if the `customMainPage` article exists at all. I'm sure we can do a lot simpler. Just checking if `baseUrl + customMainPage` gives HTTP 200 should be sufficient, shouldn't it!?

kelson42 (changes requested):

src/sanitize-argument.ts
MananJethwani: ok

src/mwoffliner.lib.ts
MananJethwani: that info is not available in downloader class need to pass that here

MananJethwani: ok I get it sorry

src/sanitize-argument.ts
kelson42: I don't think you need that anyway, you just need to make a simple get request on a URL and check if code is HTTP 200. You can do that directly with Axios.

src/sanitize-argument.ts
```diff
@@ -107,6 +120,19 @@ export function sanitize_redis(argv:any)
...
+    if(res.status !== 200) {
+      throw new Error('Failed to verify status code 200 on custom main page')
+    }
+  })
+  .catch((err) => {
+    throw err;
+  });
+  return customMainPage.replace(/ /g,'_');
+}
+
```
kelson42, line 135: This function should be remove as most of the line is the exception mgmt, which is duplicate with lin 61.

We should not have this `wiki` here, this is a default value which should be handled only in the Mediawiki module.

`replace(/ /g,'_')` should not be done twice. Only once.

kelson42 (changes requested):

src/sanitize-argument.ts
```diff
@@ -55,6 +56,20 @@ export async function sanitize_all(argv: any) {
     throw err;
   });
 
+  // sanitize Custom Main Page
+  if (argv.customMainPage) {
+    await axios.get(`${mwUrl}/wiki/${argv.customMainPage.replace(/ /g,'_')}`)
```
kelson42, line 61: * we should keep the space replacement
* you shoudl not hardcode the path `wiki`, you should use whatever hase been given in Mediawiki.

kelson42 (changes requested):

src/sanitize-argument.ts
```diff
@@ -55,6 +56,20 @@ export async function sanitize_all(argv: any) {
     throw err;
   });
 
+  // sanitize Custom Main Page
+  if (argv.customMainPage) {
+    argv.customMainPage = argv.customMainPage.replace(/ /g,'_');
+    await axios.get(`${mwUrl}/${mwWikiPath || defaultWikiPath}${argv.customMainPage.replace(/ /g,'_')}`)
```
kelson42, line 62: Underscore replacement is not necessary

src/util/const.ts
```diff
@@ -10,4 +10,5 @@ export const DB_ERROR = 'internal_api_error_DBQueryError';
 export const WEAK_ETAG_REGEX = /^(W\/)/;
 export const WEBP_CANDIDATE_IMAGE_URL_REGEX = /(\.jpeg|\.png|\.jpg)(\?.*)?$/i;
 export const BITMAP_IMAGE_MIME_REGEX = /^image+[/-\w.]+(jpeg|png|gif)$/;
-export const WEBP_CANDIDATE_IMAGE_MIME_TYPE = /image+[/]+(jpeg|png)/;
\ No newline at end of file
+export const WEBP_CANDIDATE_IMAGE_MIME_TYPE = /image+[/]+(jpeg|png)/;
+export const defaultWikiPath = 'wiki/'
```
kelson42, line 14: Constant name should be uppercase

src/sanitize-argument.ts
kelson42, line 62: Redirects should be allowed in a ios request

src/sanitize-argument.ts
```diff
@@ -55,6 +56,20 @@ export async function sanitize_all(argv: any) {
...
+    .then((res) => {
+      if(res.status !== 200) {
+        throw new Error(`customMainPage doesn't return 200 status code`);
```
kelson42, line 65: The error should inclide the tested url.

src/sanitize-argument.ts
```diff
@@ -55,6 +56,20 @@ export async function sanitize_all(argv: any) {
...
+      }
+    })
+    .catch((err) => {
+      throw err;
```
kelson42, line 69: Why we have a ctsch() here, considering that we have one at a higher level araiund sanitizeAll()

kelson42 (changes requested):

src/sanitize-argument.ts
MananJethwani: yeah can be removed

src/sanitize-argument.ts
MananJethwani: @kelson42 redirects are allowed by default in axios

kelson42 (approved):

PR #1328, opened 2020-12-15
added methods in mediawiki and download class for url

### LINKED ISSUE #1327: Methods in Mediawiki class to give an URL to retrieve given an articleId
We should have one per possible API (which allow to retrieve article content), like:
* `veApiArticleUrl(string: articleId)`
* `restApiArticleUrl(string: articleId)`
* ...

And use them wherever this makes sense.

The goal is to simplify the code and avoid many concatenation all over the code implying a kind of URL prefix and an articleId.

MananJethwani:
@kelson42 this resolves #1327

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -232,14 +232,6 @@ class Downloader {
       await this.initLocalServices();
   }
 
-  public getBaseUrl(articleId: string): string{
-    return `${this.baseUrl}${encodeURIComponent(articleId)}`
-  }
-
-  public getBaseUrlForMainPage(articleId: string): string{
-    return `${this.baseUrlForMainPage}${encodeURIComponent(articleId)}`
-  }
-
   public async checkApiAvailabilty(url: string): Promise<boolean>{
     try {
       const resp = await axios.get(url, { maxRedirects: 0, headers: { cookie: this.loginCookie }});
@@ -255,9 +247,9 @@ class Downloader {
     // accordingly. We need to set a default page (always there because
     // installed per default) to request the REST API, otherwise it would
     // fail the check.
-    this.mwCapabilities.mobileRestApiAvailable = await this.checkApiAvailabilty(this.mw.getMobileRestApiUrl(testArticleId));
-    this.mwCapabilities.desktopRestApiAvailable = await this.checkApiAvailabilty(this.mw.getDesktopRestApiUrl(testArticleId));
-    this.mwCapabilities.veApiAvailable = await this.checkApiAvailabilty(this.mw.getVeApiUrl(testArticleId));
+    this.mwCapabilities.mobileRestApiAvailable = await this.checkApiAvailabilty(this.mw.getMobileRestApiArticleUrl(testArticleId));
+    this.mwCapabilities.desktopRestApiAvailable = await this.checkApiAvailabilty(this.mw.getDesktopRestApiArticleUrl(testArticleId));
+    this.mwCapabilities.veApiAvailable = await this.checkApiAvailabilty(this.mw.getVeApiArticleUrl(testArticleId));
 
     // Coordinate fetching
     const reqOpts = objToQueryString({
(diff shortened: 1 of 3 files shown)


kelson42:
@MananJethwani I have rebased on `master`

src/Downloader.ts
```diff
@@ -232,6 +232,14 @@ class Downloader {
       await this.initLocalServices();
   }
 
+  public getBaseUrl(articleId: string): string{
+    return `${this.baseUrl}${encodeURIComponent(articleId)}`
+  }
+
+  public getBaseUrlForMainPage(articleId: string): string{
+    return `${this.baseUrlForMainPage}${encodeURIComponent(articleId)}`
+  }
```
kelson42, line 241: I understand the logic behind this change and this is not wrong. The problem I see is that the concept of "base url" is wrong. Let me try to explain the reason.

By letting the Downloader deciding how look the URLs we create a situation where it has to decide about the proper approach beetween what is available (Downloader), how the URL looks like (Mediawiki) and how they are treated (the Renderers). It fails to do that properly because most of the constraints are outside its control. This create the kind of incoherencies where a piece of code get an URL... but ultimatively it does not know what kind of URL it is (which is then hard/impossible to deal properly with).

Long story short, I believe we should keep this like it is... even wrong... we will have to change it soon anyway and this will imply its removal.

kelson42 (changes requested):

MananJethwani:
@kelson42  is it fine now?

src/Dump.ts
```diff
@@ -203,7 +203,7 @@ export class Dump {
 
         /* Push Mediawiki:Offline.css (at the end) */
         // TODO: Weak URL (might fail in a number of cases where the wiki path is not like on Wikipedia)
-        const offlineCssUrl = `${downloader.mw.webUrl.href}?title=Mediawiki:offline.css&action=raw`;
+        const offlineCssUrl = downloader.mw.getWebArticleUrlRaw('?title=Mediawiki:offline.css&action=raw');
```
kelson42, line 206: This should only take articleId in argument.

kelson42 (changes requested):

src/Dump.ts
MananJethwani: @kelson42 but there is no other place this is being used better to rename it as getWebQueryUrl ?????

src/MediaWiki.ts
```diff
@@ -109,6 +109,30 @@ class MediaWiki {
...
+
+  public getApiQueryUrl(query: string): string {
+    return `${this.apiUrl.href}${query}`
+  }
+
+  public getWebArticleUrlRaw(article: string): string {
+    return `${this.webUrl.href}${encodeURIComponent(article)}`
+  }
+
+  public getWebQueryUrl(article: string): string {
```
kelson42, line 132: Variable shoud be articleid, not article

src/Dump.ts
```diff
@@ -203,7 +203,7 @@ export class Dump {
 
         /* Push Mediawiki:Offline.css (at the end) */
         // TODO: Weak URL (might fail in a number of cases where the wiki path is not like on Wikipedia)
-        const offlineCssUrl = `${downloader.mw.webUrl.href}?title=Mediawiki:offline.css&action=raw`;
+        const offlineCssUrl = downloader.mw.getWebQueryUrl('?title=Mediawiki:offline.css&action=raw');
```
kelson42, line 206: You should use getwebarticleraw()

src/MediaWiki.ts
kelson42: Code is wrong, ideally this method should be based on getwebqueryurl()

kelson42 (changes requested):

src/MediaWiki.ts
MananJethwani: I will remove this for now

kelson42 (approved):

src/MediaWiki.ts
```diff
@@ -109,6 +109,30 @@ class MediaWiki {
...
+    return `${this.webUrl.href}${encodeURIComponent(article)}`
```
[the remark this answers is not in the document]
MananJethwani: ok

MananJethwani:
@kelson42 is it fine now?.....or articleId should be encoded here?

PR #1334, opened 2020-12-20
Remove img parsoid attributes

kelson42:


kelson42:
@MananJethwani I don't understand why the CI fails. Works for me locally.

src/util/saveArticles.ts
```diff
@@ -789,12 +789,20 @@ export function applyOtherTreatments(parsoidDoc: DominoElement, dump: Dump) {
             node.removeAttribute('rel');
         }
 
+        /* Remove a few images Parsoid attributes */
+        else if (node.getAttribute('img') {
```
MananJethwani, line 793: @kelson42 just a missing `)` here

MananJethwani (changes requested):

PR #1336, opened 2020-12-21
created function to remove empty span

MananJethwani:
@kelson42 this removes all empty spans


kelson42:
After discussion with @MananJethwan, I prefer to keep a few epty spans than running this pretty intrusive code.

PR #1343, opened 2020-12-27
Added --webpCache flag

darkenvy:
WebP conversion is probably the longest aspect of the zim building process. My team and I often find ourselves creating multiple subsets of zims (10k, 100k, 300k, 500k) in one go. It is unfortunate that the same inclusive images have to be converted once again. When  using mwoffliner for experimental websites, on bad internet or on a poor CPU, caching provides the most benefit.

I added this flag as an advanced option for those who wish to utilize a webp cache. During times when **massive** zims are being built, a failure on a 2-5 day build is devastating. Any recovery of time helps a lot. Considering that images can never be updated on wikimedias, the "validity" of cache shouldn't expire. (An updated image creates a new URI resource. Unlike articles which occurs under the same URI). 

The cached webp images are saved by their URI, so an image queried with with 2 different sets of query parameters will result in two cached images.

**Comparison Results**
_left time: no cache. Right time: same zim parameters, with cache._
100 articles = 00:00:40 -> 00:00:13
1000 articles = 00:27:54 -> 00:08:31
10k articles = WIP

The benefit is comes into play when you daisy chain down. First make a 500k article zim, then a 300k, a 100k, a 10k, 1k all with significant stacked savings.

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -616,17 +616,18 @@ class Downloader {
     if (isBitmapImageMimeType(resp.headers['content-type'])) {
       if (isWebpCandidateImageMimeType(this.webp, resp.headers['content-type']) &&
           !this.cssDependenceUrls.hasOwnProperty(resp.config.url)) {
+        const webpCache = this.webpCache;
         let webpFilename = ''; // init as empty. Only filled out if webpCache is used. needs to be in this scope.
         let webpFullPath = '';
 
-        if (this.webpCache) {
+        if (webpCache) {
           /* a hash is generated based on the url. This ensures queries with queryParams recieve
           their own file, solves the issue of unsafe chars and collisions after cleaning said chars.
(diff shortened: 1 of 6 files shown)


darkenvy:
Im not sure how to code-coverage src/cli.ts. It has no test file. One could be created, let me know if this is what you want me to do. (The code I added in cli.ts for webpCache is copied from osTmpDir btw.)

kelson42:
@darkenvy Sorry, but we already have a s3 cache which works fin. I can not merge this. you should really have open a ticket first to avoid loosinf time :(

darkenvy:
Hey @kelson42, 
because optimisationCacheUrl flag is for s3, I did not see, in the code, that it does in fact save webp (as well as other assets). The downside is that you need a s3 emulator in order to have a local cache. Having an actual s3 is silly for our use-cases. Im going to transfer discussion to https://github.com/openzim/mwoffliner/issues/1344

and no worries, I got to acclimate to the repository and the build process :).

PR #1351, opened 2020-12-30
stopping process when an article Id fails

### LINKED ISSUE #1329: method to stop scrapping when lot of article fail from rendering needed
while scrapping mdwiki [offliner command available here](https://farm.openzim.org/recipes/mdwiki/config)
due to slow mdwiki server rendering of many articles fails.

logs are provided in file given below

[2020-12-17.txt](https://github.com/openzim/mwoffliner/files/5705509/2020-12-17.txt)

MananJethwani:
@kelson42 this is to resolve #1329

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -447,6 +447,9 @@ class Downloader {
     logger.info(`Getting article [${articleId}] from ${articleApiUrl}`);
 
     const json = await this.getJSON<any>(articleApiUrl);
+    if (json.error) {
+      throw json.error;
+    }
     return await renderArticle(json, articleId, dump, this.mwCapabilities);
   }
 
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -254,7 +254,7 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
                         dump.status.articles.success += 1;
                     }
                 } catch (err) {
-                    logger.warn(`Error downloading article ${articleId}`);
+                    logger.error(`Error downloading article ${articleId}`);
                     throw err;
                 }
 
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -254,6 +254,7 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
                         dump.status.articles.success += 1;
                     }
                 } catch (err) {
+                    dump.status.articles.fail += 1;
                     logger.error(`Error downloading article ${articleId}`);
                     throw err;
                 }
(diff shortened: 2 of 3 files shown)


MananJethwani:
@kelson42 we need to change test and I think we should keep the concept of failed articles status as we will be defining a percentage or number of fails in fututre

src/Downloader.ts
```diff
@@ -447,6 +447,9 @@ class Downloader {
     logger.info(`Getting article [${articleId}] from ${articleApiUrl}`);
 
     const json = await this.getJSON<any>(articleApiUrl);
+    if (json.error) {
```
kelson42, line 450: Why not putting this in getJSON()?

src/util/saveArticles.ts
```diff
@@ -254,9 +254,8 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
                         dump.status.articles.success += 1;
                     }
                 } catch (err) {
-                    dump.status.articles.fail += 1;
-                    logger.warn(`Error downloading article [${articleId}], skipping`, err);
-                    await articleDetailXId.delete(articleId);
+                    logger.warn(`Error downloading article ${articleId}`);
```
kelson42, line 257: This should be an error, not a warning

kelson42 (changes requested):

src/Downloader.ts
MananJethwani: getJSON is used for getting other things as well so if we put it there and we want to skip it that won't happen

src/util/saveArticles.ts
MananJethwani: ok

src/Downloader.ts
kelson42: OK, but if I understand well, the `json` variable, does not have to be JSON? What about if this is `undefined`? Which part of the code set `json.error` (is that a javascript internal? Or just a value coming from the Mediawiki API?

src/util/saveArticles.ts
```diff
@@ -254,9 +254,8 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
                         dump.status.articles.success += 1;
                     }
                 } catch (err) {
-                    dump.status.articles.fail += 1;
```
kelson42, line 257: We should probably keep that line?!

kelson42 (changes requested):
One of the thing which makes things unclear is the naming of methods I believe. What is the fundamental difference beetween `downloadContent()` and `getJSON()`? Why we don't have only one method?

kelson42 (approved):

PR #1374, opened 2021-01-18
Remove surnumerous underscore in M/Flavour

### LINKED ISSUE #1373: M/Flavour has a leading underscore
See here for example:
http://library.kiwix.org/wikivoyage_ru_all_maxi/M/Flavour

It should not be the case.

kelson42:
Fixes #1373


MananJethwani (approved):
@kelson42 checked on my local as well seems fine to me

PR #1377, opened 2021-01-20
changed base modules and using debug = true for downloading modules

MananJethwani:
@kelson42 as discussed on slack I am using debug=true as for some unknown reason mw object exist if debug is taken true but not when debug=false


kelson42:
@MananJethwani Does it fixes  #1379 ?

MananJethwani:
@kelson42 yes it does sorry forgot to link

kelson42 (approved):

PR #1385, opened 2021-01-23
fixing link problem and removed other console errors by changing base scripts

### LINKED ISSUE #1222: Tables are not expanded
**wikipedia_ru_movies_maxi_2020-06.zim**
Tables are not expanded.
Example. Article about [Quentin Tarantino](https://ru.wikipedia.org/wiki/Тарантино,_Квентин#Режиссёрский_почерк):
Quentin Tarantino-Wiki

I didn't click on the tables just because they are **very large** in the expanded state!

And this is what this part of the article in **Kiwix 2.8.0** looks like (**wikipedia_ru_movies_maxi_2020-06.zim**):
Quentin Tarantino-Kiwix

MananJethwani:
@kelson42 this resolves #1222 
reordering of files is done to remove console errors as few files depends on function present in others
`.replace('function allReady( modules ) {', 'function allReady( modules ) { return true;');` is added as startup.js tries to download them considering their ready state to be false but in our case they are present and added to html as scripts.

`hackMediaWikiModule()` is changed to remove `mw.hook` error which stopped `collapsibleTable()` function from being declared (after this `mw` object is completely available for other modules)

`hackSiteModule` is added to make some changes to site module to make execution of collapsible function possible as some modules are not downloaded, I will be opening a separate ticket for that.


kelson42 (changes requested):
We need in the code an extend comment explaining what and why we do thisnregex replacement. If nothing is replaced then this means something goes wrong, maybe we should then just die?

kelson42 (changes requested):
We shoudl not have to change `colllapsibleTable()`, I see no valid reason to have to do so. If we have to do this is probably because there is a bug left somewhere else.

MananJethwani:
the last commit will solve the issue of #1398 but there might be still room for improvement

MananJethwani:
@kelson42 at this moment I have added `'mediawiki.util', 'mediawiki.page.ready'` to the base module list but, it should have been fetched from the modules list and as the module list is not complete I have added it here for now.
please let me know what should be done.

src/util/dump.ts
```diff
@@ -90,24 +91,16 @@ export async function downloadAndSaveModule(zimCreator: ZimCreator, mw: MediaWik
...
+    // will require the functions which would have been loaded by doPropagation.
     function hackStartUpModule(jsCode: string) {
-        return jsCode.replace(
-            'script=document.createElement(\'script\');',
-            `
-                    document.body.addEventListener('fireStartUp', function () { startUp() }, false);
-                    return;
-                    script=document.createElement('script');`,
-        );
-    }
```
kelson42, line 104: Why we remove this? It is not needed anymore?

src/util/dump.ts
```diff
@@ -90,24 +91,16 @@ export async function downloadAndSaveModule(zimCreator: ZimCreator, mw: MediaWik
...
-            const startUpEvent = new CustomEvent('fireStartUp');
-            document.body.dispatchEvent(startUpEvent);
-        })()`;
-        return jsCode;
+        if (!STARTUP_MODULE_REGEX.test(jsCode)) {
+            throw new Error('unable to hack startup module');
+        }
+
+        return jsCode.replace('mw.requestIdleCallback( doPropagation, { timeout: 1 } );', 'doPropagation();')
+            .replace('function allReady( modules ) {', 'function allReady( modules ) { return true;');
```
kelson42, line 103: This is a bit verbose and fragile.  Why we have `if (!STARTUP_MODULE_REGEX.test(jsCode))` in a first place?

src/util/dump.ts
MananJethwani: no it isn't there is actually no function `startup()` present anywhere in resource modules.

src/util/dump.ts
MananJethwani: @kelson42 you asked me to add a check so that if anything in the startup script changes and it stops working we can know about it immediately.

kelson42: OK but this is still super fragile. To test if it "it stops working" you need to secure each the changes your apply really apply.

kelson42 (approved):

PR #1400, opened 2021-02-11
Update old version of gcc/g++

kelson42:


MananJethwani:
@kelson42 the CDN is still down I have opened an issues on cnpmjs.org's repo here ->  cnpm/cnpmjs.org#1622
let's see if they can help with that in anyway

PR #1407, opened 2021-03-05
corrected paths for articles starting with /

### LINKED ISSUE #1368: Unable to view article names with forward slashes on a wikipedia .zim file
*@mechalincoln commented on Dec 19, 2020, 1:11 AM UTC:*

Hello. I think this issue affects multiple zim files and/or kiwix-serve, but any article name with a forward slash is either not included or the server can't parse article names with a slash.

In the following example, I navigate through my local version of the November 2020 Wikipedia maxi zim file's Device File article. The links that I click on that have forward slashes go to a Not Found page. In the same animation, I also navigate to [www.wikipedia.org](http://www.wikipedia.org) for their [Device File](https://en.wikipedia.org/wiki/Device_file) article, where the links with forward slashes properly navigate.

I'm hoping this is something that can be looked at.

[devicefile](https://user-images.githubusercontent.com/49933679/102676772-c013f300-4164-11eb-8c9f-5351b336d5a4.gif)

The URLs may just not be escaped correctly, as this works on my local copy: [http://192.168.1.97/wikipedia_en_all_maxi_2020-11/A//dev/zero](http://192.168.1.97/wikipedia_en_all_maxi_2020-11/A//dev/zero)

Kiwix is one of my favorite programs. Thank you again for maintaining it.

*This issue was moved by [kelson42](https://github.com/kelson42) from [kiwix/kiwix-tools#425](https://github.com/kiwix/kiwix-tools/issues/425).*

MananJethwani:
@kelson42 this resolves #1368

### DIFF
--- a/src/util/articleListMainPage.ts
+++ b/src/util/articleListMainPage.ts
@@ -1,7 +1,7 @@
 import { Dump } from '../Dump';
 import { encodeArticleIdForZimHtmlUrl, isWebpCandidateImageUrl } from '../util';
 
-export function makeArticleListItem(dump: Dump, articleEntry: any) {
+export function makeArticleListItem(dump: Dump, articleEntry: ArticleDetail) {
     return `<li><a href="${encodeArticleIdForZimHtmlUrl(articleEntry.title.replace(/ /g, '_'))}">${articleEntry.title.replace(/_/g, ' ')}<a></li>\n`;
 }
 
--- a/src/util/articleListMainPage.ts
+++ b/src/util/articleListMainPage.ts
@@ -2,11 +2,11 @@ import { Dump } from '../Dump';
 import { encodeArticleIdForZimHtmlUrl, isWebpCandidateImageUrl } from '../util';
 
 export function makeArticleListItem(dump: Dump, articleEntry: any) {
-    return `<li><a href="${encodeArticleIdForZimHtmlUrl('./' + articleEntry.title.replace(/ /g, '_'))}">${articleEntry.title.replace(/_/g, ' ')}<a></li>\n`;
+    return `<li><a href="${encodeArticleIdForZimHtmlUrl(articleEntry.title.replace(/ /g, '_'))}">${articleEntry.title.replace(/_/g, ' ')}<a></li>\n`;
 }
 
 export function makeArticleImageTile(dump: Dump, articleEntry: ArticleDetail, webp: boolean) {
-    return `<a class="item" href="${encodeArticleIdForZimHtmlUrl('./' + articleEntry.title.replace(/ /g, '_'))}">
+    return `<a class="item" href="${encodeArticleIdForZimHtmlUrl(articleEntry.title.replace(/ /g, '_'))}">
             <figure><img loading="lazy" src="${articleEntry.internalThumbnailUrl +
             (webp && isWebpCandidateImageUrl(articleEntry.internalThumbnailUrl) ? '.webp' : '')}" />
             <figcaption>${(articleEntry.title || '').replace(/_/g, ' ')}</figcaption></figure></a>\n`;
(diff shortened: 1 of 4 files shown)


src/util/misc.ts
```diff
@@ -316,7 +316,7 @@ export function sanitizeString(str: string) {
 // We will need the encoded URL on article load so that we can set the hrefs of anchor tag correctly,
 // but we must not encode the '/' character or else relative links may fail
 export function encodeArticleIdForZimHtmlUrl(articleId: string) {
-  return articleId && encodeURIComponent(articleId).replace(/%2F/g, '/');
+  return articleId && encodeURIComponent(articleId[0] === '/' ? `./${articleId}` : articleId).replace(/%2F/g, '/');
```
kelson42, line 319: If `articleId` is empty string it will fail probably. This test should better be done with `startsWith()` or using a regex.

kelson42 (changes requested):

src/util/articleListMainPage.ts
```diff
@@ -6,7 +6,7 @@ export function makeArticleListItem(dump: Dump, articleEntry: any) {
 }
 
 export function makeArticleImageTile(dump: Dump, articleEntry: ArticleDetail, webp: boolean) {
-    return `<a class="item" href="${'./' + encodeArticleIdForZimHtmlUrl(articleEntry.title.replace(/ /g, '_') || '')}">
+    return `<a class="item" href="${encodeArticleIdForZimHtmlUrl(articleEntry ? './' + articleEntry.title.replace(/ /g, '_') : './')}">
```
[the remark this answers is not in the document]
MananJethwani: in case the `articleEntry.title` starts with a `/` it will try to replace it with a `./` and we are already adding a `./` to it after calling `encodeArticleIdForZimHtmlUrl()` one thing that I did wrong is use `articleEntry?` in place of `articleEntry.title` sorry about that.....but other than that even if `articleentry.title`was NULL we are actually having `href` as `./` that's why I kept `./` even in case `articleEntry.title` is null

src/util/misc.ts
MananJethwani: right, will change that

src/util/articleListMainPage.ts
```diff
@@ -6,7 +6,7 @@ export function makeArticleListItem(dump: Dump, articleEntry: any) {
 }
 
 export function makeArticleImageTile(dump: Dump, articleEntry: ArticleDetail, webp: boolean) {
-    return `<a class="item" href="${'./' + encodeArticleIdForZimHtmlUrl(articleEntry.title.replace(/ /g, '_') || '')}">
+    return `<a class="item" href="${encodeArticleIdForZimHtmlUrl('./' + articleEntry.title.replace(/ /g, '_'))}">
```
MananJethwani, line 9: @kelson42 the reason for changing this line is that if `articleEntry.title` starts with a `/` it will convert it to start with `./` and then as we were already adding one `./` to it after `encodeArticleIdForZimHtmlUrl()` was called we would end up having `././`. I guess that won't cause any problem but, still would have been wrong and in case `title` was empty we would just have `./` as the href, so what I have done now does the same thing and just removes the problem of two consecutive  `./` in href

src/util/articleListMainPage.ts
```diff
@@ -2,11 +2,11 @@ import { Dump } from '../Dump';
 import { encodeArticleIdForZimHtmlUrl, isWebpCandidateImageUrl } from '../util';
 
 export function makeArticleListItem(dump: Dump, articleEntry: any) {
```
MananJethwani, line 4: @kelson42 shouldn't here type of `articleEntry` should be `articleDetail`

kelson42 (approved):

PR #1413, opened 2021-03-20
removed console.log

MananJethwani:


MananJethwani:
@kelson42 I am changing and just removing that console log I used for debugging we can open another issue for base module reordering.....console.log regression needs to be fixed

PR #1415, opened 2021-03-25
Add installation on RHEL 8 section

erbap:


kelson42 (changes requested):
I'm not in favour to explain for each GNU/Linux distribution how to install the dependences... in particular following a manual path (using packages should be better).

What in your PR might be interested is the part about `unsage` but using `--unsafe` seems to me to be a bad workaround. What is the legitimate use case?

erbap:
Strange that the installation for the dependencies for Debian and macOS are there but for RHEL based systems it's not relevant?

When there was a package for the libjpeg I would have included it but mwoffliner need a so old package that the only way to get it is by manually compiling/installing it. I had to search a while before I found a compatible packages.

I also have the feeling that for mwoffliner there is not really a lot of documentation.

For the --unsafe thing, I'm not a real nodejs developer but after beeing confronted with the error I came to https://github.com/npm/npm/issues/1259

kelson42:
> Strange that the installation for the dependencies for Debian and macOS are there but for RHEL based systems it's not relevant?

The point is that we have people here to maintain the doc for Ubuntu and macOS. Nobody in our team does use RHEL and you are right in the point that if we accept RHEL, why not then Arch... or anything else. I will remove the hints about macOS and Ubuntu. This is anyway pretty trivial and not our job anyway to document that.

> When there was a package for the libjpeg I would have included it but mwoffliner need a so old package that the only way to get it is by manually compiling/installing it. I had to search a while before I found a compatible packages.

This is 100% new to me. I'm not informed we require an old version. If this is confirmed this is a problem. Please open a dedicated ticket for that.

> I also have the feeling that for mwoffliner there is not really a lot of documentation.

Please open a ticket pointing to what is exactly missing. This will help to improve it.

> For the --unsafe thing, I'm not a real nodejs developer but after beeing confronted with the error I came to [npm/npm#1259](https://github.com/npm/npm/issues/1259)

I'm exactly a the same level like you and this is the problem and I don't want to recommend them to use an `unsafe` option without 200% understanding why they should do so and what it exactly means.

I have created a new PR taking in account the points I believe we could be a bit better. It's here https://github.com/openzim/mwoffliner/pull/1422. For rest, waiting to your dedicated tickets to go forward.

PR #1420, opened 2021-03-26
handled score extension images

### LINKED ISSUE #1418: Score Extension images are not handled properly
Notes are handled properly at:
https://af.wikiquote.org/api/rest_v1/page/html/Johann_Sebastian_Bach

but not at:
http://library.kiwix.org/wikiquote_af_all_maxi/A/Johann_Sebastian_Bach

MananJethwani:
@kelson42 this resolves #1418

### DIFF
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -235,7 +235,7 @@ export function getMediaBase(url: string, escape: boolean) {
 
   // Image thumbs
   if ((parts = IMAGE_THUMB_URL_REGEX.exec(decodedUrl)) !== null) {
-      filename = parts[1].length > parts[3].length ? parts[1] : parts[3];
+      filename = parts[3];
   }
 
   // Latex (equations) & Graphoid
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -235,22 +235,22 @@ export function getMediaBase(url: string, escape: boolean) {
 
   // Image thumbs
   if ((parts = IMAGE_THUMB_URL_REGEX.exec(decodedUrl)) !== null) {
-      filename = parts[3];
+    filename = parts[3];
   }
 
   // Latex (equations) & Graphoid
   else if ((parts = LATEX_GRAPHOID_IMAGE_URL_REGEX.exec(decodedUrl)) !== null) {
-      filename = parts[1] + '.svg';
+    filename = parts[1] + '.svg';
   }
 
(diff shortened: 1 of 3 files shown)


MananJethwani:
@kelson42 sorry I forgot about test please don't review yet

MananJethwani:
@kelson42 done.....I am using only `.png` in regex as it is mentioned [here](https://www.mediawiki.org/wiki/Extension:Score#:~:text=The%20Score%20extension%20allows%20the,into%20audio%20and%20MIDI%20files.) that this extension generates only png images

kelson42:
@MananJethwani Code factor does not pass

MananJethwani:
@kelson42 please have a look now

src/util/misc.ts
```diff
@@ -233,8 +233,15 @@ export function getMediaBase(url: string, escape: boolean) {
   let parts;
   let filename;
 
+  // Score images(music notes)
+  parts = SCORE_IMAGE_URL_REGEX.exec(decodedUrl)
```
kelson42, line 237: Missing semi colon

src/util/misc.ts
```diff
@@ -233,8 +233,15 @@ export function getMediaBase(url: string, escape: boolean) {
...
+  if(parts !== null){
```
kelson42, line 238: Missing space

kelson42 (changes requested):
The problem of this PR is that score images are 0,0001% of all images, but mwoffliner checks them first, big waste of CPU.

MananJethwani:
> The problem of this PR is that score images are 0,0001% of all images, but mwoffliner checks them first, big waste of CPU.

I would try to somehow merge the first 2 conditions to make this a bit faster I might have an idea....give me some time

kelson42:
You should make the thumb regex more specific and put your score regex after

MananJethwani:
@kelson42 instead of moving it below thumb regex I could do something like this also
```
const len = 'https://upload.wikimedia.org/'.length;
str.indexOf('score/', len) === len;
```
or change the pattern only using
```
const pattern = str.includes('/score/') ? ... : ...;
```
if these are slow according to you I will go with your solution only of moving score regex down as conditional regex don't work in js

kelson42:
No, do a proper regex please and don't test only the keyword "score".

kelson42 (approved):

PR #1423, opened 2021-03-27
continuing scrapping even if article doesn't render

### LINKED ISSUE #1419: Unable to scrape wiktionary CHR any more
https://farm.openzim.org/recipes/wiktionary_chr

Seems to be a 1.11.4 regression

MananJethwani:
this resolves #1419
@kelson42 these articles give 200 response so we needed to check if they have any content or not if they don't have any I am returning a new `RENDERING_ERROR` .


kelson42:
@MananJethwani would you be able to share here the url which generates the rendering error please?

MananJethwani:
@kelson42 here it is -> https://chr.wiktionary.org/w/api.php?action=visualeditor&mobileformat=html&format=json&paction=parse&page=likaledes
the content is not there so it couldn't be rendered

MananJethwani:
@kelson42 please review this I have rebased it.

kelson42 (changes requested):
The problem with this PR is:
* I don't have a clear/simple reproducation case
* The given explanation (in the ticket) seems wrong because it always dies, so this is not a sporadic problem. I would like to understand the problem
* I don't understand the solution... probably because I don't understand what is the problem
* There is no automated test

kelson42 (changes requested):
This case seems to happen when an article is deleted in the meantime between ns article list retrieval and content retrieval. This is a valid scenario. We could check the `oldid` to verify this case and skip (with a log message in such a case). Automated tests have to be adatped.

MananJethwani:
@kelson42 are these checks fine?

test/unit/saveArticles.test.ts
```diff
@@ -261,3 +260,16 @@ test('treat multiple subtitles in one video', async(t) => {
         'Video multiple subtitles rewriting matches');
     t.equals(contentRes.subtitles.length, 4, 'All subtitles are found for this video');
 });
+
+test('Desktop article Rendering Test' , async(t) => {
```
kelson42, line 264: Name of the test is wrong

test/unit/saveArticles.test.ts
```diff
@@ -261,3 +260,16 @@ test('treat multiple subtitles in one video', async(t) => {
...
+    const DELETED_ARTICLE_FAIL_MESSAGE = 'Throwing error if article is deleted';
```
kelson42, line 265: You should no redeclare the error message (which btw is not the same as in the code). Yo shoukd test that the exception is gneraged properly.

kelson42 (changes requested):

test/unit/saveArticles.test.ts
MananJethwani: @kelson42 this is not the message used for error, it is the message shown in the terminal when the integration test fails or passes, I had to put it in 2 places one if `renderDesktopArticle()` didn't throw any error it should fail, and once in the catch block if the error is different.

kelson42: Ok, but you should just success if proper exception... without having to create a fake exception.

kelson42 (changes requested):
Implement the test is a better way to avoid creating a fake exception.

MananJethwani:
@kelson42 is this fine now?

kelson42 (approved):

PR #1427, opened 2021-04-03
corrected page title retrieval in normalizeMwResponse

### LINKED ISSUE #1426: fetching article fails with 400 status code
have a look at https://farm.openzim.org/pipeline/c84d0fa92a4c21b9d25df506/debug
will investigate into it and let everyone know soon

MananJethwani:
@kelson42 this resolves #1426

### DIFF
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -114,7 +114,7 @@ export function normalizeMwResponse(response: MwApiQueryResponse): QueryMwRet {
 
     return Object.values(pages)
         .reduce((acc, page) => {
-            const id = (normalized[page.title] || page.title || '');
+            const id = ((normalized.hasOwnProperty(page.title) && normalized[page.title]) || page.title || '');
             if (typeof id !== 'string' || !id) {
                 logger.warn(`Article Id is invalid - expected a string but got [${id}], converting to string and continuing`);
             }
--- a/test/unit/util.test.ts
+++ b/test/unit/util.test.ts
@@ -6,9 +6,8 @@ import pathParser from 'path';
 import { sanitize_customFlavour } from 'src/sanitize-argument';
 import { encodeArticleIdForZimHtmlUrl, interpolateTranslationString, getFullUrl } from 'src/util';
 import { testHtmlRewritingE2e } from 'test/util';
-import { getMediaBase } from '../../src/util/'
-import logger from '../../src/Logger';
-import { waitForDebugger } from 'inspector';
+import { getMediaBase, normalizeMwResponse } from '../../src/util/';
+import axios from 'axios';
 
 test('util -> interpolateTranslationString', async (t) => {
     t.equals(interpolateTranslationString('Hello world', {}), 'Hello world');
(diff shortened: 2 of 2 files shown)


kelson42 (changes requested):
Please update a test with this edge-case.

MananJethwani:
@kelson42 I am a bit confused about the approach with the test case, will this be fine with you?

src/util/mw-api.ts
```diff
@@ -114,7 +114,7 @@ export function normalizeMwResponse(response: MwApiQueryResponse): QueryMwRet {
 
     return Object.values(pages)
         .reduce((acc, page) => {
-            const id = (normalized[page.title] || page.title || '');
+            const id = ((normalized.hasOwnProperty(page.title) && normalized[page.title]) || page.title || '');
```
kelson42, line 117: I don't understand exactly why this is needed and how this fixes the problem.

test/unit/util.test.ts
```diff
@@ -128,4 +127,10 @@ test('getMediaBase tests', async(t) => {
 
     // Default behaviour
     t.equal(getMediaBase('https://maps.wikimedia.org/img/osm-intl,9,52.2789,8.0431,300x300.png?lang=ar&amp;domain=ar.wikipedia.org&amp;title=%D8%A3%D9%88%D8%B3%D9%86%D8%A7%D8%A8%D8%B1%D9%88%D9%83&amp;groups=_0a30d0118ec7c477895dffb596ad2b875958c8fe', true), '589fd4e3821c15d4fcebcedf2effd5b0.png', 'Default handling');
+})
+
+test('normalizeMwResponse edge case test', async(t) => {
+   const resp = await axios.get<MwApiResponse>('https://en.wiktionary.org/w/api.php?action=query&format=json&prop=redirects|revisions|pageimages&rdlimit=max&rdnamespace=0&redirects=true&titles=constructor', { responseType: 'json', });
+   const normalizedObject = normalizeMwResponse(resp.data.query);
+   t.equal(Object.keys(normalizedObject)[0], 'constructor', 'normalizeMwResponse returns title constructor');
```
kelson42, line 135: I would have just added "constructor" in the list of articles to scrape from WPEN in `extra.e2e.test.ts`... But, actually, I have done that with master code and the test passes.. so I believe here I don't understand either the bug nor the solution. So what is the root caue of the bug?

kelson42 (changes requested):

src/util/mw-api.ts
MananJethwani: @kelson42 `normalize` doesn't have any element named constructor so by default all arrays are also special kind of objects inJS so when we write `normalized[page.title]` it tries to access the default constructor inherited from object class and assign that to `id`, using `normalized.hasOwnProperty(page.title)` checks if it is it's own property and not inherited property.

kelson42: Why `normalized` does not have an element at the key `constructor`? It should, this is how the code works! Why we have a problem now considering that this code has not been changed?

MananJethwani: @kelson42 `normalized` is not present in case of Wiktionary as you can see [here](https://en.wiktionary.org/w/api.php?action=query&format=json&prop=redirects%7Crevisions%7Cpageimages&rdlimit=max&rdnamespace=0&redirects=true&titles=constructor)

test/unit/util.test.ts
MananJethwani: in case of Wikipedia normalized is present so this can't be tested with it

MananJethwani:
@kelson42 also shouldn't this line https://github.com/openzim/mwoffliner/blob/d98d04b46abab217831f72b3ae5e53777a988040/src/util/mw-api.ts#L111

be `acc[from] = to` as we find the `from` version and need to replace it with the `to` one?

src/util/mw-api.ts
kelson42: ok, lets try this.

test/unit/util.test.ts
kelson42: OK, please just rename your test properly, something like "no title normalisation"

kelson42 (changes requested):

kelson42 (approved):

PR #1434, opened 2021-04-05
changed method of url rewriting

### LINKED ISSUE #904: URL rewritting is not economical (and probably not robust/fast)
While relative liniks are perfectly correct, they are not quite as economical as they could be. Relative links on the landing of WVEN page look of like this:

`<a title="Europe" href="../../A/Europe">Europe</a>` 

It would be sufficient for them to be coded like this:

`<a title="Europe" href="../Europe">Europe</a>`

In other words, the relative links take us two levels up and then back down into the current namespace, which isn't necessary.

I thinks this means that we have custom relative path computation. I wonder why we don't use https://millermedeiros.github.io/mdoc/examples/node_api/doc/path.html#path.relative. This would offer many advantages:
* Avoid the many bugs we have already detected - and still have open - around this feature
* Avoid many unit tests around that
* A solution for sure a lot faster than what we have (and this routine is used a lot)

MananJethwani:
@kelson42 this resolves #904 but currently should be drafted due to #1416 or is this a bug in kiwix-serve?

### DIFF
--- a/src/util/rewriteUrls.ts
+++ b/src/util/rewriteUrls.ts
@@ -6,6 +6,7 @@ import { Dump } from '../Dump';
 import MediaWiki from '../MediaWiki';
 import DU from '../DOMUtils';
 import logger from '../Logger';
+import path from 'path';
 
 export async function removeOrEncodeLink(mw: MediaWiki, dump: Dump, linkNode: DominoElement, href: string) {
     const title = mw.extractPageTitleFromHref(href);
@@ -175,9 +176,7 @@ export async function rewriteUrl(articleId: string, mw: MediaWiki, dump: Dump, l
         if (articleId.includes('/')) {
             const href = linkNode.getAttribute('href'); // href is modified above, so this is necessary
             const resourceNamespace = 'A';
-            const slashesInUrl = articleId.split('/').length - 1;
-            const upStr = '../'.repeat(slashesInUrl + 1);
-            linkNode.setAttribute('href', `${upStr}${resourceNamespace}/${href}`);
+            linkNode.setAttribute('href', path.relative(`${resourceNamespace}/${articleId}`, `${resourceNamespace}/${href}`));
         }
     }
     return { mediaDependencies };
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
@MananJethwani What do you think of last commit I made? It seems to fix the problem properly?!

MananJethwani:
@kelson42 yes I approve that this will be the right thing to do but I still think if it is covered in `kiwix-serve` would be better because usually that is how it works for all the servers in genereal doesn't it?

kelson42:
@MananJethwani Can you explain precisely waht is wrong with kiwix-serve? Because I don't get it.

MananJethwani:
in #1416 
this was mentioned by @veloman-yunkan
> This issue may be caused by the presence of a slash in the article name. `/dev/zero` is considered as a single article name rather than a path consisting of multiple components. Such a problem [has been observed before](https://github.com/openzim/zim-tools/issues/151#issuecomment-678352432), but we haven't designed an approach to deal with it.

kelson42:
@MananJethwani I asked for something precise wrong in kiwix-serve. This last comment does not deliver this info.

MananJethwani:
the thing exactly wrong IMO is in a URL for article with multiple `/`, slashes are not considered different parts of URL but they should be considered different as that's the usual behavior we see.

kelson42:
@MananJethwani and how is that a bug with kiwix-serve? The path resilution happens in the browser!

MananJethwani:
@kelson42 when you moved the issue to mwoffliner I considered that as well it's veloman's reply in the ticket that confuses me that is that a fault with the browser or `kiwix-serve`

MananJethwani:
@kelson42 while writing a test for links with double slashes I encountered an issue
I did something like this -> 
```
const $wikiLinkWithDoubleSlashInBetween = makeLink($doc, '/wiki/Blood_//_Water', '', 'Blood // Water');
await rewriteUrl(complexParentArticleId, mw, dump, $wikiLinkWithDoubleSlashInBetween);
logger.log($wikiLinkWithDoubleSlashInBetween.getAttribute('href'));
```

I got this ` ../../../../../../../wiki/Blood_/_Water` because of the last line is it because I was using `complexParentArticleId` but then also it should have worked correctly buy when I scrape `blood // water` and  a page with link to this page it works fine I confirm that.

Issue #1476, opened 2021-06-08
MWoffliner fails if --articleList files has "Windows" carriage returns


kelson42:
https://farm.openzim.org/pipeline/c0dac3ef7c1bed11dc25fb06/debug

PR #1478, opened 2021-06-08
Better indgest Windows encoded article lists

kelson42:
Fixing #1476


MananJethwani (approved):
@kelson42 shouldn't this be fixed in zimfarm as well?
they should remove windows carriage return as mwoffliner runs only on linux (AFAIK).

Issue #1464, opened 2021-05-16
wikipedia_en_physics Does not scrape anymore because of “URI malformed”


kelson42:
https://farm.openzim.org/pipeline/09921832a7608c8ca14df906/debug

kelson42:
Other case at https://farm.openzim.org/pipeline/29fcf8dc6f111c1fee487a06/debug

kelson42:
Same with https://farm.openzim.org/recipes/wikipedia_hr_all

kelson42:
Fixed by https://github.com/openzim/mwoffliner/pull/1483

kelson42:
Unfortunately the bug is not (fully) fixed, see https://farm.openzim.org/pipeline/c53c4eba6e0c8c0b7e792116/debug

PR #1483, opened 2021-06-10
Catch url.parse call to avoid crashing

kelson42:
Fixing #1464


kelson42:
Will merge as we have really many errors of that type and need to test this PR.

PR #1488, opened 2021-06-29
Better WebP polyfill

rgaudin:
The Webp Hero polyfill has a major drawback: it can only convert a WebP at once.
It seems to be a limitation of the emscripten-based libwebp but Webp Hero doesn't do
much about it except throwing an exception.
Given the `polyfillDocument()` method iterates over all `` of the document
to request convertion, it can be very frequent that images are not decoded and thus
just skipped.

The actual behavior is unpredictable as it depends a lot on how quickly the DOM is parsed
by the browser. I believe using lazy loading really helped in the case of mwoffliner
as images are frequently scattered around the page and thus not loaded alltogether.

This replaces the use of the WebPMachine from Webp Hero by a custom one that decodes
one by one (using Webp Hero's code) using a pending list.
It's quite naive but seems to do the job.

This Handler's code currently reside in a gist and is thus downloaded from there.
I've noticed there is no other such dependency in mwoffliner, so if you think it should
be published as a regular npm lib, we can go down this extra road. Let me know

### DIFF
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -11,7 +11,7 @@ import { ZimCreator, ZimArticle } from '@openzim/libzim';
 import { Dump } from '../Dump';
 import { filesToDownloadXPath } from '../stores';
 import fs from 'fs'
-import { DO_PROPAGATION, ALL_READY_FUNCTION } from './const';
+import { DO_PROPAGATION, ALL_READY_FUNCTION, WEBP_HANDLER_URL } from './const';
 
 export async function getAndProcessStylesheets(downloader: Downloader, links: Array<string | DominoElement>) {
     let finalCss = '';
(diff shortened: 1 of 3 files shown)


src/util/dump.ts
```diff
@@ -147,4 +148,18 @@ export async function importPolyfillModules(zimCreator: ZimCreator) {
         });
         zimCreator.addArticle(article);
     });
-}
\ No newline at end of file
+
+    const webpHandlerUrl = 'https://gist.githubusercontent.com/rgaudin/60bb9cc6f187add506584258028b8ee1/raw/9d575b8e25d67eed2a9c9a91d3e053a0062d2fc7/web-handler.js';
```
MananJethwani, line 152: please move this to src/util/const.ts

src/util/dump.ts
```diff
@@ -147,4 +148,18 @@ export async function importPolyfillModules(zimCreator: ZimCreator) {
...
+    const content = await axios.get(webpHandlerUrl, {headers: {}, responseType: 'arraybuffer', timeout: 60000, method: 'GET', validateStatus(status) { return (status >= 200 && status < 300) || status === 304; }})
```
MananJethwani, line 153: is `method: 'GET'` needed here? we are already using `axios.get()` here

MananJethwani (changes requested):
2 minor changes other than that LGTM

src/util/dump.ts
rgaudin: Right ; simplified

rgaudin:
Thank @MananJethwani ; fixed those two issues. What does LGTM mean? 🤷‍♂️.

Jaifroid:
Just to say we had to do the same thing, more or less, with Kiwix JS. See [Extract images sequentially in jQuery mode](https://github.com/kiwix/kiwix-js/pull/677).

MananJethwani:
> What does LGTM mean?

@rgaudin it stands for Looks Good To Me 😅

rgaudin:
yeah sorry I had my mind focused on it being JS-lingo and and didn't catch it 😅

PR #1490, opened 2021-07-04
Fix redirect retrieval if > 500 redirects

MananJethwani:
@kelson42 this fixes tests and a minor bug I found from the last merged PR looks like rgaudin forgot to change variable name from `webpHandlerUrl` to `WEBP_HANDLER_URL` and we merged it.


kelson42:
@MananJethwani this is wrong, there is actually more than 300 redirects... younshoukd not chamge test values like this without verifying things.

kelson42 (approved):

PR #1507, opened 2021-07-20
fixed articleId encoding

### LINKED ISSUE #1468: Request failed with status code 400
For https://en.wikipedia.org/api/rest_v1/page/mobile-sections/I%25CC%2587znik

see https://farm.openzim.org/pipeline/7adfa713c4bff5e9ce378a06/debug

MananJethwani:
this fixes #1468

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -509,7 +509,7 @@ class Downloader {
   }
 
   private getArticleUrl(articleId: string, isMainPage: boolean): string {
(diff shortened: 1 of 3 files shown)


kelson42:
I don't think this is the right way. We should decodeUrl the value before saving it in the redis!

kelson42:
We also need to extend one of our test to secure this works properly.

kelson42 (approved):

PR #1508, opened 2021-07-21
increased timeout from 60 sec to 120 sec

MananJethwani:


kelson42 (changes requested):
Update the usage() please.

kelson42 (approved):

PR #1515, opened 2021-08-05
changed video source for better resolution

### LINKED ISSUE #1489: Insufficient resolution of some videos within a ZIM
In this ZIM http://iiab.me/downloads/wp_en_mdwiki_2021-06.zim

For the article on gout the first video has a great resolution while the second has a poor resolution

This is the first video https://commons.wikimedia.org/wiki/File:En.Wikipedia-VideoWiki-Gout.webm

This is the second https://commons.wikimedia.org/wiki/File:Gout.webm

And this is the page it is built from https://mdwiki.org/wiki/Gout

We are needing a method to set the minimum resolution of a video instead of it just choosing the lowest possible resolution per https://github.com/openzim/mwoffliner/issues/1442.

The poor resolution issue for Osmosis videos persists even when these videos are in the same spot as the good quality non Osmosis videos. See amnesia as an example.

MananJethwani:
this fixes #1489

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -383,7 +383,27 @@ function widthXHeightSorter(a: DominoElement, b: DominoElement) {
 
     const aVal = aWidth * aHeight;
     const bVal = bWidth * bHeight;
-    return aVal > bVal ? 1 : -1;
+    return aVal > bVal ? 1 :
+        ((aVal === bVal && (b.getAttribute('src').endsWith('.vp9.webm') ||
+        (b.getAttribute('src').endsWith('webm') && !a.getAttribute('src').endsWith('.vp9.webm')))) ? 1 : -1);
+}
+
+function videoSourceElement(videoSources: DominoElement, videoElWidth: Number) {
+    let sourceEl: DominoElement;
+    let sourcesToRemove = videoSources.filter((videoSource: DominoElement) => {
+        const sourceWidth = Number(videoSource.getAttribute('data-file-width') || videoSource.getAttribute('data-width') || 0);
+        if (sourceWidth >= videoElWidth && !sourceEl) {
+            sourceEl = videoSource;
+            return false;
+        }
+        return true;
+    });
+
+    if(!sourceEl) {
+        sourceEl = sourcesToRemove[0];
+        sourcesToRemove = sourcesToRemove.slice(1);
+    }
+    return {sourceEl, sourcesToRemove};
 }
 
 export async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolean>, articleId: string, videoEl: DominoElement, webp: boolean): Promise<{ mediaDependencies: string[], subtitles: string[] }> {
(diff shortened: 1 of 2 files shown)


src/util/saveArticles.ts
```diff
@@ -412,13 +412,26 @@ export async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolea
     }
 
     videoSources = videoSources.sort(widthXHeightSorter);
+    const videoElArea = (videoEl.getAttribute('height') || 0) * (videoEl.getAttribute('width') || 0);
```
kelson42, line 415: Here we need a comment explaining a bit the problem and the solution

kelson42, line 415: I don't understand really the computation with width*heigth. Considering that all the videos have the same ratio (width/height), I think just comparing the widths should be enough.

src/util/saveArticles.ts
```diff
@@ -412,13 +412,26 @@ export async function treatVideo(mw: MediaWiki, dump: Dump, srcCache: KVS<boolea
...
+        return true;
+    });
 
-    const sourcesToRemove = videoSources.slice(1); // All but first
+    if(!sourceEl) {
+        sourceEl = sourcesToRemove[0];
+        sourcesToRemove = sourcesToRemove.slice(1);
+    }
 
     sourcesToRemove.forEach(DU.deleteNode);
```
kelson42, line 433: This is simply wrong, `sourcesToRemove` is not always defined.

kelson42 (changes requested):
What is the difference beetween `data-width` and `data-file-width`, it does not looks good to take both of them?

This should be tested with an automated test.

kelson42 (changes requested):
I think the whole approach can be simplified a lot:
* Remove `widthXHeightSorter` because we don't need that anymore as we don't take the small resolution anymore
* Just go through all the `source` node element and either keep the node (because this is a better choice than the previous) or remove it. At the end it should be only one left.
* I'm not even sure you need a dedicated function, so probably `videoSourceElement` could be deleted as well. If you want to isolate things in a dedicated function just create a function `videoSourceUrlPicker()` (current name `videoSourceElement` is pretty cryptic as well).

You should be able to do all of that in twice less lines than what we have today.... at least.

kelson42:
@tim-moody Does it do the job properly now?

tim-moody:
@WikiDocJames and I will try to test.

kelson42 (approved):
LGTM now

kelson42:
Will merge this PR, if a problem is still occuring, please open a new ticket.

WikiDocJames:
Okay we tested it 16 hours ago. But looks like you have just pushed the fix 3 hour ago. So will test again this am...

tim-moody:
@kelson42 I didn't notice that the changes weren't merged. Am I now to test with the latest docker image or a nightly or by building a branch?

tim-moody:
docker pull openzim/mwoffliner:latest says Image is up to date for openzim/mwoffliner:latest

tim-moody:
@kelson42 when do we expect this fix to be in the docker image?

PR #1522, opened 2021-08-11
audio tag fix

### LINKED ISSUE #1489: Insufficient resolution of some videos within a ZIM

MananJethwani:
this fixes #1489


kelson42 (approved):
LGTM, thx.

PR #1550, opened 2021-10-15
corrected tree structure for sections

### LINKED ISSUE #1501: Problem about tree structure of a page
In the page "Perturbation_theory_(quantum_mechanics)", some subtree are put at the same level as their parent:
image

"Generalization to multi-parameter case" is the parent node, all the others in the picture are its child nodes. But they are not hidden then the parent is collapsed. Not sure whether it is a problem of online Wiki or the offliner tool.

maxi Version "6,134,097 articles in English"

MananJethwani:
@kelson42 this fixes #1501, fixes #1501.
The main approach used here is a bit complex I will elaborate it here in points for better understanding -
1) section template now has 2 anchors one for subsection addition another for same level anchor additions
2) I have used a stack to implement the tree structure, The sections received from API are in a specific tree structure order so if we consider various toclevel to be `1 1 2 1` for an example case, we can clearly use stack to store which parent sections are present for a specific sub-section so that we can close a section anchor and subsection anchors.
3) for the first section we receive I am directly adding it to HTML and pushing its toclevel to the stack so, according to our example HTML content would look like 
```
<>
section level 1 text
__SUB_LEVEL_SECTION_2__
</>
__SAME_LEVEL_SECTION_2__
```
and the stack is [1]
note - I am adding a level of an anchor with +1 to the current toclevel just so that I don't have to use another value in the template (it works fine according to test cases I created).
4) for any section which has the same value as the top of the stack I simply remove the sub-level section anchor as no other subsection could be created for that tag and we are moving on to fill the next tag so in our case. So In our case, HTML content would be like -
```
<>
section level 1 text
</>
<>
section level 1 text (2nd section)
__SUB_LEVEL_SECTION_2__
</>
__SAME_LEVEL_SECTION_2__
```
and the stack is [1]
for the same level, I am not pushing anything upon stack as for any level, we will only have one sub section anchor and one same level anchor to track so no need to push twice.
5) if a section has a value greater than the value at the top of the stack add it as a sub-section using the anchor tag and the value at the top of the stack refer -> L161 of `utils/articleRenderer.ts`.
so according to our example, HTML content will be like -
```
<>
section level 1 text
</>
<>
section level 1 text (2nd section)
    <>
         section level 2 text
         __SUB_LEVEL_SECTION_3__
    </>
    __SAME_LEVEL_SECTION_3__
</>
__SAME_LEVEL_SECTION_2__
```
and the stack is - [1, 2]
6) if a section has a value less than the value at the top of the stack it means for the current level there are no more sections that are on the same level or are the subsection under this parent so we remove the sub-level section anchor as well as same level section anchors and then use the above-mentioned condition in point 4 and 5 to create the tree structure further.
so according to our example, HTML content will be like - 
```
<>
section level 1 text
</>
<>
section level 1 text (2nd section)
    <>
         section level 2 text
    </>
</>
<>
section level 1 text(3rd section)
__SUB_LEVEL_SECTION_2__
</>
__SAME_LEVEL_SECTION_2__
```
and the stack is - 1
7) remove all remaining anchors using the remaining values in the stack.
P.S. - sorry for the detailed and long explanation😅, but I think was needed to keep my points clearly in one go.

### DIFF
--- a/res/templates/section_wrapper.html
+++ b/res/templates/section_wrapper.html
@@ -1,4 +1,4 @@
-<details data-level="2" open>
+<details data-level="{{ section_index }}" open>
     <summary class='section-heading'><h2 id="{{ section_anchor }}">{% autoescape false %}{{ section_line }}{% endautoescape %}</h2></summary>
     {% autoescape false %}{{ section_text }}{% endautoescape %}
     __SUB_LEVEL_SECTION_{{ section_index }}__
--- a/res/templates/section_wrapper.html
+++ b/res/templates/section_wrapper.html
@@ -3,3 +3,4 @@
     {% autoescape false %}{{ section_text }}{% endautoescape %}
     __SUB_LEVEL_SECTION_{{ section_index }}__
 </details>
+__SAME_LEVEL_SECTION_{{ section_index }}__
--- a/res/templates/section_wrapper.html
+++ b/res/templates/section_wrapper.html
@@ -3,4 +3,3 @@
     {% autoescape false %}{{ section_text }}{% endautoescape %}
     __SUB_LEVEL_SECTION_{{ section_index }}__
 </details>
-__SAME_LEVEL_SECTION_{{ section_index }}__
--- a/src/util/articleRenderers.ts
+++ b/src/util/articleRenderers.ts
@@ -4,8 +4,7 @@ import {
     leadSectionTemplate,
     sectionTemplate,
     subCategoriesTemplate,
-    subPagesTemplate,
-    subSectionTemplate
+    subPagesTemplate
 } from '../Templates';
 import logger from '../Logger';
 import type {Dump} from '../Dump';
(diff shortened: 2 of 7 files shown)


kelson42:
@MananJethwani Hi, any news here?

res/templates/section_wrapper.html
```diff
@@ -1,4 +1,4 @@
-<details data-level="2" open>
+<details data-level="{{ section_index }}" open>
     <summary class='section-heading'><h2 id="{{ section_anchor }}">{% autoescape false %}{{ section_line }}{% endautoescape %}</h2></summary>
```
kelson42, line 2: I think this can not always be `h2`, it should be `h1`, `h3`.. depending of the level right?

res/templates/section_wrapper.html
```diff
@@ -1,4 +1,4 @@
...
     {% autoescape false %}{{ section_text }}{% endautoescape %}
     __SUB_LEVEL_SECTION_{{ section_index }}__
```
kelson42, line 4: Why we just don't have here something like `{{ subsections_text }}`?

src/util/articleRenderers.ts
```diff
@@ -107,6 +106,55 @@ export const renderDesktopArticle = (json: any, articleId: string, articleDetail
...
+        section_text: sections[0].text,
+        strings: dump.strings,
+    });
+
+    sectionsToSplit = sectionsToSplit.slice(1);
+
+    // Loop over sections on current level and recursively generate html for it's sub-section.
+    sectionsToSplit.forEach((currentIndex: number) => {
+        html = html.replace(`__SUB_LEVEL_SECTION_${sections[prev].toclevel}__`,
+            prev + 1 === currentIndex ? '' : getSectionsHtml(sections.slice(prev + 1, currentIndex), dump));
```
kelson42, line 138: This part should just be integrated directly in the `sectionTemplate()` call to the new paramter `subsections_text`. So no html.replace anymore.

src/util/articleRenderers.ts
```diff
@@ -107,6 +106,55 @@ export const renderDesktopArticle = (json: any, articleId: string, articleDetail
...
+            section_line: sections[currentIndex].line,
+            section_text: sections[currentIndex].text,
+            strings: dump.strings,
+        });
+        prev = currentIndex;
+    });
+
+    // Get html for last sub-section.
+    html = html.replace(`__SUB_LEVEL_SECTION_${sections[prev].toclevel}__`,
+            prev + 1 === sections.length ? '' : getSectionsHtml(sections.slice(prev + 1), dump));
```
kelson42, line 153: If previous comments are implement, that should not be necessary anymore.

src/util/articleRenderers.ts
```diff
@@ -107,6 +106,55 @@ export const renderDesktopArticle = (json: any, articleId: string, articleDetail
...
+        section_anchor: sections[0].anchor,
+        section_line: sections[0].line,
+        section_text: sections[0].text,
+        strings: dump.strings,
+    });
+
+    sectionsToSplit = sectionsToSplit.slice(1);
+
+    // Loop over sections on current level and recursively generate html for it's sub-section.
+    sectionsToSplit.forEach((currentIndex: number) => {
```
kelson42, line 136: I think you can just replace all of this by a loop which does not go through all items one by one, but goes through all items via list of slices of items. Each slice gathering the first item and the smaller ones which follow. Something like this:
```
foreach slice = getNextSectionSlice(sections) {
  sectionTemplate({
        section_index: slice[0].toclevel,
        section_id: slice[0].id,
        section_anchor: slice[0].anchor,
        section_line: slice[0].line,
        section_text: slice[0].text,
        strings: dump.strings,
        subsections_test: getSectionsHtml(slice[1..n])
});
}
```

I see no reason why you should have a specail treatment

kelson42 (changes requested):
This is not simple enough, I believe it can be still a lot simpler. I'm not even sure this version works fine.

test/unit/util.test.ts
```diff
@@ -43,6 +44,49 @@ test('Encoding ArticleId for Zim HTML Url', async(t) => {
...
+        },
+        {
+          id: 5,
+          text: 'level 3',
+          toclevel: 3,
+          line: 'Non-adiabatic states',
+          anchor: 'Non-adiabatic_states'
+        }
+    ];
+    t.equal(getSectionHtml(sections, 1, dump), `<details data-level="2" open>\n    <summary class=\'section-heading\'><h2 id="Approximate_Hamiltonians">Approximate Hamiltonians</h2></summary>\n    level 1\n    \n</details><details data-level="2" open>\n    <summary class=\'section-heading\'><h2 id="Applying_perturbation_theory">Applying perturbation theory</h2></summary>\n    level 1\n    <details data-level="2" open>\n    <summary class=\'section-heading\'><h2 id="Limitations">Limitations</h2></summary>\n    level 2\n    <details data-level="2" open>\n    <summary class=\'section-heading\'><h2 id="Large_perturbations">Large perturbations</h2></summary>\n    level 3\n    \n</details><details data-level="2" open>\n    <summary class=\'section-heading\'><h2 id="Non-adiabatic_states">Non-adiabatic states</h2></summary>\n    level 3\n    \n</details>\n</details>\n</details>`,
```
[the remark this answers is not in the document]
kelson42: This is not tested accurately, otherwise the test would have remarked that there is only `h2`.

PR #1551, opened 2021-10-16
Update multiple dependences to recent version

kelson42:
Make a global upgrade of the dependences where this is possible.


kelson42:
@MananJethwani I try to update the base packages but have difficulties with `p-map` and `axios` which need (I guess small modification of the code). Would you please able to run:
```
$npm install axios@latest
$npm install p-map@latest
```

and secure the tests pass please?

kelson42:
@kelvinhammond This is an old attempt of upgraded a few dependences. I will now close it. Actually a few one are pretty important IMO, but they require slight changes which are over my small typescript skills. I think in particular to `redis`, `async`, `axios` and `p-map`.

PR #1622, opened 2022-06-12
Update docker

kelson42:
@rgaudin  It seems the `node-redis` image has to be pushed manually right?


kelvinhammond (approved):

rgaudin:
> @rgaudin It seems the `node-redis` image has to be pushed manually right?

That's right.

Issue #1601, opened 2022-03-23
Request: mwoffliner mediawiki api selection to include mediawiki version of rest api


tim-moody:
In monitoring the urls requested by mwoffliner I observe what seems to be a check for available APIs and the code seems to use a pluggable api handler.

/w/api.php?action=query ... is tried and
/api/rest_v1/page/mobile-sections/Main_Page
/api/rest_v1/page/html/Main_Page 

https://www.mediawiki.org/wiki/API:Main_page#Other_APIs shows two versions of the rest api:

https://en.wikipedia.org/w/rest.php/v1/page/Earth
https://en.wikipedia.org/api/rest_v1/page/title/Earth

The first is for mediawiki 1.35+ and the second is ONLY FOR WIKIMEDIA PROJECTS.

You can see above that it is the second that is requested from mdwiki.org, which is mediawiki 1.35+ but not a a Wikimedia Project, and it returns a 404. However mdwiki.org would return a response to the first.

SO Would it be possible for mwoffliner to request the mediawiki version as well and include it in the selection?

The reason to do so is that it seems likely that using the /w/api endpoint leads to the misalignment problem in #1598 and I also believe to the inclusion of full articles instead of leads in the mini format. (The /w/api return does not have the expected structure and so the mwoffliner parse does not identify the lead section.)

kelson42:
Ok, but this is a new API with a dedicated json/xml output. Just tonsay this is not a trivial task. But the request is legitim and we will do it. This has not been done so far because this API does not provide a mobile variant. The MWoffliner approach since 4 years was to force the mobile output (but this was not the case before). This paradigm is going to change (again) with #1623.

kelson42:
At this stage, we mostly need to wait the local Parsoid to be removed.

tim-moody:
makes sense.

kelson42:
Local Parsoid has been removed. But i believe we should rearchitecture a bit the code to provide a standard interface. Each alternative to scrape HTML working then asxa module and sharing the same inzerface.

kelson42:
> @kelson42 I am asking for the api pattern https://mdwiki.org/w/rest.php/v1/page/Gout

@tim-moody Sorry but I just realised this API end-point does not deliver HTML. Does the new Mediawiki Rest API does has an end-point to deliver HTML at all? Even Desktop?

tim-moody:
[...]

PR #1644, opened 2022-07-10
No local parsoid

### LINKED ISSUE #1623: Remove (local) Parsoid from the dependences/source
The Javascript based Parsoid has been deprecated years ago. We use it in MWoffliner to provide HTML in case there would be no appropriate backend running in the remote instance. This is a pity, but this old piece of software brings now problem because of the many outdated dependences it has. It should be removed... and the depending features as well.

kelson42:
This PR removes the local JS Parsoid and related options. The reason is that this feature was depending on the Javascript based Parsoid and [this parser is out of maintenance than more than a year](https://www.mediawiki.org/wiki/Parsoid). This situation was starting to generate many problems, in particular in relation with deprecated dependences.

The removal of the local Parsoid option will lead to the situation that, if the remote instance of Mediawiki does not provide an API delivering HTML for its articles, then MWoffliner won't be able to scrape it anymore. This should not be a problem with Wikimedia based wikis, but otherwise this will be the case if no VisualEditor is installed on the remove wiki.

Luckily, since end of 2019, Mediawiki core started to provide a [core REST API](https://www.mediawiki.org/w/index.php?title=API:REST_API) which provides HTML. Therefore recent Mediawiki instances should provide out-of-the-box HTML. But, this is still not supported so far in MWoffliner, see #1601 feature request.

An other problem is that local Parsoid option was delivering mobile HTML. This is important because since many years MWoffliner does always deliver only mobile HTML (except for the welcome page). Core API Rest API does not deliver mobile HTML and AFAIK this is not going to happen soo (if it ever happen). Therefore we will have to implement/reimplement probably option to allow/disallow Desktop HTML rendering (and therefore get out of the mobile only paradign). All of this has still to be discuss... and does anyway not change anything about the fact that we can not rely longer on deprecated code.

Fixes #1623


kelson42:
@kelvinhammond Do you feel confortable to review this PR or should I just merge it?

kelson42:
I’ll just merge the PR then. Hopefuly this is good enough.

PR #1665, opened 2022-10-16
config: add scripts for Wiktionary's show/hide toggles

### LINKED ISSUE #1033: Wiktionary translations can't be shown

zhuowei:
This fixes #1033.

The `ext.gadget.VisibilityToggles` and `ext.gadget.defaultVisibilityToggles` only seem to exist on Wiktionary: when I tested with English Wiktionary, they fixed the toggles. When I tested with English Wikipedia, the scripts returned `mw.loader.state({
    "ext.gadget.defaultVisibilityToggles": "missing"
});`

I can't find any good way to determine whether a wiki uses toggles: https://phabricator.wikimedia.org/T161278 seems to suggest that the API doesn't surface gadgets in `modulescripts`?

@ MananJethwani found that it's not possible to scrape them from `RLPAGEMODULES` either, at least through the API: https://www.mediawiki.org/wiki/Topic:W2g7t4u4whlmkqho

Is there a way to add these for Wiktionary only, maybe via Wiktionary's Zimfarm config instead? Or maybe pull the wiki's homepage via the regular, non-API endpoint and check for gadget scripts?

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -63,7 +63,9 @@ const config = {
           'mediawiki.page.gallery.styles',
           'mediawiki.skinning.content.parsoid'].join('|'),
       ],
-      js: ['startup', 'jquery', 'mediawiki.base', 'site', 'mediawiki.util', 'mediawiki.page.ready'],
+      js: ['startup', 'jquery', 'mediawiki.base', 'site', 'mediawiki.util', 'mediawiki.page.ready',
+        'jquery.cookie', 'mediawiki.cookie', 'mediawiki.storage', 'ext.gadget.VisibilityToggles',
+        'ext.gadget.defaultVisibilityToggles'],
     },
 
     // Output directories for storing mw related js and css resources


kelson42:
@zhuowei Thank you for your PR, I will review it.

tacsipacsi:
Instead of importing a custom gadget from Wiktionary, I think you should import the standard module `jquery.makeCollapsible`, and convince English Wiktionary to use it. This will be present on any Wikimedia wiki, helping more users than importing the custom gadget. (Or maybe the standard module isn’t broken at all?) For example, Hungarian Wiktionary uses the standard module, see https://hu.wiktionary.org/wiki/Sablon:trans-top as an example (unfortunately I couldn’t test whether it works, as Kiwix doesn’t seem to work on my Android 12 phone at all).

zhuowei:
> Hungarian Wiktionary uses the standard module

I checked: Hungarian Wiktionary on Desktop shows the box fully expanded by default; tapping it closes it - the opposite of English Wiktionary.

On mobile, it always shows the box expanded without any way to collapse it: https://hu.m.wiktionary.org/wiki/világ%C3%ADtótorony

Kiwix does the same as mobile currently, and always shows the box expanded: https://library.kiwix.org/wiktionary_hu_all_nopic_2022-09/A/világ%C3%ADtótorony

I did look into adding a CSS override that would force the infobox open on English Wiktionary, but I'd rather get the show/hide toggle working if possible.

tacsipacsi:
> I checked: Hungarian Wiktionary on Desktop shows the box fully expanded by default; tapping it closes it - the opposite of English Wiktionary.

True, due to the low number of languages per article, I decided not to collapse them by default. But this is just a matter of configuration: adding the `mw-collapsed` class would make it collapsed by default.

> On mobile, it always shows the box expanded without any way to collapse it: [https://hu.m.wiktionary.org/wiki/világ%C3%ADtótorony](https://hu.m.wiktionary.org/wiki/vil%C3%A1g%C3%ADt%C3%B3torony)

Oh yeah, maybe I should advocate for the standard module to work on mobile first…

> Kiwix does the same as mobile currently, and always shows the box expanded: [https://library.kiwix.org/wiktionary_hu_all_nopic_2022-09/A/világ%C3%ADtótorony](https://library.kiwix.org/wiktionary_hu_all_nopic_2022-09/A/vil%C3%A1g%C3%ADt%C3%B3torony)

Does Kiwix use the mobile or desktop/default versions of modules (where this matters)? If the desktop/default one, importing `jquery.makeCollapsible` would add collapsibility on huwiktionary in Kiwix regardless of it not working on the mobile web.

kelson42:
@zhuowei I still don’t have tested, but two remarks already:
* Obviously it has to continue to work with other sites like Wikipedia without generating errors
* Please add a comment in the code itself before the newly added modules, so we can understand the logic behind this… Maybe at some point the API will deliver the proper list and we will be able to remove this « custom » list!

kelson42:
@zhuowei I have superseeded this PR with #1701 to be able to add a comment. Even if the fix is not the perfect solution, this seems indeed the very best thing to do and this will make a lot of happy people. Thank you very much for this PR.

Issue #1674, opened 2022-10-26
Add support for Node 19.x


JameelKaisar:
Add support for Node 19.x

kelson42:
Seems not possible for the moment because not supported by the action. Will close for 5e moment, and we will support nodejs 20 when available.

PR #1676, opened 2022-10-26
Add support for Node 19.x

JameelKaisar:
Fixes: #1674

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


kelson42:
@JameelKaisar There is a problem, the CI does not run anymore

JameelKaisar:
The CI ran on the PR you merged. It failed but I'll fix it.

kelson42:
Strongly suspect that the problem here is that Node.js 19 is still not available in the env. proposed by Github.

JameelKaisar:
Then maybe we should not merge this for now.

PR #1678, opened 2022-10-26
Readjust Tests

JameelKaisar:
Readjust Tests


kelson42 (changes requested):
It's important we are not too tolerant here, if your reduce the minimum, reduce the maximum as well from the same amount.

JameelKaisar:
Fine. I'll reduce the minimum as well.

Issue #1666, opened 2022-10-16
CI is currently broken


kelson42:
On both my workstation and in the Github Action CI.

PR #1679, opened 2022-10-26
Fix CI

JameelKaisar:
Fixes: #1666 

The server returns Error 403 when 'user-agent' is passed and is an empty string. However when 'user-agent' is not passed at all, we get the desired response from the server.


JameelKaisar:
After fixing the empty user-agent in test file, the following two tests are failing:
1. 'Webp compression working for cmyk color-space images' in 'test/unit/downloader.test.ts'
```
operator: equal
expected: 'image/webp'
actual:   'image/jpeg'
```
2. 'passed test for jpg' in 'test/unit/webpAndRedirection.test.ts'
```
operator: ok
expected: true
actual:   false
```

kelson42:
> operator: equal
> expected: 'image/webp'
> actual:   'image/jpeg'

This is actually the symptom I have/had on my machine

JameelKaisar:
Any idea why user-agent should affect the file type?

Issue #1668, opened 2022-10-16
Publish automatically the mwoffliner package on npm


kelson42:
based on release tag on master

PR #1682, opened 2022-10-26
Add workflow to publish automatically to NPM

JameelKaisar:
This PR adds a workflow to publish automatically to NPM whenever a new tag starting with "v" is created.

Fixes: #1668


kelson42 (changes requested):
Few requests:
* name should better be "Publish to npmjs.com"
* Never use a moving release name in CI, so use `ubuntu-22.04` or similar (because it can change anytime and you don't control this... then your CI might fail surprisingly)
* It is not mandatory (but maybe a good security) to run the test in a release workflow, if you want to do so please find a way to run the otherwokflow to avoid code redundancy (see https://docs.github.com/en/actions/using-workflows/reusing-workflows to reuse a workflow)
* A release tag should only be valid if on branch `master`. I would prefer something like:
```
on:
  push:
    tags:
      - 'v{0,1}[0-9]+.[0-9]+.[0-9]+'
    branches:
      - main
      - master

```
* Document/Update the release procedure in `CONTRIBUTING.md`

JameelKaisar:
- Since the tags are related to commits (and not directly to branches), I think we don't need to check for the branch whenever a tag is pushed. Also, adding the following will run `pusblish.yml` workflow on every push to `main` or `master`.
```
on:
  push:
    ...
    branches:
      - main
      - master
```
- I merged Releasing and Publishing under same heading. The package gets published automatically whenever a new tag is pushed. I think the `dev/publish.sh` file is redundant now and also the `publishing script` in `ci.yml`, should I remove them?

kelson42 (changes requested):
* I don't want that someone put a tag on anything else than `master` and publish. This is why I have put this constraint
* You can delete `dev/publish.sh`
* In the documentation, it should be `git push origin vX.X.X` (keep the current branch, be specific about the tag you poush)
* In the documentation add a small sentence explaining that the CI will do the publishing work to npmjs.com

kelson42 (approved):

PR #1687, opened 2022-10-29
Create ru.json

Artem13327:
I warn you there is one problem in the first translation. Since what word will be there depends on some factors. And I chose the option that is more suitable


kelson42:
Please make anew PR.

Issue #1556, opened 2021-11-26
Skip problematic pages rather than halt


Sophivorus:
Hi! We've been trying to generate a ZIM file for appropedia.org for some time now. However, we've encountered several issues with problematic pages that halted the script, such as the ones described on [issue #1534](https://github.com/openzim/mwoffliner/issues/1534) but also others not reported here. The latest halt was due to a timeout from the page https://www.appropedia.org/Africa_community_action which seems pretty normal except perhaps for the map towards the bottom that may take some time to load. However these maps are pretty widespread throughout Appropedia (and other wikis) so removing them is hardly an option for us.

Thus I wonder if instead of continuing to fix problematic pages one by one, is there a configuration option that can be enabled or developed to skip problematic pages rather than halt the script completely. Thanks!

kelson42:
The latest scrape has been stopped becuase 'Cervical Cancer Screening and Treatment Self-Administered Readiness Test' is not readble. See https://www.appropedia.org/Cervical_Cancer_Screening_and_Treatment_Self-Administered_Readiness_Test.

There is no option to skip pages which are not readable and I don't really want to build that option because it will build "broken" ZIM files (with dead links for example).

But we should understand why this page is not readable and secure that such pages are not part of the ZIM snapshot. Do you know which extension is responsible for blocking this article?

kelson42:
Hmmm, I guess this is https://www.mediawiki.org/wiki/Extension:CategoryLockdown

kelson42:
@Sophivorus I guess, at this stage, the most pragmatic solution would be that you provide a liste of articles ids you want us to scrape (and which are readable) and we would scrape them. A one column CSV with one title (utf8, underscores) with articles ids.

Sophivorus:
Ah, I didn't notice the page was locked too (admins can see locked pages), my bad, sorry!
However, since this issue is about skipping problematic pages rather than about any specific problematic page, and since I think implementing such a feature would be reasonable, I won't close this issue yet, but feel free to do so if you have reasons not to implement the feature.
As to the list of ids, it's not that simple but I'll give it some thought and write back when I have something.

kelson42:
Extension:CategoryLockdown is clearly the real problem for now. We need a way to handle it. https://www.appropedia.org/Thermal_Ablation_of_Cervical_Pre-Cancer_Lesions Fails to scrape.

kelson42:
[...]

PR #1703, opened 2022-12-31
add skipBroken argument to skip erroneous articles

uriesk:
Adds a --skipBroken argument in order to just skip Articles that throws errors

Usecase:
When creating ZIM file from fandom wikis, the backend sometimes throws errors when the Article is too complex for their backend and it returns 503, like [Minecraft:Bedrock Edition data values/Metadata_table](https://minecraft.fandom.com/wiki/Bedrock_Edition_data_values/Metadata_table)
I need to skip those (very few, like 2 out of 8k) to successfully run mwoffliner.

### DIFF
--- a/articles
+++ b/articles
@@ -1,0 +1,4 @@
+Warden
+Bedrock_Edition_data_values/Metadata_table
+Resource_pack/Folders
+Creeper
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -443,7 +443,7 @@ async function execute(argv: any) {
     await getMainPage(dump, zimCreator, downloader);
 
     logger.log(`Getting articles`);
-    const { jsModuleDependencies, cssModuleDependencies } = await saveArticles(zimCreator, downloader, mw, dump);
+    const { jsModuleDependencies, cssModuleDependencies } = await saveArticles(zimCreator, downloader, mw, dump, !!argv.skipBroken);
 
     logger.log(`Found [${jsModuleDependencies.size}] js module dependencies`);
     logger.log(`Found [${cssModuleDependencies.size}] style module dependencies`);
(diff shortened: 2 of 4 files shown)


articles
```diff
@@ -0,0 +1,4 @@
+Warden
```
kelson42, line 1: This file should not be there (the whole file).

src/mwoffliner.lib.ts
```diff
@@ -443,7 +443,7 @@ async function execute(argv: any) {
     await getMainPage(dump, zimCreator, downloader);
 
     logger.log(`Getting articles`);
-    const { jsModuleDependencies, cssModuleDependencies } = await saveArticles(zimCreator, downloader, mw, dump);
+    const { jsModuleDependencies, cssModuleDependencies } = await saveArticles(zimCreator, downloader, mw, dump, !!argv.skipBroken);
```
kelson42, line 447: This option should be handled like the others

kelson42 (commented):
Thank you for your PR. But please, next time, first open a ticket and propose a solution to avoid what will come next.

I don't want such an option like this because it generate automatically incoherent ZIM files. I mean that this generates per design HTML article with broken link. This is something we do our maximum to avoid.

That said I understand the need, but this is not the best way to fix it IMO. What I would propose is to have something like `--articleListToIgnore` working on a similar principle like `--articleList`.

uriesk:
I will follow this up with an issue. It already existed in the past #1556

How do you want such an option to be handled?
I did handle it in the same way as getCategories.
Or do you want it like webp, and are fine with destructuring a potentially undefined value?

kelson42:
Yes, reading #1556 is interesting.

The solution would be to read this list of article to ignore and remove them from the list of articles retrieved at the first stage of the scraping process.

PR #1711, opened 2023-01-04
Detect [url].[ext]/revision/.* image-urls as webp convertible

### LINKED ISSUE #1710: `--webp` option assumes that image links end-up with an extension
But this is not alway the case, look for example to this Fandom wiki page:
https://minecraft.fandom.com/de/wiki/Minecraft_Wiki

Image `href` values are like:
https://static.wikia.nocookie.net/minecraft_de_gamepedia/images/9/90/Minecraft_Wiki_header.svg/revision/latest/scale-to-width-down/300?cb=20211024194004

The URL rewriting generates:
http://192.168.178.24:8181/content/minecraftwiki_de_all_maxi_2023-01/I/ebe7173eb06396374c79c11fd3a51cb3

The image is in the ZIM at the URL `I/ebe7173eb06396374c79c11fd3a51cb3.webp`

uriesk:
fixes #1710


kelson42:
@uriesk Thank you for your PR. Reviewing it, I have chosen to follow a slightly different past and introduce a unit test at  #1711. But your PR looked good otherwise.

PR #1712, opened 2023-01-05
Handle properly Fandom image src links

### LINKED ISSUE #1710: `--webp` option assumes that image links end-up with an extension

kelson42:
Fixes #1710
Superseeds #1711


uriesk:
Consider that the Downloader is deciding on whether he converts or not based on the content-type, which is not available at the article scrapping stage. So a new regexp for [url].[ext]/revision/* can't be avoided.

pavel-karatsiuba (approved):
LGTM

PR #1245, opened 2020-08-01
issue/1176/redis-iterate

midik:
Fix #1176

PR #1714, opened 2023-01-06
Alternative implementation of 'iterateItems()'

### LINKED ISSUE #1176: Remarks about RedisKVS.iterateItems()
Looking around in MWoffliner code I came to `RedisKvs.iterateItems()` (and `RedisKvs..iterateItems()`). https://github.com/openzim/mwoffliner/blob/master/src/util/RedisKvs.ts#L165

If I understand properly this function is at the core of the iteration over all articles/images/, has a `while (true)` loop.... but not log entry at all! Sounds a good candidate to me about where it could fail.

So here a few remarks/proposals:
* it should be a `do... while()` test, not a `while`.
* `depleted` is useless, we should just return/exit (if necessary, should not).
* `RedisKvs.scan()` `reject(err)` seems not handled at all!
* I don't really get it between `RedisKvs.scan()` and `pendingScan: Promise<ScanResult>`. To me they should be merged in one method and the only thing shared should be the `scanCursor`.
* This line `const items = Array.from(data, (x, k) => k % 2 ? undefined : [x, data[k + 1]]).filter((x) => x);` is a mystery to me and I wonder what it does.
* We should not `break` IMO, the `while` loop should be exited properly with its checks if necessary.
* The line `scanCursor = cursor` create the conditions of a problem by lately resetting the scanCursor which might have been changed meanwhile by an other async method.
* There is no security considering that the same element of Redis can be returned many times.
@midik Maybe you have more to say/change to this. I have the strong feeling that this iterator is not done properly.

uriesk:
Different implementation of `.iterateItems()`.

- Does not use p-map.
- Should be neither a performance benefit nor regression.
- Potential better handling of errors.
- Addresses originally raised issues in #1176.
- There are no workerIds, it keeps track of those by the number of currently active workers alone, which is now returned to the worker functions instead of the id.

testing and benchmarks welcome

Fixes #1176

Superseeds #1245

### DIFF
--- a/src/util/RedisKvs.ts
+++ b/src/util/RedisKvs.ts
@@ -162,45 +162,70 @@ export class RedisKvs<T> {
     });
   }
 
-  public async iterateItems(numWorkers: number, func: (items: KVS<T>, workerId: number) => Promise<void>) {
-    const workers = Array.from(Array(numWorkers).keys());
+  public iterateItems(
+    numWorkers: number,
+    func: (items: KVS<T>, runningWorkers: number) => Promise<void>,
+  ) {
+    return new Promise((resolve, reject) => {
+      let runningWorkers = 0;
+      let isScanning = false;
+      let done = false;
+      let failed = false;
+      let scanCursor = '0';
 
-    let scanCursor = '0';
-    let depleted = false;
-    let pendingScan: Promise<ScanResult> = Promise.resolve(null);
+      const scan = async () => {
+        if (runningWorkers >= numWorkers || isScanning) {
+          return;
+        }
+        if (done || failed) {
+          if (!runningWorkers) {
+            resolve();
+          }
+          return;
+        }
+        isScanning = true;
 
-    await pmap(
-      workers,
-      async (workerId) => {
-
-        while (true) {
-          pendingScan = pendingScan.then(() =>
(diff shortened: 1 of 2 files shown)


kelson42:
@uriesk Thank you for your PR! Not sure you are over, but if "yes", please assign myself and @pavel-karatsiuba as reviewer. Your speed challenges me anyway, so don't wait a feedback from me before the WE.

Please put a small high level description in the PR explaining the benefits (and new limitations) of this alternative implementation and how this is relased to the ticket. It's important to grab the essence of your patch for people who are not typescript devs... or simply want to avoid a complete code scan.

On my side, will do my best to review the code and make a few benchmarks.

Key point is to see if it help to fix these sporadic and silent freezes we have with huge wikis. BTW, if you have identified such scenarios, please put log entries (only in verbose mode) so we might be able to identify them on the Zimfarm.

Anyway, such big scrapes can obly be done on rhe Zimfarm, which means post-merging. I can not run on the zimfarm wirh code coming from a feature branch. Thetefore, we mostly need to check for regressions here during the review.

pavel-karatsiuba (approved):
LGTM

kelson42 (approved):
LGTM

Issue #1699, opened 2022-12-30
Use ESLint instead of TSLint


pavel-karatsiuba:
TSLint has been deprecated as of 2019.
Use ESLint with modern rules.

kelson42:
Fixed by #1738

PR #1719, opened 2023-01-07
Batch wikilinks redis requests on article parsing

### LINKED ISSUE #1718: Batch isMirrored redis requests on article parsing
When an article is parsed, every single wikilink within the article checks if the linked article is going to be parsed too and available in [isMirrored](https://github.com/openzim/mwoffliner/blob/main/src/util/rewriteUrls.ts#L16), which gets sequentially awaited.

I checked how big the impact is, and on parsing the 19.577 articles of the [pokemon wiki](https://pokemon.fandom.com) i got 2.685.074 wikilinks, spending 53min17s awaiting redis HGET.
A basic improvement would be to batch the requests per article and filter out duplicates.
It's impossible to say how much the performance impact will be, since we are in workers waiting for network requests, but it's well worth it to try.

(follow up to what what i raised in a comment in  https://github.com/openzim/mwoffliner/issues/1176#issuecomment-1372409102)

uriesk:
Instead of checking every wikilink one-by-one if it is going to be fetched, batch all links of an article together.
Use `HEXISTS` instead of `HGET` for checking the existence of a field.

I did some basic benchmarks with smaller wikis and it is about 5% faster at article scrapping phase. I think only the zimfarm on larger wikis can tell us the actual impact.

closes #1718

### DIFF
--- a/src/util/RedisKvs.ts
+++ b/src/util/RedisKvs.ts
@@ -1,5 +1,4 @@
 import {cpus} from 'os';
-import pmap from 'p-map';
 import type {RedisClient} from 'redis';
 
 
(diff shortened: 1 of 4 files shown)


kelson42:
@uriesk 5% is quite good actually. Don't forget to remove draft when you are so far.

uriesk:
@kelson42 i was hoping for more. It's hard to benchmark because the bandwidth fluctuates by more than those 5%

got a question
When a redirect gets found, it links to the original, it was the same in the old code.
new:
https://github.com/uriesk/mwoffliner/blob/pipelining/src/util/rewriteUrls.ts#L247
old:
https://github.com/openzim/mwoffliner/blob/main/src/util/rewriteUrls.ts#L24
This is deliberately, right?

uriesk:
This looks good now.
I hope we can try a larger wiki with the zimfarm with those changes soon and see if things improved :)

kelson42 (approved):
I have made made a superficial review. It seems to work but I see no speedup at all. That said, the principle is better. I think, and this would be something additional, that each function would be nefit of a small comment explaining what it does, in particular `rewriteUrlNoArticleCheck()`.

Quite a bit of code has been moved/changed, just hope that nothing fondamental in the link treatment has been changed. That only the architecturing has changed.

If OK for @pavel-karatsiuba, I would merge it and give it a try, but don't really expect much impact anyway.

uriesk:
Thought so.
I am running it in  an environment with limited CPU speed but good bandwidth and have to time the article downloading phase specifically to see a tiny improvement.

That being said, spaming less requests to redis is always a good idea and it's one potential issue out of the way.

`rewriteUrlNoArticleCheck()` is the same as `rewriteURL` was before, just some ifs flattened, mediaDependencies in the arguments and wikilink parsing at the end moved out.
I can do elaborate comments when the switch to eslint is happening. There is lots of potential refactoring going to happen, with how touchy eslint can be.

kelson42:
@pavel-karatsiuba works on #1699 which should be ready very soon.

pavel-karatsiuba (approved):

PR #1721, opened 2023-01-10
Detect when saveArticles worker freezes

uriesk:
When a saveArticles worker freezes, detect it and throw an error (indicating at what stage the freeze happened) - rather than continuing with the other workers for days and then freezing at 100%.

I will keep this as a draft suggestion.
Can consider it if [wikipedia_en_all_maxi](https://farm.openzim.org/recipes/wikipedia_en_all_maxi) keeps freezing.

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -97,6 +97,7 @@ class Downloader {
   public readonly mw: MediaWiki;
   public loginCookie: string = '';
   public readonly speed: number;
+  public readonly requestTimeout: number;
   public baseUrl: string;
   public baseUrlForMainPage: string;
   public cssDependenceUrls: KVS<boolean> = {};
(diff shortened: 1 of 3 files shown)


kelson42:
@uriesk Test running at https://farm.openzim.org/pipeline/5b113c88aca15ced0709cb36

uriesk:
> @uriesk Test running at https://farm.openzim.org/pipeline/5b113c88aca15ced0709cb36

I was watching it and it's the reason why i made this here.
I was noticing that the the new build had a constant speed, always about 330 seconds between the 0.1% reports.
While the old build, that froze, got larger and larger gaps between and it probably spent its last 30% with multiple broken workers.

image

I plotted the gaps, top image the current running one, bottom the old build.
The new build just got a peak, so it might be already dead and we just don't know it yet.

also bash one-liner to get the gaps out of a textfile with the stdout called `stdoutfile`:
```bash
OD="0"; for i in `cat stdoutfile | grep "Progress downloading articles" | sed -e 's/^\[log\] \[//' -e 's/].*//g'`; do ND=`date -d $i +%s`; expr $ND - $OD; OD=$ND; done
```

uriesk:
superseded by #1728

PR #1743, opened 2023-01-19
Re-add a page with subpages

### LINKED ISSUE #1739: E2E artileList tests failing
in all recent pull requests, two e2e tests fail,

1. `Simple customMainPage` in ` test/e2e/extra.e2e.test.ts`
because it gets 6 out of 7 pages, probably because [Wikipedia:Books/archive/Ears_nose_throat](https://en.wikipedia.org/wiki/Wikipedia:Books/archive/Ears_nose_throat) can not be found

2. `Simple article List` in test/e2e/en10.e2e.test.ts
because it gets less files than it expects.

kelson42:
New attempt to fix #1739

### DIFF
--- a/test/e2e/extra.e2e.test.ts
+++ b/test/e2e/extra.e2e.test.ts
@@ -24,7 +24,8 @@ describe('Extra', () => {
 1%_(South_Park)
 İznik
 Egyptian_hieroglyphs
-AC/DC`;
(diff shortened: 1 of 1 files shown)


kelson42:
@uriesk I have added a page with subpage... but I don't see any change on the code coverage!

PR #1744, opened 2023-01-19
fix subpage code coverage

### LINKED ISSUE #1739: E2E artileList tests failing

uriesk:
fix #1739

### DIFF
--- a/test/e2e/extra.e2e.test.ts
+++ b/test/e2e/extra.e2e.test.ts
@@ -24,6 +24,7 @@ describe('Extra', () => {
 1%_(South_Park)
 İznik
 Egyptian_hieroglyphs
+Wikipedia:Books/archive/Cancer care
 AC/DC`;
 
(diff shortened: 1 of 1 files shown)


uriesk:
@kelson42 i haveno idea what is going on there.
If i put a logger.error in the Subpage code, it gets printed when testing, so the code is being hit.
But codeconv doesn't get it.
I would just merge this one and call it a day.

PR #1748, opened 2023-01-21
Set delay to wait until S3 uploading finished

### LINKED ISSUE #1733: S3 testing should be part of the CI
It seems is not, probably because the credentials are not available.

pavel-karatsiuba:
Fix #1733

### DIFF
--- a/test/unit/downloader.test.ts
+++ b/test/unit/downloader.test.ts
@@ -237,13 +237,9 @@ describe('Downloader class', () => {
     })
 
     test('Delete image from S3', async () => {
-      const randomImageUrl = async () => {
-        const url = await getRandomImageUrl()
-        return isImageUrl(url) ? url : randomImageUrl() // recursion to get URL with image in needed format
-      }
 
       // Check Etag Flow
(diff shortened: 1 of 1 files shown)


kelson42:
@pavel-karatsiuba What does this PR exactly? S3 errors should make the CI die, there should not be S3 errors.

pavel-karatsiuba:
@kelson42  images are uploads to S3 asynchronously but the test is not waiting to finish uploading. So I added a delay to wait for uploading.

kelson42 (approved):
The original test code what poorly written, but this PR has been problematic in many manners:
* First it made the whole core S3 upload synchronous because the test was broken. This is a no-go on the principle but as well because it slows down everything.
* Then it has added a delay (which seems an acceptable approach to me even if pretty unelegant, that said I don't have a better one). OK for that.
* But it changed the logic of the test in a bad manner: just adding `not` to avoid undestanding what is wrong.
* On the top I have seen that many other things were quite wrong with the test and not effort has been made to improve it, even at comment level (which is a bit sad)

I have fixed all of this - properly hopefuly.

PR #1751, opened 2023-01-23
alternative retry strategy for downloading files

### LINKED ISSUE #1750: Slow-down on 429 maybe not optimal
Currently, when we get a 429, we slow down by running less downloads in parallel. So basically by indirectly limiting bandwidth.
But if the rate-limit upstream is by requests per minute, this is not optimal, because we will eventually run into lots of small files, hitting the limit sooner and slowing down much further than we have to, when we eventually download larger files again.

Do we know how the upstream wikimedia rate-limit works on downloading files?

uriesk:
Don't immediately retry failed file requests again.
Instead use the already existing filesToRetryXPath store multiple times to retry after the rest is done.

fixes #1750

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -393,25 +393,36 @@ class Downloader {
     })
   }
 
-  public async downloadContent(_url: string): Promise<{ content: Buffer | string; responseHeaders: any }> {
+  public async downloadContent(_url: string, retry = true): Promise<{ content: Buffer | string; responseHeaders: any }> {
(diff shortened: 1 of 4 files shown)


uriesk:
looks good

kelson42:
@uriesk Thx, will have a look this WE and merge if OK in priority

kelson42 (approved):
LGTM

pavel-karatsiuba (approved):
LGTM

PR #1752, opened 2023-01-25
Move inline JS from html to js file

### LINKED ISSUE #1578: Remove inline javascript to comply with some CSP
There is at least some pieces of inline javascript in every page, like :
 ```
  <script>
    function importScript() { return 1 } // this is to avoid the error from site.js
  </script>
  <script>let articleId = 'Wikivoyage:Offline_reader_Expedition/Europe_Home_page'</script>
````
the articleId is used afterwards to handle webp.

The content can be browsed anyway, at least on the few ZIM files I tested

pavel-karatsiuba:
Setting variable articleId was removed from HTML.
Article ID was added to DOM element attribute to read it from javascript.
Fixes #1578

### DIFF
--- a/res/script.js
+++ b/res/script.js
@@ -1,3 +1,5 @@
+function importScript() { return 1 } // this is to avoid the error from site.js
+
 window.onload = function () {
 
     /* Collapsing of the sections */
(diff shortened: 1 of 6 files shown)


kelson42:
@pavel-karatsiuba Please put a PR description

kelson42 (approved):
LGTM

Issue #1663, opened 2022-10-05
Implement Metadata LongDescription


kelson42:
with a new command line argument `--longdescription`

kelson42:
@pavel-karatsiuba Hopefuly node-libzim allows to do so, it is basically like the « description » zim metadata. Let me know if something unclear.

PR #1755, opened 2023-01-26
Implement Metadata LongDescription

pavel-karatsiuba:
Add new parameter customZimDescription which set LongDescription to zim file

Fix: #1663


kelson42 (changes requested):
* value should not be sliced
* If not set it should not be empty string, it should simply be undefined
* Values shoukd sanitized and if too long an error shoukd be generated

kelson42 (changes requested):
* Argument check and sanitation should be in `sanitize-argument.ts`
* Please add same form `description` ([should not be longer than 80 characters](https://wiki.openzim.org/wiki/Metadata))
* please add an automated test to test properly all the ZIM Metadata using `zimdump`. For example `zimdump show --url "M/LongDescription".

Issue #1662, opened 2022-09-25
site.js should be loaded last to avoid "Cannot read properties of undefined (reading 'getParamValue')"


zhuowei:
To reproduce:
- Open a mwoffliner-created Zim file in any reader e.g. https://library.kiwix.org/wiktionary_en_all_maxi_2022-09/A/potato
- See browser console:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'getParamValue')
    at site.js:108:14
    at fire (jquery.js:3500:31)
    at Object.add [as done] (jquery.js:3559:7)
    at site.js:106:35
```
- Note that JavaScript features such as expanding the translations/conjugations on the page do not work

---

I believe what's happening is this:
- site.js calls `mw.loader.using('mediawiki.util')`

```
mw.loader.using('mediawiki.util').done(function() {
    /** &withmodule= query parameter **/
    if (mw.util.getParamValue('withmodule'))
```
- mwoffliner forces this function to return immediately, since it's assumed that in a .zim, all scripts are loaded ahead of time:

https://github.com/openzim/mwoffliner/blob/abd51529c00541b6fb5004e1d47917c9e5c34147/src/util/dump.ts#L104

- but the page's JavaScript list puts site.js ahead of mediawiki.util.js:

```
  <script src="../-/mw/jsConfigVars.js"></script>
  <script src="../-/mw/startup.js"></script>
<script src="../-/mw/jquery.js"></script>
<script src="../-/mw/mediawiki.base.js"></script>
<script src="../-/mw/site.js"></script>
<script src="../-/mw/mediawiki.util.js"></script>
<script src="../-/mw/mediawiki.page.ready.js"></script>
<script src="../-/mw/ext.tmh.player.js"></script>
```

- I believe the list of JS is generated in:
  - https://github.com/openzim/mwoffliner/blob/54dfafff4086255180b6d3ddd3fc196754926560/res/templates/page.html#L32
  - https://github.com/openzim/mwoffliner/blob/5b5cbdd8321b3e22ecc67d0aebd9278dadbc6cbe/src/util/saveArticles.ts#L895
  - https://github.com/openzim/mwoffliner/blob/5b5cbdd8321b3e22ecc67d0aebd9278dadbc6cbe/src/util/saveArticles.ts#L311

- so we can probably fix this by sorting the list of JavaScript files so that anything named "site.js" and/or is sourced from "MediaWiki:Common.js" gets put at the end?

[...]

PR #1757, opened 2023-01-27
The sequence of the js files loading was changed to fix the "unknown variable" error

pavel-karatsiuba:
site.js script should be loaded after mediawiki.util.js to fix the error: "Cannot read properties of undefined (reading 'getParamValue')" 

Fixture: #1662

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -73,8 +73,8 @@ const config = {
         'startup',
         'jquery',
         'mediawiki.base',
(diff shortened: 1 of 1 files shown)


kelson42:
@pavel-karatsiuba I will merge this. Please put your findings in the #1662 and we will fix the rest in an other PR.

kelson42 (approved):

PR #1763, opened 2023-02-01
treatment links with *.oga file

pavel-karatsiuba:
Save oga files with the same behavior as ogg.
If novid format is set then replace <a> tag with HTML text.


kelson42 (changes requested):
@pavel-karatsiuba Please update one of the tests to secure this scenario is tested.

kelson42 (approved):
LGTM

PR #1766, opened 2023-02-03
if <script> with 'mw.config.set' is not found then grab <script> with config variables

### LINKED ISSUE #1662: site.js should be loaded last to avoid "Cannot read properties of undefined (reading 'getParamValue')"

pavel-karatsiuba:
Scraper takes article configuration from <script> tag of HTML page.
Right now API is not provided such configuration.
With this PR the configuration takes in another way.
Fixes #1662

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -369,7 +369,7 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
   }
 }
 
-async function getModuleDependencies(articleId: string, mw: MediaWiki, downloader: Downloader) {
+export async function getModuleDependencies(articleId: string, mw: MediaWiki, downloader: Downloader) {
   /* These vars will store the list of js and css dependencies for
     the article we are downloading. */
   let jsConfigVars = ''
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -411,10 +411,12 @@ async function getModuleDependencies(articleId: string, mw: MediaWiki, downloade
   for (let i = 0; i < scriptTags.length; i += 1) {
     if (scriptTags[i].text.includes('mw.config.set')) {
       jsConfigVars = regex.exec(scriptTags[i].text)[0] || ''
+      jsConfigVars = `(window.RLQ=window.RLQ||[]).push(function() {${jsConfigVars}});`
+    } else if (scriptTags[i].text.includes('RLCONF') || scriptTags[i].text.includes('RLSTATE') || scriptTags[i].text.includes('RLPAGEMODULES')) {
+      jsConfigVars = scriptTags[i].text
     }
   }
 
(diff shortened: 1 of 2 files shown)


kelson42 (commented):
Should we not die if nothing is detected, so no `jsConfigVars`? Should we not build an automated test for this?

pavel-karatsiuba:
> Should we not die if nothing is detected, so no `jsConfigVars`? Should we not build an automated test for this?

I have added an automated test to show which data we are getting. Empty data is not a problem because we are taking these data from the source page and we assume that the source page has the right format.

kelson42 (approved):

PR #1769, opened 2023-02-05
Remove Node.js 16 and add 19

kelson42:


kelson42:
Merging, as @pavel-karatsiuba is not available for the next few days and we need to move forward.

PR #1770, opened 2023-02-05
CI workflow cleaning

kelson42:


kelson42:
@pavel-karatsiuba not been around for a while, I will merge this to allow to move easily forward. Hopefuly there is nothing broken/weird in my PR.

PR #1780, opened 2023-02-14
Treat hidden DOM elements before treating media files.

### LINKED ISSUE #849: Audio/Video within blacklisted CSS containers is still downloaded
Take this html:
```html
<div class='noprint'>
    <audio><source src='...'/></audio>
</div>
```

The source url is added to the list of medias to download in pre-processing, but the element itself is removed later in the treatment process. This means we download the content, but it's not visible

### LINKED ISSUE #1765: audio files present in zim file, but not in article
Scrap an article with an ogg audio file [like this one](https://en.wikipedia.org/wiki/Read_my_lips:_no_new_taxes):

```
npm start -- --adminEmail=contact@kiwix.org --articleList=Read_my_lips:_no_new_taxes --mwUrl=https://en.wikipedia.org/ --webp
```

In the resulting zim file, the audio is not in the article.
But the ogg file is visible in zimdum.
Tested this with current dev and with 1.11.12. It is the same in both cases.

pavel-karatsiuba:
Before downloading media files need to remove hidden elements from HTML

fixes #849
fixes #1765

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -433,6 +433,19 @@ async function processArticleHtml(
   let mediaDependencies: Array<{ url: string; path: string }> = []
   let subtitles: Array<{ url: string; path: string }> = []
   let doc = domino.createDocument(html)
+
+  const ruRet = await rewriteUrlsOfDoc(doc, articleId, redisStore, mw, dump)
+  doc = ruRet.doc
+  mediaDependencies = mediaDependencies.concat(
+    ruRet.mediaDependencies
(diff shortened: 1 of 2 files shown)


kelson42:
@pavel-karatsiuba please add a test. addind to one of the e2e test, an article with such a scenario should be enough

kelson42:
@pavel-karatsiuba please add a test. addind to one of the e2e test, an article with such a scenario should be enough

Does it fix #1765 as well?

kelson42 (approved):

Issue #674, opened 2019-04-14
--verbose should be able to get a log threshold


kelson42:
The `--verbose` option works as a boolean option for now. This should continue to be able to work that way but we should also be able to put a log level, under which, any log message should not be displayed.

For example `--verbose=warning` will print all errors above warning, ie. warning and error probably.

kelson42:
@uriesk Does that proposal sounda a reasonable way to be more flexible with logging printing on terminal?

uriesk:
@uriesk yes, sounds great

kelson42:
@pavel-karatsiuba seems ready to implement to me

PR #1785, opened 2023-02-16
Refactoring of Logger module

pavel-karatsiuba:
Use ['info', 'log', 'warn', 'error'] as values for --verbose

fixes: #674


kelson42 (changes requested):
* the value given should be checked, `--verbose=ddd` just pass. Like any other argument it should be check and if wrong an error should be displayed and mwoffliner should exit.
* In the usage, default value if nothing specified should be documented.
* It should be explain as well that the chosen level print also all the level greather than this. So if you should `warn`, you will get `warn` and `errors`
* Everything should be written to STDOUT except the errors, which should be printed to the STDERR

pavel-karatsiuba:
Added sanitizing for verbose option.
The verbose option can be empty or one of "info", "log", "warn", "error", or "quiet".
An empty value is equal to "info".
The default level is "error".
If --verbose=quiet then don't print messages.

kelson42 (changes requested):
Please rebase and resolve conflicts, have also left a small style feedback

kelson42 (approved):
LGTM

Issue #1784, opened 2023-02-16
Is `--addNamespaces="100"` no longer working? (to add Wikipedia's portal pages to ZIM files for schools!)


holta:
See `--addNamespaces="100"` at the bottom of these 2 recipes:

- https://farm.openzim.org/recipes/wikipedia_en_all/config
- https://farm.openzim.org/recipes/wikipedia_en_all_maxi/config

I believe this is supposed to add Wikipedia's portal pages to a ZIM file, and if I'm understanding @kelson42 this was working in the past.  Questions:
- Is this parameter no longer working?
- If so, any idea why?
- Or, am I looking in the wrong place when trying to find [English Wikipedia's ~540 portal pages](https://en.wikipedia.org/wiki/Wikipedia:Contents/Portals) within https://library.kiwix.org/viewer#wikipedia_en_all_maxi_2022-05/ ?

Schools are _always_ urgently ask for _far more colorful pages_ so as to animate Wikipedia-centric learning &mdash; among younger people especially &mdash; who commonly turn against Wikipedia (possibly for life?)  When elders too often present Wikipedia as text-heavy + ponderous 🐢 

So the question is: Can **colorful + youth-friendly** portal pages like the following (more graphical i.e. with much less text, very intentionally more inviting!) be made available + easily findable offline &mdash; ideally in many of the 87 Wikipedia languages[*] that offer portals?
- https://en.wikipedia.org/wiki/Portal:Ancient_Egypt
- https://en.wikipedia.org/wiki/Portal:Biology
- https://en.wikipedia.org/wiki/Portal:Caribbean
- https://en.wikipedia.org/wiki/Portal:Comics
- https://en.wikipedia.org/wiki/Portal:COVID-19

[*] 87 languages appear to have Wikipedia portals, as indicated in the top-right of https://en.wikipedia.org/wiki/Wikipedia:Contents/Portals

Brief code context: https://github.com/openzim/mwoffliner/search?q=addnamespaces

kelson42:
@pavel-karatsiuba please verify this ticket is valid, I’m really surprised it does not work?

holta:
_Thank you @pavel-karatsiuba:_

1) Discoverability (how to find these 540 portal pages) is the key UX question if have suggestion(s) for teachers?  (After we hopefully find them!)

[...]

PR #1787, opened 2023-02-17
added test for option addNamespaces

pavel-karatsiuba:
fixes: #1784


kelson42:
@pavel-karatsiuba AFAIK this does not explain nor fix what is reported in the bug ticket.

kelson42 (approved):
LGTM

Issue #1792, opened 2023-02-23
Simplify Axios Error in the log


kelson42:
When we have an Axios error, the log is too verbose, we get something like:
```
[error] [2023-02-03T03:13:40.230Z] Error downloading article World_History/The_Rise_of_Dictatorship_and_Totalitarianism/Quick_Quiz
Failed to run mwoffliner after [1441s]: {
	"message": "Request failed with status code 403",
	"name": "AxiosError",
	"stack": "AxiosError: Request failed with status code 403\n    at settle (file:///tmp/mwoffliner/node_modules/axios/lib/core/settle.js:19:12)\n    at IncomingMessage.handleStreamEnd (file:///tmp/mwoffliner/node_modules/axios/lib/adapters/http.js:512:11)\n    at IncomingMessage.emit (node:events:539:35)\n    at IncomingMessage.emit (node:domain:475:12)\n    at endReadableNT (node:internal/streams/readable:1344:12)\n    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)",
	"config": {
		"transitional": {
			"silentJSONParsing": true,
			"forcedJSONParsing": true,
			"clarifyTimeoutError": false
		},
		"adapter": [
			"xhr",
			"http"
		],
		"transformRequest": [
			null
		],
		"transformResponse": [
			null
		],
		"timeout": 10000000,
		"xsrfCookieName": "XSRF-TOKEN",
```

and then come hundreds of line. This is not useful and brings problem, see https://github.com/openzim/zimfarm/issues/751.

We should not have all this details (at least not what in the `config` part of the json... but actually I want a simple error with the HTTP error code and only the most important information.

FledgeXu:
Hi, I want to do this issue as my good first issue. Could you assign it to me?

FledgeXu:
Hi, I want to do this issue as my good first issue. Could you assign it to me?

PR #1795, opened 2023-02-23
Cleanup AxiosError redundant message.

FledgeXu:
Fix: https://github.com/openzim/mwoffliner/issues/1792

### DIFF
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -21,6 +21,7 @@ import {
   WEBP_CANDIDATE_IMAGE_MIME_TYPE,
 } from './const.js'
 import { fileURLToPath } from 'url'
+import { AxiosError } from 'axios'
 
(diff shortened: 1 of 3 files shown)


kelson42:
@FledgeXu Thank you!

pavel-karatsiuba (changes requested):
Please add tests for cases of fully filled err response and if the error object is filled only partially.

kelson42:
I would be OK to only unit test the cleanup function.

FledgeXu:
Ok, I will add it later.

FledgeXu:
I added the unit test case for cleanup function and I also adjust the location of the function.

pavel-karatsiuba (approved):
LGTM

pavel-karatsiuba (approved):

FledgeXu:
Thanks for your help!

PR #1798, opened 2023-02-27
Use comma separated URL's as value for --articleList parameter

### LINKED ISSUE #1729: Wikipedia en astronomy adding related WikiProjects
Please add project [WikiProject Spaceflight](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Spaceflight) to [Wikipedia en astronomy](https://farm.openzim.org/recipes/wikipedia_en_astronomy).

It will be nice if you can create a poll so users can vote.

Or modify  [Wikipedia en astronomy](https://farm.openzim.org/recipes/wikipedia_en_astronomy) in **Aerospace** with the addition of:

- [Wikipedia:WikiProject Astronomy](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Astronomy)
- [WikiProject Spaceflight](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Spaceflight)
- [Wikipedia:WikiProject Rocketry](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Rocketry)
- [Wikipedia:WikiProject Aircraft](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Aircraft)
- [Wikipedia:WikiProject Aviation](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Aviation)

pavel-karatsiuba:
Added availability to set --articleList value equal to comma-separated URL's with articles list with one title per line

fix #1729

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -38,7 +38,7 @@ import {
   saveStaticFiles,
   writeFilePromise,
   importPolyfillModules,
-  readFileOrUrlByLine,
+  extractArticleList,
   getTmpDirectory,
 } from './util/index.js'
 import S3 from './S3.js'
@@ -274,7 +274,7 @@ async function execute(argv: any) {
   let articleListToIgnoreLines: string[]
   if (articleListToIgnore) {
     try {
-      articleListToIgnoreLines = await readFileOrUrlByLine(articleListToIgnore)
+      articleListToIgnoreLines = await extractArticleList(articleListToIgnore)
       logger.info(`ArticleListToIgnore has [${articleListToIgnoreLines.length}] items`)
     } catch (err) {
       logger.error(`Failed to read articleListToIgnore from [${articleListToIgnore}]`, err)
@@ -285,7 +285,7 @@ async function execute(argv: any) {
   let articleListLines: string[]
   if (articleList) {
     try {
-      articleListLines = await readFileOrUrlByLine(articleList)
+      articleListLines = await extractArticleList(articleList)
       if (articleListToIgnore) {
         articleListLines = articleListLines.filter((title: string) => !articleListToIgnoreLines.includes(title))
       }
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -39,6 +39,7 @@ import {
   writeFilePromise,
   importPolyfillModules,
   readFileOrUrlByLine,
+  getTmpDirectory,
 } from './util/index.js'
 import S3 from './S3.js'
 import RedisStore from './RedisStore.js'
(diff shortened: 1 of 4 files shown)


src/mwoffliner.lib.ts
```diff
@@ -290,7 +291,7 @@ async function execute(argv: any) {
   let articleListLines: string[]
   if (articleList) {
     try {
-      articleListLines = await readFileOrUrlByLine(articleList)
+      articleListLines = await readFileOrUrlByLine(articleList, downloader.streamRequestOptions, tmpDirectory)
```
kelson42, line 294: Please don't add additional parameter only for technical reason:
* There is a global tmp directory defined, use it
* considering that the downloader is always the same, no reason to pass it as argument.

kelson42 (changes requested):
I see no reason to complexify call to `readFileOrUrlByLine()`

kelson42:
@pavel-karatsiuba Any news here?

src/mwoffliner.lib.ts
pavel-karatsiuba: I moved the `readFileOrUrlByLine()` function into `misc.ts` because, from the old place, it cannot be tested simply.
The temporary directory is not globally defined, it is defined in `mwoffliner.lib.ts`, and after executing the script all temp files are removed. That's why I added it as a parameter.
Also, tests use another tmp directory.

Same situation with the downloader. The Downloader object creates inside `execute()` function in the file `mwoffliner.lib.ts`. `streamRequestOptions` method depends on downloader params which are passed to constructor.

If I remove both parameters then I need to move the `readFileOrUrlByLine()` function back inside the `execute()` function in the file `mwoffliner.lib.ts`. After this, I can't test it because it will be function inside another function.

kelson42: Do the revamping work that all of this fits propely…. If you have to add useless parameters like this, that just mean you have a deeper code organisation problem. A problem you will have to fix anyway at some point, so just do it early.

kelson42 (changes requested):
From a high level perspective it looks a lot more sane now, still a few things to change IMO:
* (this was wrong in the past already) but we should not test URL with `includes('http')`, we should at least test if it starts with `http` and probably better check if there is a proper protocol sheme. Maybe using something https://www.npmjs.com/package/isurl is appropriate.. or redo a similar function ourself (and test it)
* Please do the split at the very top of `readFileOrUrlByLine()` and loop on it. That way it will be clearer and you will avoid the recursivity.
* Find a solution to avoid copying the Axios options (and avoid the cut/paste), maybe you have to move these option out of the Downloader class (if you can not export them)

src/util/misc.ts
```diff
@@ -424,56 +425,47 @@ export function cleanupAxiosError(err: AxiosError) {
...
+      .split(',')
+      .filter((n) => n)
+      .map(async (part) => {
+        let item: string | string[] = part.trim()
+        if (item.indexOf('http') === 0) {
+          try {
+            const url = new URL(item)
+            item = await downloadListByUrl(url.toString())
+          } catch (e) {
+            // URL is not valid continue processing
```
kelson42, line 454: Why a silent exception here? If a list can not be downloaded it should stops in a handled manner?

Issue #1797, opened 2023-02-27
Search result returns text with formatting code


Popolechien:
Tested both on Kiwix-server and Kiwix-android, not sure this is the right place but here we go:

Looked at Wikispecies ([2023-02](https://library.kiwix.org/viewer#wikispecies_en_all_maxi_2023-02/A/Main_Page)) and searched for _Terminalia chebula_. Because this is a plant species name in latin, it is displayed in italics.

I get the following results with the `<i> </i>` in the search bar:
- on Kiwix-server
![Capture d’écran 2023-02-27 à 10 20 04](https://user-images.githubusercontent.com/21675885/221526580-36077f48-628a-4d75-8bda-dcfa77f2b981.png)
- on Kiwix-android
![Screenshot_20230227-101807](https://user-images.githubusercontent.com/21675885/221526709-a1dd668c-9655-40fc-bbbe-57ba0950c6a1.png)

kelson42:
@Popolechien Indeed, probably a bug in MWoffliner. Actually we had this very symptom in the past already (still need to find the ticket), therefore looks almost like a regression to me.
@pavel-karatsiuba Can you please tell us what is going on here with this article?

pavel-karatsiuba:
By specification, the title tag must be text-only. But for some reason title contains tags.
I have added a patch to strip tags from the title.

kelson42:
@pavel-karatsiuba Make sense, I will check upstream if I can find a ticket.

kelson42:
At the source this is a bug in Mediawiki, reported here https://phabricator.wikimedia.org/T331120

Jaifroid:
The corresponding article in Wikipedia has the title of the article in italics, but the search entry isn't showing `<i>` either in Kiwix Serve or in Kiwix JS.

It's important not to strip out legitimate search characters. For example, `&lt` is a legitimate search term for a Wikipedia article, yielding an article on the `Less than` sign `<`. There is a Directory Entry with a `dirEntry.url` that is simply `&lt` (see screenshot). This mustn't be stripped. It is a correct and intentional search term on Wikipedia.

![image](https://user-images.githubusercontent.com/4304337/236662712-3949aa0e-9a36-4ec6-af59-3e8422500abe.png)

![image](https://user-images.githubusercontent.com/4304337/236662848-c48bfa78-b3a5-43b6-9183-224905d2cc58.png)

kelson42:
@Jaifroid Pretty sure this is a duplicate

[...]

PR #1801, opened 2023-03-01
Strip html tags from title

pavel-karatsiuba:
Strip HTML tags from the title to avoid HTML tags in the search results

fix: #1797

### DIFF
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -280,13 +280,13 @@ export function getMediaBase(url: string, escape: boolean) {
 }
 
 export function getStrippedTitleFromHtml(html: string) {
-  const doc = domino.createDocument(html)
-  const titleEl = doc.querySelector('title')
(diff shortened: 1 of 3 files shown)


kelson42 (changes requested):
Please create a unit test for this and test the special cases.

src/util/misc.ts
```diff
@@ -283,7 +283,7 @@ export function getStrippedTitleFromHtml(html: string) {
   const doc = domino.createDocument(html)
```
kelson42, line 283: I'm choqued by this line which is super CPU expensive. There is no way we coukd do that quicker?! I know this is not the topic of this PR, but stil...

kelson42 (changes requested):

kelson42 (approved):
Please secure trailing space and tabulations or anything like this are removed.

PR #1802, opened 2023-03-03
Strip html tags from title

pavel-karatsiuba:
Strip html tags with htmlentities from the title
Fix: #1797


kelson42:
@pavel-karatsiuba there is nothing to fix, these htmlentities can be really valid and you have no way to know.

kelson42:
@pavel-karatsiuba If there is maybe something to fix, this is removing the piece of code whihc removes nodes within the title node, because this is invalid.

Issue #1805, opened 2023-03-08
It should not be possible to create a ZIM with and empty Metadata "Description"


kelson42:
For the moment it seems to be possible to create a ZIM with empty "Description" ZIM metadata.

Please check as well other mandatory metadata https://wiki.openzim.org/wiki/Metadata

pavel-karatsiuba:
The metadata `Illustration_48x48@1` is not set.
I will make the function to check other mandatory fields.
To add `Illustration_48x48@1` into metadata I need a separate ticket or at least a description.

kelson42:
> The metadata `Illustration_48x48@1` is not set.
> 
> I will make the function to check other mandatory fields.
> 
> To add `Illustration_48x48@1` into metadata I need a separate ticket or at least a description.

Yes, please create one. Shoukd be made in place of the favicon.... but same content.

PR #1806, opened 2023-03-08
Print error and stop the script if mandatory Metadata key is not set

pavel-karatsiuba:
If the mandatory Metadata key is not set then print Error and stop the script

Fixes: #1805


kelson42:
@pavel-karatsiuba It should not be an exception. An exception is a unamanged unexcpected situation in a software. It should be handled properly. That means as early as possible and A proper error message should be given to the user so he understands what is missing.

kelson42 (changes requested):
I see many weaknesses with your implementation:
* The checking is done pretty late, it get all the page details downloaded before checking this (which is useless)
* I know this is not strictly the topic of this ticket... but have you remarked that if the `Description` comes from the Mediawiki, then nothing stops it to be more than 80 charaters?! I kind of expect that this works fine as well.
* If error, the MWoffliner exits with code `0` which is wrong. Remark that if I a give a wrong `--customZimDescription` on the command line then it fails with exit code `2` (which is correct)
* If there is an error then I see something like this (not only I would just prefer to stop pretty quickly but on the top of this the messages are misleading):
```
[info] [2023-03-12T13:56:24.324Z] Getting JSON from [https://bm.wikipedia.org/w/api.php?action=query&format=json&prop=redirects%7Crevisions%7Ccoordinates&rdlimit=max&rdnamespace=0&colimit=max&rawcontinue=true&generator=allpages&gapfilterredir=nonredirects&gaplimit=max&gapnamespace=0&gapcontinue=Interlingua]
[log] [2023-03-12T13:56:24.967Z] Got [407 / 407] articles chunk from namespace [0]
[info] [2023-03-12T13:56:24.967Z] Getting JSON from [https://bm.wikipedia.org/w/api.php?action=query&format=json&prop=redirects%7Crevisions%7Ccoordinates&rdlimit=max&rdnamespace=0&colimit=max&rawcontinue=true&generator=allpages&gapfilterredir=nonredirects&gaplimit=max&gapnamespace=0&gapcontinue=Nsoro]
[log] [2023-03-12T13:56:25.585Z] Got [413 / 820] articles chunk from namespace [0]
[log] [2023-03-12T13:56:25.586Z] A total of [1206] articles has been found in namespace [0]
[log] [2023-03-12T13:56:25.586Z] Got ArticleIDs in 2.09 seconds
[log] [2023-03-12T13:56:25.586Z] Total articles found in Redis: 1207
[warn] [2023-03-12T13:56:25.586Z] Couldn't find strings file for [bm], falling back to [en]
[log] [2023-03-12T13:56:25.587Z] Doing dump
[log] [2023-03-12T13:56:25.587Z] Writing zim to [/home/kelson/code/mwoffliner/out/wikipedia_bm_all_2023-03.zim]
[error] [2023-03-12T13:56:25.587Z] Metadata "{key}" is required
[log] [2023-03-12T13:56:25.587Z] Finished dump
[log] [2023-03-12T13:56:25.587Z] Closing HTTP agents...
[log] [2023-03-12T13:56:25.588Z] All dumping(s) finished with success.
[log] [2023-03-12T13:56:25.588Z] Flushing Redis DBs
[info] [2023-03-12T13:56:25.588Z] Finished running mwoffliner after [5s]
[log] [2023-03-12T13:56:25.588Z] Exiting with code [0]
[log] [2023-03-12T13:56:25.589Z] Deleting temporary directory [/dev/shm/mwoffliner-1678629383338]
```

You need to rethink the architecture of this. First you need to consider that ZIM metadata come AFAIK either from the mediawiki metadata (one short HTTP request at start) or from the command line. Considering this I could propose to:
* Isolate the string metadata constraints in a form of a const table (mandatory, max length, ...)
* Use this table in the sanitizing of command line arguments (when necessary)
* Isolate the metadata sanitizing in a dedicated module (`mwoffliner.lib.ts` is already quite big)
* Reuse the string metadata constraints (created earlier) to check if all the metadata are correct
* Exit in a similar manner if wrong input via command line or just wrong metadata.

pavel-karatsiuba:
@kelson42 If I add sanitizing the Description as we made it for the command line parameter, then many downloads will fail. Maybe I need to trim the Description to 80 characters?

kelson42:
@pavel-karatsiuba obviously you can... even if is is a super minor improvment.... if an improvment at all (strongly suspect Mediawiki to do it already)

kelson42:
@pavel-karatsiuba Any news here? Please rebase/fix conflict before doing anything else.

pavel-karatsiuba:
@kelson42 Ready to review. With the last change, I fixed the test because the new sanitizing rules do not allow Description higher than 80 characters.

kelson42 (changes requested):
> Use this table in the sanitizing of command line arguments (when necessary)

This is not complete, for example max size of `LongDescription` is not checked

src/util/metaData.ts
```diff
@@ -0,0 +1,39 @@
...
+const schema = {
+  type: 'object',
+  properties: {
+    Creator: { type: 'string', minLength: 1 },
+    Description: { type: 'string', maxLength: 80, minLength: 1 },
+    Language: { type: 'string', minLength: 1 },
+    Publisher: { type: 'string', minLength: 1 },
+    Title: { type: 'string', maxLength: 30, minLength: 1 },
+  },
+  required: ['Creator', 'Description', 'Language', 'Publisher', 'Title'],
```
kelson42, line 15: Should be handled in properties as a `boolean`

src/util/metaData.ts
```diff
@@ -0,0 +1,39 @@
...
+  additionalProperties: true,
+}
+
+const validate = ajv.compile(schema)
+
+export const validateMetadata = (metaData) => {
+  const valid = validate(metaData)
```
kelson42, line 22: I think wording is `sanitize` not `validate`, not great to create inconsistency in the vocabulary.

test/unit/util.test.ts
```diff
@@ -412,4 +413,49 @@ describe('Utils', () => {
       await expect(extractArticleList('http://valid-wrong-url.com/')).rejects.toThrow('Failed to read articleList from URL: http://valid-wrong-url.com/')
     })
   })
+
+  describe('metaData', () => {
```
kelson42, line 417: Your automated tests don't really test the edge cases with undefined, wrong type, missing values.

kelson42 (changes requested):
Still work left... See https://github.com/openzim/zim-tools/pull/339/files, this is the kind of result I  was expected here as well.

src/util/metaData.ts
pavel-karatsiuba: In this case, I used the Ajv JSON schema validator. For this library, the required fields should be passed as an array.

src/util/metaData.ts
pavel-karatsiuba: In this case, we only validate the data with some rules but sanitizing implies that data will be modified with some rules. So I propose not renaming the function.

src/util/metaData.ts
kelson42: OK... Not sure this is the right tooling... but let deal with it.

kelson42 (approved):
LGTM

Issue #1808, opened 2023-03-09
Add Illustration_48x48@1 Metadata keyword


pavel-karatsiuba:
The Illustration_48x48@1 Metadata keyword is mandatory for ZIM file (https://wiki.openzim.org/wiki/Metadata) but it is not added by mwoffliner.

PR #1809, opened 2023-03-10
Added Metadata Illustration_48x48@1

pavel-karatsiuba:
Mandatory Metada keyword `Illustration_48x48@1` was added.

Fix: #1808


kelson42:
The specification specificaly requires to not create the favicon entry anymore... so why this PR does not remove this?

pavel-karatsiuba:
If I remove the favicon then tests will not pass. zimcheck shows the error: `Favicon is missing`

kelson42:
@pavel-karatsiuba Good point, please open a ticket in zimm-tools. We will need to fix and release quickly a new zim-tools.

Issue #1536, opened 2021-09-05
Double image description


Inbefortus:
ZIM: 2021-08 German Wikipedia

There are two image descriptions, although there should be only one (this is not the case everywhere).

http://library.kiwix.org/wikipedia_de_all_maxi/A/Belgien
http://library.kiwix.org/wikipedia_de_all_maxi/A/Nationalpark_Eifel

![Screenshot_20210905-093214_Samsung Internet](https://user-images.githubusercontent.com/71934042/132119410-5c07c525-8233-4d7e-ac0f-dfd7a9229207.jpg)
![Screenshot_20210905-093145_Samsung Internet](https://user-images.githubusercontent.com/71934042/132119412-562e9bdb-17f8-4d55-8aa2-6af738806cef.jpg)
![Screenshot_20210905-094031_Samsung Internet](https://user-images.githubusercontent.com/71934042/132119413-d5a637bb-7d9f-4410-ba9e-ed2ead70c84b.jpg)

kelson42:
Interesting, probably a bug in MWoffliner linked to some kind of change in HTML output.

kelson42:
@pavel-karatsiuba Can you please explain and fix this edge scenario?

kelson42:
This bug was related to `mobile-section` which has been deprecated (by MediaWiki and MWoffliner) AND the problem does not occur anymore in our latest ZIM file (done with `ActionParse` API end-point). Therefore closing because outdated and not occuring anymore.

Issue #1664, opened 2022-10-13
Using Wikipedia APIs mobile-sections endpoint is deprecated


johngian:
Related to: https://phabricator.wikimedia.org/T319303

It looks like kiwix is using the `mobile-sections` endpoint served by Wikipedia REST API:
https://en.wikipedia.org/api/rest_v1/#/Mobile

This endpoint is considered deprecated and we are working towards sunsetting it.
Instead of `mobile-sections` its suggested to use the page content for mobile and specifically:
https://en.wikipedia.org/api/rest_v1/#/Page%20content/getContent-mobile-html

kelson42:
@johngian Good to know, nee to read more about the rationals behind this. Hopefully the new end-points are not too different.

kelson42:
@johngian It seems the old is deprecated, but I wonder that you recommend to use a new API end-point which is effectively tagged as « experimental » and therefore not productionreqdy!? See https://de.wikipedia.org/api/rest_v1/#/Page%20content/getContent-mobile-html

Jaifroid:
Given some outdated information as reported in #1397 is caused by using the deprecated mobile-sections API, and that this is only going to get more urgent, we need to work out what is needed to make this switch. I don't know these APIs well, but from what I can see, the mobile-sections endpoint produces JSON or something similar, whereas mobile-html produces (evidently) html.

Memory383:
> this is only going to get more urgent.

**Using the deprecated mobile-sections API is a really severe issue.** It is really frustrating for users and editors.

_For example:_

 I have added more than 500 photos in the German Wiktionary this year. Not a single of those photos appear on the .zim version of the Wiktionary. Also, many translations and definitions are missing. _**Every month the issue is worse !**

By the way, @johngian thanks for opening this issue on GitHub. I am the author of the "Bug report" on Phabricator.

kelson42:
for the moment I'm a bit puzzled:
* new endpoint is still experimental
* references link don't work
* I see strange edit icon at the right of each section title

It seems simply not rreally ready to me.

Jaifroid:
@kelson42 It looks like mwoffliner can get the scripts and styles needed for the mobile view by using this part of the API:

https://en.wikipedia.org/api/rest_v1/#/Page%20content/get_page_mobile_html_offline_resources__title___revision_

Issue #1544, opened 2021-09-24
Map location markers are wildly inaccurate in (at least) Wikivoyage


Jaifroid:
Please compare the red dot marking the location of Cambridge in the three renderings below. On the left is online German Wikivoyage (correct). In the middle is the rendering of the map in Kiwix Desktop. On the right is the rendering of the map in Kiwix JS. The fact that the two Kiwix renderings are so wrong (Cambridge appears to be located in south London!) suggests an issue in the ZIM, not in the readers.

![image](https://user-images.githubusercontent.com/4304337/134625699-bb59c1c4-ba8f-45eb-8f33-74ea5c4a5cc2.png)

Source: `wikivoyage_de_all_maxi_2021-09.zim`

Jaifroid:
I think I've traced this error to the inclusion of an image caption beneath the image (in image above, you can see that the image caption is not present in the original online version). Examining this in dev tools, we can see the problem:

![image](https://user-images.githubusercontent.com/4304337/134629377-11591e14-24af-4f70-bfd2-d2f039bd1f6c.png)

The table within which the marker is absolutely positioned is increased in height from 165x200 to 167x244.1. The extra 44 pixels is the displacement. If I delete the caption div from the DOM in Kiwix JS, the dot is then repositioned normally:

![image](https://user-images.githubusercontent.com/4304337/134629796-f5c81a90-da87-481a-9ba0-23a13b4f3a09.png)

kelson42:
@Jaifroid Thx for tbe bug report and analysis. I will try to confirm and fix.

kelson42:
Looks indeed bad on Wikivoyage:
![image](https://user-images.githubusercontent.com/1029718/134809023-35895e72-3b43-48a0-8978-5ed8f2d53a68.png)

But interstingly, looks good on Wikipedia:
![image](https://user-images.githubusercontent.com/1029718/134809048-f44182fb-3ed8-426c-827c-84587837db67.png)

kelson42:
[...]

PR #1816, opened 2023-03-23
Don't write image descriptions twice

pavel-karatsiuba:
If the image contains the same description in the nearest DOM element then do not add an image description because it will be displayed twice.

Fix: #1536

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -773,6 +773,16 @@ function makeThumbDiv(dump: Dump, parsoidDoc: DominoElement, imageNode: DominoEl
   return thumbDiv
 }
 
+function getInnerTextFromElement(domElement: DominoElement) {
+  const innerNodes = domElement.childNodes
+  for (const i in innerNodes) {
+    if (innerNodes[i].nodeType === domElement.TEXT_NODE) {
+      return innerNodes[i].textContent.trim()
+    }
+  }
+  return null
+}
+
 function treatImageFrames(mw: MediaWiki, dump: Dump, parsoidDoc: DominoElement, imageNode: DominoElement) {
   const image = imageNode.getElementsByTagName('img')[0] || imageNode.getElementsByTagName('video')[0]
 
@@ -795,7 +805,7 @@ function treatImageFrames(mw: MediaWiki, dump: Dump, parsoidDoc: DominoElement, 
   thumbcaptionDiv.setAttribute('class', 'thumbcaption')
   const autoAlign = dump.mwMetaData.textDir === 'ltr' ? 'left' : 'right'
   thumbcaptionDiv.setAttribute('style', `text-align: ${autoAlign}`)
-  if (description) {
+  if (description && getInnerTextFromElement(imageNode.parentNode) !== getInnerTextFromElement(description)) {
     thumbcaptionDiv.innerHTML = description.innerHTML
   }
 


kelson42:
@pavel-karatsiuba in a first place, before codin, understanding of the problem. why a few images suffer of this problem and other don't? I guess the bug ia not in the original dom?

pavel-karatsiuba:
On the original Wikipedia page, I see other tags and it does not contains the description for the image, but only for the block.

The easiest way is not to use the `figcaption` tag as a description. But maybe it is used for other pages.
My fix is testing if we have the same text from the `figcaption` tag and from `td` tag.

This is the HTML from API:
```
<td colspan="2" style="margin:5px 0;text-align:center">
    <figure class="mw-halign-center">
        <a>...</a>
        <figcaption>Das Gebiet des Nationalparks</figcaption>
    </figure>
    Das Gebiet des Nationalparks
</td>
```

kelson42:
@pavel-karatsiuba Does our parser is adapted at all to parse the "figure" DOM element? Have you read https://www.mediawiki.org/wiki/Parsoid/Parser_Unification/Media_structure/FAQ/en ?

pavel-karatsiuba:
I have read this document and see that our code is fit to described format.
But the problem is the same: we are getting two equal texts.
The current fix is not acceptable because we need to use text from `figcaption` tag.

So I propose to remove the text which is placed after the `figure` tag.

kelson42:
@pavel-karatsiuba How do you explain the text is not displayed in duplicate on Wikipedia itself? The problem behind this is big.

pavel-karatsiuba:
@kelson42 I discovered more about this issue and found the ticked with almost the same problem: #1544. In this ticket, I see the link to the upstream ticket, which is opened.

kelson42:
I understand, please explain.

pavel-karatsiuba:
https://de.wikipedia.org/api/rest_v1/page/html/Nationalpark_Eifel
On the REST desktop page source code I see also two descriptions. But in Browser I don't see a second description because it is hidden by a simple CSS rule: 
```
figure[typeof~='mw:File'] > figcaption {
    display: none;
}
```
mwoffliner also has the same CSS rule. But mwoffliner uses the [mobile version](https://de.wikipedia.org/api/rest_v1/page/mobile-sections/Nationalpark_Eifel) and API for the mobile version does not provide the `typeof` attributes.

To see the difference, you can compare the next pages in the browser:
https://de.wikipedia.org/api/rest_v1/page/html/Nationalpark_Eifel
https://de.wikipedia.org/api/rest_v1/page/mobile-html/Nationalpark_Eifel

kelson42:
@pavel-karatsiuba Why you refer to https://de.wikipedia.org/api/rest_v1/page/mobile-sections/Nationalpark_Eifel ?! AFAIK MWoffliner does not use this API (this is the topic of https://github.com/openzim/mwoffliner/issues/1664)!

pavel-karatsiuba:
Sorry, I didn't know about the old and new versions of API. I thought that [mobile-html](https://de.wikipedia.org/api/rest_v1/page/mobile-html/Nationalpark_Eifel) and [mobile-sections](https://de.wikipedia.org/api/rest_v1/page/mobile-sections/Nationalpark_Eifel) return the same data but with a different view.

This is the ticket that described the same situation: https://phabricator.wikimedia.org/T291779
I think this is a duplicate ticket. The problem described in this ticket is the same as in this one: https://github.com/openzim/mwoffliner/issues/1544. Both tickets have problem with description which should be hidden. But we can't hide it simply because for `mobile-section` we did not get `typeof` attribute.

My PR is fixing only the current ticket.

kelson42:
Closing, we don't use this end-point anymore.

PR #1821, opened 2023-03-29
added test to check the right scraped multimedia files

### LINKED ISSUE #890: Multiple problems in scraping of multimedia content
I have created a small test about multiple type of content and ways to include them, but everything is standard. It is available here https://en.m.wikipedia.org/wiki/User:Kelson/MWoffliner_CI_reference.

I have scarped it with `1.9.4` and this was a bit disappointing. We have a here many problems, most of them being that the content is simply not made available. I think such a page should be really tested properly to secure that we don't have anymore big problem around multimedia content displaying.

pavel-karatsiuba:
The test is scraping the page with different types of multimedia files.

Fixes: #890

### DIFF
--- a/src/util/rewriteUrls.ts
+++ b/src/util/rewriteUrls.ts
@@ -97,9 +97,11 @@ function rewriteUrlNoArticleCheck(articleId: string, mw: MediaWiki, dump: Dump, 
       lat = parts[4]
       lon = parts[5]
     } else if (rel === 'mw:MediaLink') {
-      const shouldScrape = (href.includes('.pdf') && !dump.nopdf) || ((href.includes('.ogg') || href.includes('.oga')) && !dump.nopic && !dump.novid && !dump.nodet)
+      const scrapePdf = href.includes('.pdf') && !dump.nopdf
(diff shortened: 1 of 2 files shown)


kelson42 (changes requested):
* What about the different format testing?
* Are you sure User:Kelson/MWoffliner_CI_reference is actually scraped properly?

kelson42:
No feedback here :(

Actually it's not. This is what I get on the first section of the upstream
image

Here is what I get with this branch
image

Here the remarks:
* The two visible images seems OK. They are blurred, but I guess this is because of the quality loss the recompression brings with.
* The first of the 3 links is OK as well as it points to a description page which is not existing in the ZIM
* The two other/last links are wrong. They point to an external link, but they should point to the version embedded in the ZIM (and it should be embedded in the ZIM). If they would be an external link, then should be handled link an external link (with the appropriate CSS icon). But here we should actually include the file and point to it.

pavel-karatsiuba:
On the bottom of the zim page, I see also external links to the `.webm` file and the `.mp3` file.
Do I need to save `.webm` and `.mp3` into zim file or add an icon to show the 'external' file?

kelson42:
@pavel-karatsiuba none of these links are external. An external links has a precise meaning in Mediawiki, see https://www.mediawiki.org/wiki/Help:Links#External_links

Issue #1818, opened 2023-03-26
Duplicate in cmd line arguments are not checked (was: --verbose does not accept empty value)


kelson42:
```
$ mw --mwUrl="Https://de.wikipedia.org/" --verbose

....

**********

verbose should be empty or one of [info, log, warn, error, quiet].

**********

```

pavel-karatsiuba:
@kelson42 Excidently found out that you will get such an error if you use `--verbose` parameter more than one time.

Example to reproduce the error:
`$ mw --verbose=info --adminEmail="contact@kiwix.org" --mwUrl="Https://de.wikipedia.org/" --verbose`

kelson42:
No way to limit tp one with yargs?

kelson42:
> @kelson42 Excidently found out that you will get such an error if you use `--verbose` parameter more than one time.
> 
> Example to reproduce the error: `$ mw --verbose=info --adminEmail="contact@kiwix.org" --mwUrl="Https://de.wikipedia.org/" --verbose`

OK, was definitly to stupid to get that error on my side. Sorry for the misleading ticket/PR.

PR #1823, opened 2023-04-01
Stop script and throw error if used same cli parameter twice

pavel-karatsiuba:
When the user accidentally passes to the command line the same parameter more than one time then he can get unexpected scraper behavior.

This PR is checking parameters and if one parameter is used more than one time then the script stops and an error throws.

fix: #1818

### DIFF
--- a/src/sanitize-argument.ts
+++ b/src/sanitize-argument.ts
@@ -10,6 +10,7 @@ import * as QueryStringParser from 'querystring'
 import { isValidEmail } from './util/index.js'
 import * as path from 'path'
 import { fileURLToPath } from 'url'
+import { parameterDescriptions } from './parameterList.js'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
@@ -17,6 +18,8 @@ const __dirname = path.dirname(__filename)
 export async function sanitize_all(argv: any) {
   // extracting all arguments
   const { articleList, addNamespaces, speed: _speed, adminEmail, mwUrl, customZimFavicon, optimisationCacheUrl, verbose, customZimLongDescription, customZimDescription } = argv
+
+  sanitizeDoubleUsedParameters(argv)
 
   sanitize_articlesList_addNamespaces(articleList, addNamespaces)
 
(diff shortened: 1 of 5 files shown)


pavel-karatsiuba:
Attention: the yargs parser can take more than one same parameter. It is processed such parameters as an Array of values. So Some of mwoffline parameters can be passed more than one time. These are: 'articleList', 'articleListToIgnore', 'addNamespaces'

kelson42:
@pavel-karatsiuba Why no review is requested here? Are you over?

But I'm not happy with this PR for many reasons:
* If there is [an architecture decision to take](https://github.com/openzim/mwoffliner/issues/1818#issuecomment-1488504883), I expect you to wait this is validated. Here you have taken on you to chose solution (1) which is IMO the wrong one. Why you don't have taken solution (2), which is the builtin solution (reading https://github.com/yargs/yargs/issues/1318 offers many ways)?
* Where have you read that any of these arguments `--articleList, --articleListToIgnore, --addNamespaces` allow multiple occurences? Are you mixing "multiple instances" with "comma separated values"?!
* AFAIK, there is only one argument which is legitim to be duplicated and this is `--format`... and this is written in the `usage()`: `Each --format argument....`
* In the test, why using `execa` which is super high level, expensive, etc... instead of the testing `yargs` with the right configuration?
* Error message `Parameter ... can't take value "["https://bm.wikipedia.org","https://bm.wikipedia.org"]` is cryptic for a non-tech user. You should write `Parameter ... can only be used once`.

pavel-karatsiuba:
- #1818 here you are using `mw` command. In the other PR I saw that `mw` is equal to `./node_modules/.bin/ts-node-esm ./src/cli.ts --adminEmail=kelson@kiwix.org --osTmpDir=/dev/shm --verbose --optimisationCacheUrl=http://s3.us-west-1.wasabisys.com/?bucketName=org-kiwix-dev-mwoffliner&keyId=6JE56A5IWY5Z3H3L3KH6&secretAccessKey=...`. So the problem is not about `--verbose` but about using `--verbose` twice.
- Those decisions are equal. But if I use `duplicate-arguments-array` option then doubled parameter will be silently ignored and used last from the command. My decision is to show the error if you use the same parameter twice.
- You are right. Only `format` parameter can be used more than one time. Because of this, we can't use `duplicate-arguments-array` option.
- Yes, such tests are not acceptable.
- I will change the message.

pavel-karatsiuba:
`cli.ts` is written in such a manner that it can't be tested partially. So before changing tests, the `cli.ts` should be redesigned. It should be split by a couple of files and inside cli.ts should be only executed one function.

kelson42:
@pavel-karatsiuba
* PR is still not assigned nor requested for review... I have basically no clue what is current status.
* Why not using the `[check()](https://github.com/yargs/yargs/issues/1318#issuecomment-494859667)`?
* Unit test `sanitizeDoubleUsedParameters`function not `cli.ts`
* Test the case which should not generate an error which is with many `--format`

pavel-karatsiuba:
The test was changed.

If I use `.check()` then we did not get any benefits. But we will get another format of the error message.
We can use `.check()` to sanitize all parameters, then the format of the error will be the same for all sanitizing errors. But this should be another ticket.

Issue #1825, opened 2023-04-04
--format parameter is not working properly


pavel-karatsiuba:
The command 
`npm run mwoffliner -- --verbose=info --adminEmail="contact@kiwix.org" --articleList="User:Kelson/MWoffliner_CI_reference" --mwUrl="https://en.m.wikipedia.org/" --format=nopic --format=nopdf --format=novid` is building 3 zim files.
The 'nopic' zim file does not contain any pictures and video, but 'nopdf' contains a pdf file, and 'novid' contains a video. 
I expected that 'nopdf' will be without any pdf and 'novid' will be without video files.

Also if I mix the order of the parameters then I take another result. But this result will be also wrong.

PR #1826, opened 2023-04-04
Clearing Redis media storage after each dump

pavel-karatsiuba:
Clearing Redis media storage after each dump

fixes: #1825

### DIFF
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -45,7 +45,7 @@ export async function getAndProcessStylesheets(downloader: Downloader, redisStor
         while ((match = cssUrlRegexp.exec(body))) {
           let url = match[1]
 
-          /* Avoid 'data', so no url dependency */
+          /* Avoid 'data', so no URL dependency */
           if (!url.match('^data')) {
             const filePathname = urlParser.parse(url, false, true).pathname
             if (filePathname) {
--- a/src/RedisStore.ts
+++ b/src/RedisStore.ts
@@ -8,7 +8,6 @@ class RedisStore implements RS {
   private storesReady: boolean
 
   private _filesToDownloadXPath: RKVS<FileDetail>
-  private _mediaToDownloadXPath: RKVS<FileDetail>
   private _filesToRetryXPath: RKVS<FileDetail>
   private _articleDetailXId: RKVS<ArticleDetail>
   private _redirectsXId: RKVS<ArticleRedirect>
@@ -59,15 +58,8 @@ class RedisStore implements RS {
     }
   }
 
-  public async flushMediaToDownloadXPath() {
-    if (this._client.isReady && this.storesReady) {
-      logger.log('Flushing Redis DB for storing media')
-      await this._mediaToDownloadXPath.flush()
-    }
-  }
-
   public async checkForExistingStores() {
-    const patterns = ['*-media', '*-files', '*-media-retry', '*-detail', '*-redirect']
+    const patterns = ['*-media', '*-media-retry', '*-detail', '*-redirect']
     let keys: string[] = []
     for (const pattern of patterns) {
       keys = keys.concat(await this._client.keys(pattern))
@@ -85,13 +77,7 @@ class RedisStore implements RS {
   }
 
   private async populateStores() {
-    this._mediaToDownloadXPath = new RedisKvs(this._client, `${Date.now()}-media`, {
-      u: 'url',
-      n: 'namespace',
-      m: 'mult',
-      w: 'width',
-    })
-    this._filesToDownloadXPath = new RedisKvs(this._client, `${Date.now()}-files`, {
+    this._filesToDownloadXPath = new RedisKvs(this._client, `${Date.now()}-media`, {
       u: 'url',
       n: 'namespace',
       m: 'mult',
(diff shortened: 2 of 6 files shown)


kelson42:
@pavel-karatsiuba please explain clearly the nature of the problem and the nature of the solution andbits consequences

pavel-karatsiuba:
All paths to files are stored in the Redis store before downloading. Then the script is reading file paths one by one from Redis and downloads them. If we use a parameter `--format` then we don't save to Redis some type of files. So they are filtered before being saved to the Redis.
If we don't use `--format` parameter or if we use one value for the parameter `--format` then all works well.

In the case if we use two '--format' parameters then all media files are filtered with the first '--format' parameter. And paths for filtered media will be stored in Redis storage. Before using the next '--format' parameter we are not clearing Redis storage and all media data from the first iteration will be used also for the second iteration.

For example, the page contains pdf and jpg files. We are using a script with parameters `--format=nopic --format=nopdf`. In the first iteration, the script will filter the jpg file and save into a Redis store only the path to pdf. Then for the next iteration script will save in the Redis store only the path to the jpg file, this is right, but in the Redis, we already have a pdf from the first iteration. So as result in the nopdf ZIM file we will see jpg and pdf files.

My fix is clearing the Redis store after each dump. And I used an additional Redis store to separate media files and other files for example CSS. So my solution removes only paths to media files but does not remove paths to CSS.

pavel-karatsiuba:
Why we I need a new Redis store:
The scraper stores all paths to files in the Redis store. And then reads paths from this store and downloads files. In this store, the script is saving paths to all media files like ogg, png, pdf, and plus CSS. This is a good solution if we use the command without the `--format` parameter or if we use one time `--format` parameter.
But if we use more than one time `--format` then we need to remove all paths to media files after each iteration. But we don't need to remove paths to CSS. That's why I am using one store for media files and another store for other files. After each iteration, I am clearing the store with paths to media files and did not touch the store for other files.

kelson42:
@pavel-karatsiuba Thank you for your explanations, the problem is clear now to me and the solution seems appropriate. This is a major bug and thank you very much for detecting and finxing it.

One point is still unclear to me. Populating the medias happens based on the parsing of HTML, so if we remove everything, then the CSS should be repopulated?! I also don't understand why you called somethin `media` when it seems it was only for CSS?!

kelson42 (commented):
Still few points to clarify/fix.

pavel-karatsiuba:
For storing CSS I used storage with the "Files" prefix. For media files, I used storage with the "Media" prefix.
I also found that subtitles also should be stored with media files because the treatment of the media files also takes effect on subtitles. I made a commit with fixes.

pavel-karatsiuba:
The first thing I made is clearing the Redis store which contains media files and CSS after finishing the current iteration. On the next iteration media files were repopulated but not CSS. In the source code, I see a condition that is not allowing to do this: https://github.com/openzim/mwoffliner/blob/main/src/util/dump.ts#L61
That's why I made another storage for Media files and I am clearing it after each iteration independent from Files storage.

Subtitles are files in `.vtt` format. They are downloading near media files and can be skipped by the `--format=novid` parameter.

kelson42:
> Subtitles are files in `.vtt` format. They are downloading near media files and can be skipped by the `--format=novid` parameter.

Now I remember, subtitles need clearly to be removed after each dump indeed.

src/util/dump.ts
```diff
@@ -58,8 +58,7 @@ export async function getAndProcessStylesheets(downloader: Downloader, redisStor
               url = getFullUrl(url, cssUrl)
               url = url.indexOf('%') < 0 ? encodeURI(url) : url
 
-              /* Download CSS dependency, but avoid duplicate calls */
-              if (!downloader.cssDependenceUrls.hasOwnProperty(url) && filename) {
+              if (filename) {
                 downloader.cssDependenceUrls[url] = true
```
kelson42, line 62: I think this is wrong… there is no reason why we should allow to download twice the same url… but no time to make the full investigation. Not sure what this `cssDependenciesUrls` is exactly about… but either we should oeep it working like it was or this should be cleaned as well between to dumps.

kelson42: What about javascript dependencies?

src/util/dump.ts
```diff
@@ -58,8 +58,7 @@ export async function getAndProcessStylesheets(downloader: Downloader, redisStor
               url = getFullUrl(url, cssUrl)
               url = url.indexOf('%') < 0 ? encodeURI(url) : url
```
kelson42, line 59: Really wonder about that line: it should be either the opposite… or we shoukd encode all the time.

kelson42 (commented):
Few comments

src/util/dump.ts
pavel-karatsiuba: This PR is clearing Redis store after creating zim file. With one command the script can create more than one zim file only if you put more than one `--format` parameter. The `--format` parameter takes effect only on media files. So other files will be the same for each zim file. That's why we don't need to repopulate js files. They should be downloaded one time and used for each zim file.

src/util/dump.ts
pavel-karatsiuba: This row encodes the URL if it is not encoded. It does not allow encoding already encoded URL.

PR #1838, opened 2023-05-11
Move to libzim9

kelvinhammond:


kelson42:
@kelvinhammond Thank you again for being back on this super-important PR. Just wanted to ask what is the status? Do you think this is ready to review? Do we just need to release latest node-libzim to get the CI passing?

kelvinhammond:
CI build fails but looks good to review to me @kelson42

kelson42:
@kelvinhammond I had a look and there a bit of work relatwd to links and content. I will work on this on your PR and come back to you.

kelson42 (changes requested):
I had a look to the PR. Here are my comments:
* MWoffliner builds a ZIM file, which is a good start
* The ZIM file built is incorrect by many way, you can see it easily if you open it with `kiwix-serve`
* One problem is that new ZIM flies (using latest libzim) do handle the ZIM namespaces differently. They were before open and free to use and now they are hidden behind the scene in the libzim. We did that to basically how the writting of content for the ZIM developer in the `C` namespace. This means that everything should be in the `C` namespace now (transparent for MWoffliner) whereas in the past we have put the images in the `I` namespace and the article in the `A` namespace. All of this has a few implication how we writte the content and how we create the links in the HTML. I could handle this kind of problem.
* The other problem is that it seems the libzim creates corrupted content. I have tried with the "Animation" article in the Wikipedia in English. It seems that many images are corrupted, for example `FelixTheCat-1919-FelineFollies_silent.ogv.180p.vp9.webm` which is now not a valid webm file. I suspect the reason is that you write anything through the `StringItem` which is IMO not appropriate for blobs.

IMO we should really fix the problem with blobs in priority.

kelson42:
We should now rebase on latest HEAD now that `main` branch is ready for this move and now that node-libzim has been fixed

kelson42:
@audiodude Please have a look to this PR, this has been done before the rearchitecturing on fall 2023, but otherwise most of the code should be OK and just to be taken.

benoit74:
Superseded by https://github.com/openzim/mwoffliner/pull/2194

PR #1867, opened 2023-07-24
test commit for multimedia content test

VadimKovalenkoSNF:
Test multimedia tests with Github actions because they rely on Zimcheck.


VadimKovalenkoSNF:
This PR is no longer relevant.

PR #1865, opened 2023-07-21
AWS SDK jsv3 migration

VadimKovalenkoSNF:
Superseeds #1851
Fixes #1840

1. Migrate S3 module to aws sdk for js v3.
2. Add proper error handling.
3. Refactor S3 unit tests.

PR #1870, opened 2023-07-28
Improve Codecov configuration

kelson42:


.codecov.yml
```diff
@@ -1,7 +1,16 @@
...
+    require_ci_to_pass: yes
+
 coverage:
   status:
     project:
       default:
-        threshold: 0.1%
-    patch: false
-    changes: false
+        threshold: 1%
```
VadimKovalenkoSNF, line 9: Increasing the threshold is a workaround for me. Please take a look at my comment [here](https://github.com/openzim/mwoffliner/pull/1865#issuecomment-1657694463).

kelson42: Definitly, it's just a configuration which has proven to work for an other repository. It seems to me more appropriate, but it does not fix the problem/questioning we have with #1865

.codecov.yml
```diff
@@ -1,7 +1,16 @@
...
+    patch:
+      default:
+        target: 90%
```
VadimKovalenkoSNF, line 12: 90% looks pretty strict to me. We can use this value for now and see how it will affect MCS decommission and other changes related to migration/updation things.

VadimKovalenkoSNF (approved):

Issue #1357, opened 2021-01-02
Remove MWCapabilities


kelson42:
`MWCapabilities` is a dedicated class instanciated within `Downloader`. This is not the right place. This should be available in the `Mediawiki` class.

I want to have these 4 method, implemented as lazy in `Mediawiki`:
* `hasVeApi()`
* `hasCoordinatesApi()`
* `hasDesktopRestApi()`
* `hasMobileRestApi()`

Probably the a lazy method should be as well implemented to get the MainPage:
* `getMainPage()`

Remark: This is a necessary step to re-architecture the Mediawiki-Downloader-Renderer

kelson42:
@uriesk @pavel-karatsiuba IMO one of girst things to do in 1.13.0 before implementing support of any new backend. Do you agree? Have other comments?

uriesk:
I agree.
And currently we `getArticleUrl` by appending the articleId to a baseUrl (which got choosen based on the Capabilities).
But not all APIs will always allow us to do that.
I would like a concept similar to this:

```js
const apis = {
  'ACTION': {
    getArticleUrl: (articleId) => {
      return `${baseUrl.href}action=visualeditor&mobileformat=html&format=json&paction=parse&page=${encodeURIComponent(articleId)}`;
    },
  },
  'RESTV1': {
    getArticleUrl: (articleId) => {
      return `${baseUrl.href}/api/rest_v1/${encodeURIComponent(articleId)}/html`;
    },
  },
}

const choosenEndpoint = apis['ACTION'];

const url = choosenEndpoint.getArticleUrl('Article_Title');

```

kelson42:
@uriesk I think we are on the same wave length! Before building any new API endpoint support, we need first to reorganise the current modules in a way content endpoints calls an response treatment are modularised, ie. share AFAP a common interface/API for the rest of MWoffliner.

That way it will be easier to understand an test. 

I want to open a dedicated ticket for that for a long time but actually I'm not sure how the fix should exactly look like... so I have open only this small one which was more obvious to me.

kelson42:
@pavel-karatsiuba Here we will a rearchitecturing proposal from you.

pavel-karatsiuba:
@kelson42 
Your proposed changes are a good and easy step to start changing architecture.
If we start from these changes then makes sense to decompose whall Downloader class and leave here only the methods needed to download content. And other methods need to be moved to other places.
[...]

PR #1839, opened 2023-05-11
Rearchitecture MWOffliner HTML/CSS/JS scraping (part #1)

DonAlexandro:
## Github issue
Fixes partly https://github.com/openzim/mwoffliner/issues/1830

#### 16.05.2023
1. Moved article treatment methods from saveArticle.ts to article.treatment class
2. Refactored the code with the new class
3. Fixed unit tests for the new class
4. Moved media treatment methods from saveArticle.ts to media.treatment class
5. Refactored the code with the media.treatment class
6. Fixed unit tests for the media.treatment class

#### 15.05.2023
1. Completely finished refactoring pieces of code dedicated to specific API endpoints
2. Fixed all existing unit tests
3. Fixed and added a couple of new ones for directors 

#### 12.05.2023
1. Finished unit test
4. Almost finished replacing old code with new directors

#### 11.05.2023
1. Created directors for all URLs
2. Updated an initial implementation of URL builder
5. Studied how to run unit tests on the project and started  writing some for builder and its directors

#### 21.08.2023
1. Remove mw capabilities for desktop and visual rendering in favor of respective lazy methods in MediaWiki class (fixes https://github.com/openzim/mwoffliner/issues/1357).
2. Convert MediaWiki class into the singleton.
3. Create renderer builder class to handle rendering between different kind of API (such as Wikimedia Desktop, VisualEditor and placeholder for WikimediaMobile).
4. Refactored unit tests to handle renderers and MediaWiki singleton.

PR #1878, opened 2023-08-04
Remove MWCapabilities object (partial impl)

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1357

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -11,26 +11,126 @@ import basicURLDirector from './util/builders/url/basic.director.js'
 import BaseURLDirector from './util/builders/url/base.director.js'
 import ApiURLDirector from './util/builders/url/api.director.js'
 
+interface MWCapabilities {
+  apiAvailable: boolean
+  veApiAvailable: boolean
+  coordinatesAvailable: boolean
+  desktopRestApiAvailable: boolean
+}
 class MediaWiki {
   public metaData: MWMetaData
   public readonly baseUrl: URL
-  public readonly modulePath: string
-  public readonly webUrl: URL
-  public readonly apiUrl: URL
-  public readonly veApiUrl: URL
-  public readonly restApiUrl: URL
-  public readonly mobileRestApiUrl: URL
-  public readonly desktopRestApiUrl: URL
+  public readonly modulePathConfig
   public readonly getCategories: boolean
   public readonly namespaces: MWNamespaces = {}
   public readonly namespacesToMirror: string[] = []
 
   private readonly wikiPath: string
+  private readonly restApiPath: string
   private readonly username: string
   private readonly password: string
   private readonly apiPath: string
   private readonly domain: string
   private apiUrlDirector: ApiURLDirector
+  private baseUrlDirector: BaseURLDirector
+
+  private _veapiUrl: URL
+  private _restApiUrl: URL
+  private _apiUrl: URL
+  private _modulePath: string
+  private _webUrl: URL
+  private _desktopRestApiUrl: URL
+  // TODO: Mobile builder was removed since there is no /mobile-sections endpoint
+
+  // Set default MW capabilities
+  private readonly mwCapabilities: MWCapabilities
+
+  /**
+   * veApiUrl based on top of 'new ApiURLDirecto'
+   */
+  public get veapiUrl(): URL {
+    if (!this._veapiUrl) {
+      // TODO: This depend on baseUrlDirector.buildURL(this.apiPath) and looks like a weak solution
+      this._veapiUrl = this.apiUrlDirector.buildVisualEditorURL()
+    }
+    return this._veapiUrl
+  }
+
+  /**
+   * restApiUrl, apiUrl, modulePath, webUrl and desktopRestApiUr are based on top of 'new BaseURLDirector'
+   */
+  public get restApiUrl(): URL {
+    if (!this._restApiUrl) {
+      this._restApiUrl = this.baseUrlDirector.buildRestApiURL(this.restApiPath)
+    }
+    // TODO: define usage of this property
+    return this._restApiUrl
+  }
+
+  public get apiUrl(): URL {
+    if (!this._apiUrl) {
+      this._apiUrl = this.baseUrlDirector.buildURL(this.apiPath)
+    }
+    return this._apiUrl
+  }
+
+  public get modulePath() {
+    if (!this._modulePath) {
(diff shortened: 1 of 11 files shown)


src/MediaWiki.ts
```diff
@@ -11,26 +11,126 @@ import basicURLDirector from './util/builders/url/basic.director.js'
...
+  private _desktopRestApiUrl: URL
+  // TODO: Mobile builder was removed since there is no /mobile-sections endpoint
+
+  // Set default MW capabilities
+  private readonly mwCapabilities: MWCapabilities
+
+  /**
+   * veApiUrl based on top of 'new ApiURLDirecto'
+   */
+  public get veapiUrl(): URL {
```
kelson42, line 51: `veApiUrl()`, `restApiUrl()`, etc. are shortcuts and offer lazyness to the URL directors. I see little value to have them. I suggest to just something like  this.apiUrlDirector.getVisualEditorURL()` which works in a lazy manner.

If you you think `this.apiUrlDirector.getVisualEditorURL()` is really too cumbersome, then OK to just keep the redirection part, but I see no reason why the lazyness is not built in the Director itself considering that the configuration of it happens at construction time.

src/MediaWiki.ts
```diff
@@ -11,26 +11,126 @@ import basicURLDirector from './util/builders/url/basic.director.js'
...
+    if (!this._veapiUrl) {
+      // TODO: This depend on baseUrlDirector.buildURL(this.apiPath) and looks like a weak solution
+      this._veapiUrl = this.apiUrlDirector.buildVisualEditorURL()
+    }
+    return this._veapiUrl
```
kelson42, line 56: Not easy, considering that you have so many different styles and developers behind the codebase... but think about:
* `Url` vs `URL`
* `VE` vs `VisualEditor`.

src/MediaWiki.ts
```diff
@@ -11,26 +11,126 @@ import basicURLDirector from './util/builders/url/basic.director.js'
...
+    }
+  }
+
+  public hasCoordinatesApi = async function (downloader?: Downloader): Promise<any> {
+    const validNamespaceIds = this.namespacesToMirror.map((ns) => this.namespaces[ns].num)
+    const reqOpts = {
+      action: 'query',
+      format: 'json',
+      // TODO: Do we need this.mwCapabilities.coordinatesAvailable here?
+      prop: `redirects|revisions${this.mwCapabilities.coordinatesAvailable ? '|coordinates' : ''}${this.getCategories ? '|categories' : ''}`,
```
kelson42, line 118: Here there is a chicken-egg problem, not sure how this happened, but you clearly need to put `|coordinates` to then see if this is supported.

src/MediaWiki.ts
```diff
@@ -39,23 +139,23 @@ class MediaWiki {
...
     this.apiUrlDirector = new ApiURLDirector(this.apiUrl.href)
 
-    this.veApiUrl = this.apiUrlDirector.buildVisualEditorURL()
-
-    this.restApiUrl = baseUrlDirector.buildRestApiURL(config.restApiPath)
-    this.desktopRestApiUrl = baseUrlDirector.buildDesktopRestApiURL(config.restApiPath)
-
-    this.modulePath = baseUrlDirector.buildModuleURL(config.modulePath)
+    // Default capabilities
+    // TODO: check whether to remove this object
```
kelson42, line 152: It seems you are almost done and ready to get rid of this `mwCapabilites`.

src/MediaWiki.ts
```diff
@@ -85,12 +185,17 @@ class MediaWiki {
...
+        .then(async (resp) => {
           if (resp.data.login.result !== 'Success') {
             throw new Error('Login Failed')
           }
 
+          /*
+           TODO: Cookie is shared between Downloader and Mediawiki, probably antipattern. Use as interim solution for now.
+           Also, double-check possible race condition - cookies should be set before checking capabilities.
+           */
           downloader.loginCookie = resp.headers['set-cookie'].join(';')
```
kelson42, line 197: Indee... here would help to clearly define what is the duty of the `Mediawiki` and `Downloader` class. Once this is clear, easier to refactor.

src/MediaWiki.ts
```diff
@@ -296,6 +401,12 @@ class MediaWiki {
 
     return mwMetaData
   }
+
```
kelson42, line 404: A few comments have been lost here, maybe on purpose?! Just try to keep things commented when needed.

src/util/mw-api.ts
```diff
@@ -253,3 +254,12 @@ export function mwRetToArticleDetail(obj: QueryMwRet): KVS<ArticleDetail> {
   }
   return ret
 }
+
```
kelson42, line 257: Maybe a comment about what the purpose of this? At a first look, I can not remember.

kelson42:
@VadimKovalenkoSNF I had a first look and for the MWCapabilities, even if the work is not over, it looks good and coherent to me. I see a lots of code which is not related to the fix, not sure if this is not rebased properly on main, if you have other commits unrelated, ....

src/util/mw-api.ts
VadimKovalenkoSNF: I replaced checkApiAvailabilty() method from Downloader because it is more like a utility function.

src/MediaWiki.ts
VadimKovalenkoSNF: I'd like that we have a dedicated ticket about naming conventions.

src/MediaWiki.ts
kelson42: For VE, the answer is in the validation of https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points. For `URL`, I clearly prefer to keep full uppercase for constants.

src/MediaWiki.ts
kelson42: For VE, the answer is in the validation of https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points. For `URL`, I clearly prefer to keep full uppercase for constants but up to you.

VadimKovalenkoSNF: I renamed `ve` to `visualEditor` but regarding `Url` I'd rather keep it as is because there are about ~580 entries across mwoffliner that use the same capitalization.

VadimKovalenkoSNF:
This has been fixed in https://github.com/openzim/mwoffliner/pull/1839

PR #1886, opened 2023-08-21
1881/modularization article treatment - Rearchitecture MWOffliner HTML/CSS/JS scraping (part #2)

### LINKED ISSUE #1830: Rearchitecture MWOffliner HTML/CSS/JS scraping part
Mediawiki provides many API end-points to retrieve HTML/CSS/JS. This is because:
* History (things are evolving, in particular in connection ot the wiki rewritting efforts)
* Wikimedia has it's own set of API end-points at https://en.wikipedia.org/api/rest_v1/
* Mobile or Desktop HTML (they are not the same in Mediawiki)

Depending of Mediawiki versions and how the Mediawiki is configured, not all of them are available.

Currently MWoffliner supports a few of them (see https://github.com/openzim/mwoffliner/issues/1357 and the source code to get more details). But we will need to deprecate and introduce a few new ones: see https://github.com/openzim/mwoffliner/issues/1664 and https://github.com/openzim/mwoffliner/issues/1601.

Mediawiki API HTML/CSS/JS API end-points landscape is not that easy to understand. The other API end-points are more stable and available in all Mediawiki instances (therefore not a problem and not really a topic for this ticket). The only point which is important to understand is that they allow to retrieve HTML and the associated CSS/JS modules (there is a sophisticated module loader called "ResourceLoader") for each article.

We retrieve mobile HTML most of the time, but for a few things (or as fallback) we rely on the Desktop rendering.

In both cases we retrieve the HTML and run transformation on it to make is then directly usable in a ZIM file snapshot (so offline).

The problem is that the part in charge of retrieving the HTML and parsing it is quite a mess:
* No clear separation between the pieces of code dedicated to specific API end-points
* No common interface to use in the same way a module dealing with end-point number 1 or end-point number 2 (although they both provide a way to get HTML)
* A general weak architecture around Mediawiki and Downloader which make any modification complicated

We need to take the code and revamp it a few classes which are clearly identiable (https://github.com/openzim/mwoffliner/issues/1357 is a very small part of that job). Once done, it should be easy/clean to implement the support of additional end-points.

### LINKED ISSUE #1872: Titles of articles are not scraped in WikiMedia ZIMs since July 2023
The latest scrape of English Wikivoyage (29th July) does not appear to include the title of each article in the scrape. Compare the two images below of an article from last month's English Wikivoyage and this month's. Notice how the larger-font title "Santa Cruz de Mompox" is missing from the latest [library.kiwix.org](https://library.kiwix.org/viewer#wikivoyage_en_all_maxi_2023-07/A/Santa_Cruz_de_Mompox) version. It's the same for all other articles as far as I can tell.

image

### LINKED ISSUE #1881: Modularization of article treatment
While saving articles there are multiple article transformations and treatments that apply to multimedia and text content. Considering there will be different kinds of API calls for regular articles (WikimediaMobile for Wikimedia Projects) and the main page (WikimediaDesktop or VisualEditor, see [here)](https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points) there should be created different treatments.

These treatments should be modularized depending on the type of functionality and purpose.
https://github.com/openzim/mwoffliner/pull/1839 partly solves this by dividing treatments by articles (usually text content) and multimedia (images, video etc).

VadimKovalenkoSNF:
Fixes #1881
Fixes #1830
Fixes #1872


src/util/renderers/abstract.renderer.ts
```diff
@@ -732,5 +733,23 @@ export abstract class Renderer {
     return parsoidDoc
   }
 
+  protected injectHeader(content: string, articleDetail: any): string {
```
kelson42, line 736: Please rename to `injectH1TitleToHtml()`, adding a bit of documentation why with need it would be usefull as well.

test/unit/saveArticles.test.ts
```diff
@@ -4,7 +4,7 @@ import domino from 'domino'
 import { setupScrapeClasses } from '../util.js'
 import { saveArticles } from '../../src/util/saveArticles.js'
 import { ZimArticle } from '@openzim/libzim'
-import { mwRetToArticleDetail, DELETED_ARTICLE_ERROR } from '../../src/util/index.js'
+import { mwRetToArticleDetail, DELETED_ARTICLE_ERROR, ARTICLE_HEADER_CLASS } from '../../src/util/index.js'
```
kelson42, line 7: Should not come from `const.js`

kelson42 (changes requested):
@VadimKovalenkoSNF Looks good, but most of the work is not done (or I get it wrong?!) Where is all the part related to the HTML processing?

src/Downloader.ts
```diff
@@ -323,10 +329,14 @@ class Downloader {
 
     return articleRenderer.render({
       data,
+      redisStore,
```
kelson42, line 332: Any reason why the redisStore is not a singleton? Would be easier to handle?!

src/util/renderers/visual-editor.renderer.ts
```diff
@@ -54,19 +53,26 @@ export class VisualEditorRenderer extends Renderer {
...
-
-    if (articleDetail?.title) {
-      header.appendChild(doc.createTextNode(articleDetail.title))
+  public async render(renderOpts: RenderOpts): Promise<any> {
+    try {
+      const result: RenderOutput = []
+      const { redisStore, articleId, articleDetail, webp, _moduleDependencies, dump } = renderOpts
+      const { html, displayTitle } = await this.getHTML(renderOpts)
+      if (html) {
+        const { finalHTML, mediaDependencies, subtitles } = await super.processHtml(html, redisStore, dump, articleId, articleDetail, _moduleDependencies, webp)
```
kelson42, line 62: `getHTML()` vs `processHTML()`, please decides one or the other form, be keep it coherent.

kelson42 (changes requested):
LGTM in general, I see not oddity or anything schoking me. That said a few small things to improve and have open IMO an important discussion about the role and the articulation of the software around `Dump.ts`.

src/util/renderers/visual-editor.renderer.ts
VadimKovalenkoSNF: Due to the logic behind `getHTML()` I incline to rename it to the `processRender()` or `preRender()`. `processHTML` will stand for treatments as before. Or are you concerned about capitalization `HTML` vs `Html`?

src/Downloader.ts
VadimKovalenkoSNF: Yes, Redis store will be handled more conveniently as a singleton. I did these changes in 231e403433a5a3d1a31370fe2ecf03a0fda317a0

test/unit/saveArticles.test.ts
VadimKovalenkoSNF: Ok, I added the article header class as a regular string now.

src/util/renderers/abstract.renderer.ts
VadimKovalenkoSNF: Done.

src/Downloader.ts
```diff
@@ -323,10 +329,14 @@ class Downloader {
...
+      webp,
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: I'm a bit concerned about the multiple responsibilities of `saveArticles()`. For instance, `const { jsModuleDependencies, cssModuleDependencies } = await saveArticles(zimCreator, downloader, redisStore, dump)` [here](https://github.com/openzim/mwoffliner/blob/main/src/mwoffliner.lib.ts#L422C5-L422C121) allows to get js and CSS dependencies and at the same time write HTML into zim and modify dump instance while render process. Maybe this is also a symptom of weak architecture here. As for DumpOpts, I agree that this is a better idea to pass options instead of a full dump instance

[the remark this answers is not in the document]
kelson42: I agree, I would really like to have this js/css module handling done in a dedicated method

src/util/renderers/visual-editor.renderer.ts
```diff
@@ -15,37 +14,37 @@ export class VisualEditorRenderer extends Renderer {
     super()
   }
 
-  public async render(renderOpts: RenderOpts): Promise<any> {
-    const { data, isMainPage, articleId, articleDetail } = renderOpts
+  private getDisplayTitleFromVisualEditor(strippedTitle: string, articleId: string) {
```
kelson42, line 17: Name here seems redundant, why not just `getDisplayedTitle()`?

src/util/renderers/visual-editor.renderer.ts
```diff
@@ -15,37 +14,37 @@ export class VisualEditorRenderer extends Renderer {
...
-      strippedTitle = getStrippedTitleFromHtml(dataHtml)
-      result.push({
-        articleId,
-        displayTitle: strippedTitle || articleId.replace('_', ' '),
-        html: dataHtml,
-      })
-      return result
+      html = isMainPage ? data.visualeditor.content : super.injectH1TitleToHtml(data.visualeditor.content, articleDetail)
+      strippedTitle = getStrippedTitleFromHtml(html)
+      displayTitle = this.getDisplayTitleFromVisualEditor(strippedTitle, articleId)
```
kelson42, line 40: I don't know all the details around these two methods, but no way to make simpler? Maybe in one pass?

src/util/saveArticles.ts
```diff
@@ -134,19 +128,41 @@ async function downloadBulk(listOfArguments: any[], downloader: Downloader): Pro
 async function getAllArticlesToKeep(downloader: Downloader, articleDetailXId: RKVS<ArticleDetail>, dump: Dump, mainPageRenderer: Renderer, articlesRenderer: Renderer) {
   await articleDetailXId.iterateItems(downloader.speed, async (articleKeyValuePairs) => {
     for (const [articleId, articleDetail] of Object.entries(articleKeyValuePairs)) {
+      const nonPaginatedArticleId = articleDetail.title
```
kelson42, line 131: Here:
* What is this variable for? I don't understand why we now talk about `nonPaginated`? Is that related to the category handling?
* A `title` is not an `articleId`, here we have a least a problem with variable naming.

src/util/saveArticles.ts
```diff
@@ -293,24 +296,44 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
         curArticle = articleId
         const promises: [string, Promise<Error>][] = []
 
+        const nonPaginatedArticleId = articleDetail.title
```
kelson42, line 299: Same as before, a "title" is not an "articleId".

test/unit/bootstrap.ts
```diff
@@ -2,18 +2,18 @@
 /// <reference path="../../src/types.d.ts" />
 
 import 'dotenv/config'
-import RedisStore from '../../src/RedisStore'
-import { config } from '../../src/config'
+import RedisStore from '../../src/RedisStore.js'
```
kelson42, line 5: We write typescript, so why `.js` and not `.ts`?

kelson42 (changes requested):
Few things, but looks good in general.

src/util/renderers/visual-editor.renderer.ts
kelson42: I hardly can follow why a function in charge of retrieving HTML should be called `preRender()`! This is just unclear what it does. Maybe `retrieveHTML()` would be a better name? Yes I'm worry about capitalisation, make it coherent, one way or the other.

src/util/renderers/visual-editor.renderer.ts
kelson42: We should here rename to talk about the `H1Title`, so all the function related to this "title" have a similar name.

kelson42:
I would also suggest to move `src/util/renders` to `src/renders`

test/unit/bootstrap.ts
VadimKovalenkoSNF: Withoud `.js` extension I have this ts-lint error in my IDE: 
`Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.`

So I have to add `.js` at the end of import to fix this problem. By the way, this approach is used in a lot of places on the mwoffliner. More info about this bug can be found here ( https://github.com/microsoft/TypeScript/issues/49083 ) but I don’t think we need to pay much attention to it since it doesn’t break anything.

src/util/saveArticles.ts
VadimKovalenkoSNF: I got rid of `nonPaginatedArticleId` because it was confusing indeed.

src/util/renderers/visual-editor.renderer.ts
VadimKovalenkoSNF: I removed `this.getDisplayTitleFromVisualEditor()` in favor of the explicit title parsing like `displayTitle = strippedTitle || articleId.replace('_', ' ')`

src/util/renderers/visual-editor.renderer.ts
VadimKovalenkoSNF: I decided to stick to the `retrieveHtml()` here.

test/unit/bootstrap.ts
kelson42: @VadimKovalenkoSNF Please open a dedicated ticket about that (I don't really have the time right now to go through the whole stuff). Even if this looks like a detail, I hardly can understand how it could be OK to have Typescript code in `.js`.

src/util/saveArticles.ts
kelson42: It seems you have removed it as well, I don't see it. AFAIK there should be nowhere a confusion between "title" and "articleId":
* "Title" is the title of the article, as ZIM metadata and should also be the same as the content of the `<title>` DOM node
* "articleId"  is the identifier of the article, this is typically the one we have in the Mediawiki URL

If there is no proven strong mix of these two concept, then I'm OK to close this discussion. Otherwise a ticket should be made.

src/Downloader.ts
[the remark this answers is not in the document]
kelson42: OK, to me it sounds we should almost have a #3 of rearchitecture to get rid of `saveArticles()` and put the main routine at a better place. That said, unsure about how the solution should look like and when we should do this. At this stage, I propose to just keep things like they are as long as a solution does not appear clearly. Once this PR merge, we should really focus then to fix the bugs and get again the mobile-html stuff working finel. That's the priority. We can close this discussion.

kelson42:
@VadimKovalenkoSNF If OK for a final review, then please request a new review. On my side it looks good, but would like to make a last review pass this evening.

test/unit/bootstrap.ts
VadimKovalenkoSNF: I don't think it does make sense to open a separate ticket since this is Typescript side effect and not a mwoffliner bug.

src/util/saveArticles.ts
VadimKovalenkoSNF: I've updated getModuleDependencies() in unit tests, so the ambiguity between `title` and `articleId` is no longer present for this method.

src/Downloader.ts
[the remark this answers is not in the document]
VadimKovalenkoSNF: Agree, lets focus on bugs and mobile-html first.

VadimKovalenkoSNF:
> I would also suggest to move src/util/renders to src/renders
Done

test/unit/bootstrap.ts
kelson42: where the root cause is located, is secondary. What is portant is impact and tracking of it. Without ticket, we can not track.

VadimKovalenkoSNF: I opened related ticket - https://github.com/openzim/mwoffliner/issues/1900

VadimKovalenkoSNF:
Codefactor fixes are here - https://github.com/openzim/mwoffliner/pull/1901

kelson42 (approved):
LGTM

Issue #1887, opened 2023-08-23
One or two e2e tests should be modularized


kelson42:
Now that the overall architecture is healthier we should modularized the e2e tests.

The problem is that for the moment each e2e tests deal only with one API end-point. But we support 2-3 at leasts. So to be serious each e2e tests should run the very same tests on all supported end-points to sercure the treatments applys properly for each case.

We should do that for one or two tests to secure it works fine, and afterward, piece by piece, we will move all e2e tests to tests all supported API end-points.

VadimKovalenkoSNF:
1. Current e2e tests are lying in a folder [test/e2e/articleLists.test.ts](https://github.com/openzim/mwoffliner/tree/main/test/e2e)
2. At the moment, mwoffliner supports **WikimediaDesktop** and **VisualEditor** API capabilities. **WikimediaMobile** API to be done.
3. There are a bunch of unit / international tests in [test/unit](https://github.com/openzim/mwoffliner/tree/main/test/unit). This folder contains tests for renderers, URL builders, article and media treatments, AWS S3, Redis store, etc. Overall test structure in this folder is a bit unstructured IMO.

I'm thinking about this strategy to start:
1. Take **test/e2e/bm.e2e.test.ts** 
2. Force bm wiki test to use both renderers ( **WikimediaDesktop** and **VisualEditor**) sequentially.  
3. Possibly, replace/duplicate some test cases from article and media treatments in **test/e2e/bm.e2e.test.ts**.

@kelson42 what do you think?

kelson42:
@VadimKovalenkoSNF Your plan LGTM. T me its important that we have a clear, elegant and reusable "code template" to run an (e2e) test easily against all supported end-points without much hasdle and duplication of code.

PR #1898, opened 2023-08-30
Modularizing e2e tests

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1887


src/parameterList.ts
```diff
@@ -39,6 +39,7 @@ export const parameterDescriptions = {
   osTmpDir: 'Override default operating system temporary directory path environment variable',
   customFlavour: 'A custom processor that can filter and process articles (see extensions/*.js)',
   optimisationCacheUrl: 'S3 url, including credentials and bucket name',
+  renderName: 'Custom render API',
```
kelson42, line 42: `forceRender: Force the usage of a specific API end-point/render, automatically chosen otherwise. Accepted values: [ VisualEditor, WikimediaDesktop. WikimediaMobile ]`. More details at https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points`

test/e2e/rendererList.ts
```diff
@@ -0,0 +1 @@
+export const renderers = ['WikimediaDesktop', 'VisualEditor']
```
kelson42, line 1: This render list should somehow be already available on the code base. I don't see why it should be redefined... or do you want to test less than what is available?

test/e2e/rendererList.test.ts
```diff
@@ -0,0 +1,55 @@
+import * as mwoffliner from '../../src/mwoffliner.lib.js'
```
kelson42, line 1: I don't understand the purpose of this new test. It does not go through a loop to test elegantly all api-end point?!

test/e2e/bm.e2e.test.ts
```diff
@@ -9,78 +10,85 @@ import { zimdumpAvailable, zimdump } from '../util.js'
...
-    for (const dump of outFiles) {
-      if (dump.nopic) {
-        // nopic has enough files
-        expect(dump.status.files.success).toBeGreaterThan(14)
-        // nopic has enough redirects
-        expect(dump.status.redirects.written).toBeGreaterThan(170)
-        // nopic has enough articles
-        expect(dump.status.articles.success).toBeGreaterThan(700)
+  for (const renderer of renderers) {
+    // Test only render API that is supported by the wiki
```
kelson42, line 14: why, we want to test all end-points?! we should better want to test against a wiki which support all end-points... which is the case here anyway.

kelson42 (changes requested):
I have left a few comments

test/e2e/rendererList.test.ts
VadimKovalenkoSNF: It tests the renderName param for bm.wikipedia.org.

kelson42:
@VadimKovalenkoSNF Can you please:
* Fix a few comments regarding the code allowing to force a dedicated render
* Move this part in an other PR (wihtout the testing part)

That way, I would merge straight, as this LGTM. We could then keep this PR for focus on the testing only.

kelson42:
@VadimKovalenkoSNF This PR should be rebased/revamped, now that I have merged #1901. Hopefully, then I will better understand the way how you solve the problem.

test/e2e/rendererList.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      console.log('Zimcheck not installed, skipping test')
+    }
+
+    rimraf.sync(`./${testId}`)
+    const redisScan = await execa('redis-cli --scan', { shell: true })
+    // Redis has been cleared
+    expect(redisScan.stdout).toEqual('')
+  })
+
+  test('Scrape article from bm wiki should throw error when using VisualEditor renderName', async () => {
```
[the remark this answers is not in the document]
VadimKovalenkoSNF: Yes, it is bm.wikipedia.org, I made this description more explicit.

test/e2e/rendererList.ts
VadimKovalenkoSNF: We don't keep a list of renderers in a separate place (for now). Only `type renderName` in **abstract.renderer.ts**

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+  const now = new Date()
+  const testId = `mwo-test-${+now}`
+
+  const parameters = setDefaultParams(renderer, testId)
+  await execa('redis-cli flushall', { shell: true })
+  const outFiles = await mwoffliner.execute(parameters)
+  await expect(zimcheck(outFiles[0].outFile)).resolves.not.toThrowError()
+
+  const articleFromDump = await zimdump(`show --url A/${parameters.articleList} ${outFiles[0].outFile}`)
```
VadimKovalenkoSNF, line 60: There is the ability to get an article HTML form dump and test it directly along with other dump output. This might be useful for e2e testing.

test/e2e/rendererList.ts
kelson42: We do in `sanitize_forceRender()`. Should be defined somewhere where each pice of software can access it, maybe in `const.ts`?

test/e2e/bm.e2e.test.ts
kelson42: so why we do have `if (renderer === 'WikimediaDesktop)'`, the loop should apply for all renderers, not only `WikimediaDesktop`.

test/e2e/rendererList.test.ts
kelson42: OK, I get it now

test/e2e/rendererList.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    redis: process.env.REDIS,
+    format: ['nopic'],
+    articleList: 'Fàransi, Kanada',
+  }
+
+  beforeAll(async () => {
+    await execa('redis-cli flushall', { shell: true })
+  })
+
+  test('Scrape article from bm wiki using WikimediaDesktop renderName', async () => {
```
kelson42, line 26: It's "bm.wikipedia.org"

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        await commonTreatmentTest(renderer)
+      })
+    }
+    if (renderer === 'VisualEditor') {
+      test('VisualEditor e2e', async () => {
+        await commonTreatmentTest(renderer)
+      })
+    }
+  }
+})
```
kelson42, line 92: You can do better: should be doable in 3 lines.

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+  if (!zimdumpIsAvailable) {
+    console.log('Zimcdump not installed, skipping test')
+    return
+  }
+  const now = new Date()
+  const testId = `mwo-test-${+now}`
+
+  const parameters = setDefaultParams(renderer, testId)
+  await execa('redis-cli flushall', { shell: true })
```
kelson42, line 56: I don't know where this is from, but this is very very ugly:
* Because of execa
* Because of flushing everything (maybe other table from other processes)

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  return false
+}
+
+const commonTreatmentTest = async (renderer) => {
+  if (!zimcheckIsAvailable) {
+    console.log('Zimcheck not installed, skipping test')
+    return
+  }
+  if (!zimdumpIsAvailable) {
+    console.log('Zimcdump not installed, skipping test')
```
kelson42, line 49: typo

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+jest.setTimeout(200000)
+
+let zimcheckIsAvailable
+let zimdumpIsAvailable
+
+beforeAll(async () => {
+  zimcheckIsAvailable = await zimcheckAvailable()
+  zimdumpIsAvailable = await zimdumpAvailable()
+})
+
```
kelson42, line 19: If we could avoid to do that for each test, this would be really better. No way to do simpler?

test/e2e/treatments.e2e.test.ts
```diff
@@ -0,0 +1,92 @@
...
+const setDefaultParams = (renderName: string, testId: string) => {
+  return {
+    mwUrl: 'https://en.wikipedia.org',
+    adminEmail: 'test@kiwix.org',
+    outputDirectory: testId,
+    redis: process.env.REDIS,
+    articleList: 'User:Kelson/MWoffliner_CI_reference',
+    forceRender: renderName,
+  }
+}
```
kelson42, line 29: Same here, we should have a function just doing straight the job and returning the `outFiles`. The variables would simply be the `articleList` and the `renderName`, maybe the `mwUrl` as optional parameter. This file should really be reduced to the bar minimum of line with only the interesting stuff which is basically the `commonTreatmentTest()`.

At the end we might just achieve to have one call with the test in callback and that's all?!

kelson42 (changes requested):
Once we are at the end of the simplification, we should fix the `nodet` test as well.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: All e2e tests in **test/e2e** folder use **execa** to flush redis store before each test run.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: This `beforeAll` hook is placed outside `describe` blocks, so it will execute only once.

test/e2e/bm.e2e.test.ts
VadimKovalenkoSNF: Because `bm.wikipedia.org` (which is the target of this test) doesn't support VE. When it meets VisualEdtitor it will fail, so I explicitly set `if (renderer === 'WikimediaDesktop)'. If it was `en.wikipedia.org`, your concern would be valid. But here we are faced with a problem about many-to-may relationships in tests: check multiple wikis with multiple endpoints at one go.

kelson42: VE does work on bm.wikipedia.org, check help on api.php endpoint.

VadimKovalenkoSNF: Yes, you're right, I've updated bm.e2e.test.ts

test/e2e/rendererList.ts
VadimKovalenkoSNF: I've replaced the list of renderers to `const.ts` as you suggested.

test/e2e/rendererList.test.ts
VadimKovalenkoSNF: Fixed.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: Done, pls check.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: Fixed

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: I've created `getOutFiles()` function where I keep all the logic about creating ZIM and returning `outFiles`, pls check.

VadimKovalenkoSNF:
> Once we are at the end of the simplification, we should fix the nodet test as well.
I fixed this in the scope of this PR.

VadimKovalenkoSNF:
There is a side effect of e2e testing of all renderers is drastically increased time of test running.

test/e2e/bm.e2e.test.ts
```diff
@@ -1,86 +1,90 @@
...
-  const testId = `mwo-test-${+now}`
-
-  const parameters = {
-    mwUrl: 'https://bm.wikipedia.org',
-    adminEmail: 'test@kiwix.org',
-    outputDirectory: testId,
-    redis: process.env.REDIS,
-    format: ['nopic'],
-  }
+  for (const renderer of RENDERERS_LIST) {
```
kelson42, line 12: All of this will come in each e2e test, why not something like `testWithAllRenderers([options like "format" or  mwUrl"], {callback code containing only the test})`

With variables like `outFiles` or `Dump`.

We should have a test with the only the very unique specific part.

Here I see far too much duplicate code

kelson42 (changes requested):

kelson42:
> There is a side effect of e2e testing of all renderers is drastically increased time of test running.

Yes, this will be a problem and we will have to handle it to have e2e which are smaller/quicker to execute, so no scrape with many articles

test/e2e/treatments.e2e.test.ts
kelson42: The problem is not the execution time, but that we have to place this piece of code almost in all e2e test.

test/e2e/treatments.e2e.test.ts
kelson42: I don't really understand why yor comment come here, but zimcheck should be run everywhere, so why I have eachtime to request it? Create a test template/fonction/classs so this is part is present only once in the whole codebase.

test/e2e/treatments.e2e.test.ts
kelson42: Better, but why we have an `if` on Render names? All renders should be tested.

test/e2e/treatments.e2e.test.ts
kelson42: It's not an execution problem, it's a code problem, these lines should be only once coded in the whole codebase.

test/e2e/treatments.e2e.test.ts
kelson42: OK, but at the same time you see, already in that PR, that for many e2e tests you need to access to the `dump` object! So this is is only soving the problem halway.

Why not having something like `testWithAllRenders(mwUrl, format, ..., callback{})` and in the callback you get access to `outFiles` to `dump` and whatever is necessary for e2e test to be done specifically.

e2e should only contain their very specific lines and nothing else.

This wish is actually for everything, please factorise your code the most optimal manner. The less code it is the better.

kelson42 (changes requested):
Still too complex, we should have one test with no boilerplate, nothing which is repeated from test to test

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: I faced the problem that I couldn't nest a test block inside another test or callback function in Jest. For comparison purpose I created **test/e2e/en.e2e-callback.test.ts** with **test/testAllRendersTemplateCallback.ts** and  **test/e2e/en.e2e.test.ts** with **test/testAllRendersTemplate.ts**. Pros and cons:
1. test/testAllRendersTemplate.ts - have to keep all the treatment tests inside but related **test/e2e/en.e2e.test.ts** is pretty small.
2. **test/testAllRendersTemplateCallback.test.ts** - returns only the outFile and overall has a single responsibility. Cons: you can't avoid forEach loop in related **test/e2e/en.e2e-callback.test.ts** and other e2e tests that will use this template.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: I made it just for safety for cases if some wiki doesn't support a specific wiki.

VadimKovalenkoSNF: This test file has been refactored and renamed to **test/e2e/en.e2e.test.ts**.

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: Done in **test/testAllRenders.ts**

test/e2e/treatments.e2e.test.ts
VadimKovalenkoSNF: Callback inside for..of renders loop has been implemented **test/testAllRenders.ts**

test/testAllRenders.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  await execa('redis-cli flushall', { shell: true })
+  const outFiles = await mwoffliner.execute(parameters)
+
+  return outFiles
+}
+
+export async function testAllRenders(mwUrl: string, articleList: string, format: string | string[], callback) {
+  const zimcheckIsAvailable = await zimcheckAvailable()
+  const zimdumpIsAvailable = await zimdumpAvailable()
```
kelson42, line 32: This is a remark I have made in other PR. Test once and if not available just exit the test with error. These tools are mandatory. Does not make sense to test for their presence in each test.

test/testAllRenders.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (!zimcheckIsAvailable || !zimdumpIsAvailable) {
```
kelson42, line 34: You can assume these tools are there.

test/unit/saveArticles.test.ts
```diff
@@ -79,69 +80,51 @@ describe('saveArticles', () => {
...
-
-    const articleDoc = domino.createDocument(result[0].html)
-
-    const sections = Array.from(articleDoc.querySelectorAll('section'))
-    const leadSection = sections[0]
-    expect(sections.length).toEqual(1)
-    expect(leadSection.getAttribute('data-mw-section-id')).toEqual('0')
-  })
+  for (const renderer of RENDERERS_LIST) {
+    if (['WikimediaDesktop', 'VisualEditor'].includes(renderer)) {
```
kelson42, line 84: I want all the tests to run through all renderers: always from now.

kelson42 (changes requested):
A lot better, but still a few things to fix

test/unit/saveArticles.test.ts
VadimKovalenkoSNF: Fixed.

test/testAllRenders.ts
VadimKovalenkoSNF: Added process.exit(1) if no zimtool available.

test/testAllRenders.ts
VadimKovalenkoSNF: I updated this condition to exit if no zimtool available (if I understood your comment correctly here).

kelson42: Not really I guess, you test that once at the whole test process:
* either it is there and you can assume in all the tests, they are there, so you don't test the presence of zimcheck and zimdump anymore
* or you just exit with an error

kelson42 (changes requested):
Test zimcheck/zimdump only once in the whole test process!

test/testAllRenders.ts
VadimKovalenkoSNF: I added lazy **checkZimTools()**

PR #1901, opened 2023-09-06
Introduce forceRender param

VadimKovalenkoSNF:
Create `forceRender` param to CLI, this is an enhancement for https://github.com/openzim/mwoffliner/issues/1887

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -651,7 +651,7 @@ class Downloader {
     const articleData = await this.getJSON<any>(articleApiUrl)
 
     if (articleData.error) {
-      const errorMessage = `Unable to retrieve js/css dependencies for article '${title}': ${articleData.error.code}`
+      const errorMessage = `Unable to retrieve js/css dependencies for article '${this}': ${articleData.error.code}`
       logger.error(errorMessage)
 
       /* If article is missing (for example because it just has been deleted) */
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -179,19 +179,12 @@ class Downloader {
         { condition: await MediaWiki.hasWikimediaDesktopRestApi(), value: MediaWiki.desktopRestApiUrl.href },
         { condition: await MediaWiki.hasVisualEditorApi(), value: MediaWiki.visualEditorApiUrl.href },
       ])
-
-      logger.log('Base Url: ', this.baseUrl)
-      logger.log('Base Url for Main Page: ', this.baseUrlForMainPage)
-
-      if (!this.baseUrl || !this.baseUrlForMainPage) throw new Error('Unable to find appropriate API end-point to retrieve article HTML')
     } else {
       switch (forceRender) {
         case 'WikimediaDesktop':
           if (MediaWiki.hasWikimediaDesktopRestApi()) {
             this.baseUrl = MediaWiki.desktopRestApiUrl.href
             this.baseUrlForMainPage = MediaWiki.desktopRestApiUrl.href
-            logger.log('Base Url: ', this.baseUrl)
-            logger.log('Base Url for Main Page: ', this.baseUrlForMainPage)
             break
           }
           break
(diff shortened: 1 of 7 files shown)


test/e2e/forceRender.test.ts
```diff
@@ -0,0 +1,55 @@
+import * as mwoffliner from '../../src/mwoffliner.lib.js'
```
VadimKovalenkoSNF, line 1: I decided to keep this test as this is the separate testing of `forceRender` param with bm.wikipedia.org

kelson42:
@VadimKovalenkoSNF Thank you, please fix codefactor so I can merge.

VadimKovalenkoSNF:
> please fix codefactor so I can merge.

Done

src/Downloader.ts
```diff
@@ -164,23 +164,49 @@ class Downloader {
...
+
+      if (!this.baseUrl || !this.baseUrlForMainPage) throw new Error('Unable to find appropriate API end-point to retrieve article HTML')
+    } else {
+      switch (forceRender) {
+        case 'WikimediaDesktop':
+          if (MediaWiki.hasWikimediaDesktopRestApi()) {
+            this.baseUrl = MediaWiki.desktopRestApiUrl.href
+            this.baseUrlForMainPage = MediaWiki.desktopRestApiUrl.href
+            logger.log('Base Url: ', this.baseUrl)
+            logger.log('Base Url for Main Page: ', this.baseUrlForMainPage)
```
kelson42, line 192: Why these twho lines are duplicated everywhere? Just do it once at the end would be a better approach IMO.

src/util/saveArticles.ts
```diff
@@ -247,7 +247,7 @@ export function getArticleUrl(downloader: Downloader, dump: Dump, articleId: str
 /*
  * Fetch Articles
  */
-export async function saveArticles(zimCreator: ZimCreator, downloader: Downloader, redisStore: RS, dump: Dump) {
+export async function saveArticles(zimCreator: ZimCreator, downloader: Downloader, redisStore: RS, dump: Dump, forceRender = null) {
```
kelson42, line 250: This looks wrong to me, we should transmit the renders objects directly

src/util/saveArticles.ts
```diff
@@ -562,52 +606,21 @@ export function applyOtherTreatments(parsoidDoc: DominoElement, dump: Dump) {
...
-      for (const section of sections) {
-        if (
-          section.children.length ===
-          Array.from(section.children).filter((child: DominoElement) => {
-            return child.matches('summary')
-          }).length
-        ) {
-          DU.deleteNode(section)
-        }
+    // Mobile view === details
```
kelson42, line 609: Should we not move this (remove empty sections) in the renders code?

src/util/saveArticles.ts
```diff
@@ -526,6 +534,42 @@ export function applyOtherTreatments(parsoidDoc: DominoElement, dump: Dump) {
       }
     }
   }
+}
+
+const clearNodes = (parsoidDoc: DominoElement, filtersConfig: any) => {
```
kelson42, line 539: Should we not have this in renders as part of the treatments?

kelson42 (changes requested):
Few questions still open my side

kelson42:
@VadimKovalenkoSNF A rebase (with conflict fixing) will be needed here

src/util/saveArticles.ts
VadimKovalenkoSNF: We need to keep `forceRender` argument in saveArticles. When run the `execute()` function, we grab `forceRender` from params and pass it to `saveArticles()` because we instantiate the renderer builder with `auto` or `specific` render type option there. Or you suggest to instantiate it in `src/mwoffliner.lib.ts` directly and pass entire `mainPageRenderer` and `articlesRenderer` into saveArticles()?

src/Downloader.ts
VadimKovalenkoSNF: Good catch, fixed

src/util/saveArticles.ts
VadimKovalenkoSNF: Yes, this moved to the abstract renderer now after rebasing on top of recent changes with https://github.com/openzim/mwoffliner/pull/1886

src/util/saveArticles.ts
VadimKovalenkoSNF: Done.

src/util/saveArticles.ts
kelson42: IMO we face here again the problematic around bad architecture of `saveArticles()`. Skipping for the moment.

kelson42 (approved):
LGTM

PR #1903, opened 2023-09-11
Implement new Wikimedia mobile end-point support/render

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1664
Fixes: https://github.com/openzim/mwoffliner/issues/1916
Blocked by https://github.com/openzim/mwoffliner/issues/1925


VadimKovalenkoSNF:
This patch is not ready for the review yet. I found that article treatments are not working for mobile as I expected first. I pushed preliminary commit b7f6e55cf040a961203b1030c3d6e1c90ee2e7c3 to pass output from `page/mobile-html` directly to zim creator, but this could not be considered as a final solution - Kiwix app doesn't render it correctly though in local Kiwix server I can see exact same response as I get from `page/mobile-html` endpoint. There is another approach which might be better - instead of `page/mobile-html`, use `page/html` and process sections using scripts from [deprecated MCS](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/sections/parsoidSections.js) but inside mwoffliner. This will allow to preserve functionality that has been working before but without `mobile-sections` endpoint. On the other hand, it might require additional effort to apply it inside mwoffliner.

VadimKovalenkoSNF:
Current implementation of mobile-html render has additional treatments:
1. Connect mobile js and css that are required for mobile-html.
2. Remove Edit icon near each section.
3. Restore lazy load images. Current mobile-html output returns span with a lazy-load placeholder that is not transforming into the image on scroll in Kiwix app.
4. Restore sections' appearance.
5. Disable event listeners for sup elements. This can be achieved only on the client side, so I added an additional script tag with this functionality to the output.
6. Added additional styles to override redundant or invalid CSS rules to make final output more robust.

src/Downloader.ts
```diff
@@ -170,6 +170,7 @@ class Downloader {
     if (!forceRender) {
       //* Objects order in array matters!
       this.baseUrl = basicURLDirector.buildDownloaderBaseUrl([
+        { condition: await MediaWiki.hasWikimediaMobileRestApi(), value: MediaWiki.mobileRestApiUrl.href },
         { condition: await MediaWiki.hasWikimediaDesktopRestApi(), value: MediaWiki.desktopRestApiUrl.href },
```
kelson42, line 174: I don't really know why the `Rest` came in here?! It does not match the [official names on the wiki](https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points)

I would realy want to keep the word `Rest` for [these API](https://www.mediawiki.org/wiki/API:REST_API)

src/MediaWiki.ts
```diff
@@ -58,8 +60,10 @@ class MediaWiki {
   public _modulePathOpt: string // only for whiting to generate modulePath
   public webUrl: URL
   public desktopRestApiUrl: URL
+  public mobileRestApiUrl: URL
```
kelson42, line 63: Here the same, I see no value of `Rest` in the variable name, to the contrary. Or do I have forgotten something important?

src/renderers/renderer.builder.ts
```diff
@@ -23,14 +28,19 @@ export class RendererBuilder {
...
       case 'mobile':
-        // TODO: return WikimediaMobile renderer
-        break
+        if (hasWikimediaMobileRestApi) {
+          return new WikimediaMobileRenderer()
+        }
+        logger.error('No available mobile renderer.')
+        process.exit(1)
       case 'auto':
         if (hasWikimediaDesktopRestApi) {
```
kelson42, line 37: Logic seems wrong, mobile rendering should have the top priority?!

src/util/saveArticles.ts
```diff
@@ -260,21 +242,27 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
...
+    articlesRenderer = await rendererBuilder.createRenderer(rendererBuilderOptions)
   } else {
     rendererBuilderOptions = {
-      renderType: 'auto',
+      renderType: 'desktop',
     }
+    mainPageRenderer = await rendererBuilder.createRenderer(rendererBuilderOptions)
+    // If the mobile renderer API is not available, switch articles rendering to the auto mode instead
+    rendererBuilderOptions.renderType = hasWikimediaMobileRestApi ? 'mobile' : 'auto'
+    articlesRenderer = await rendererBuilder.createRenderer(rendererBuilderOptions)
```
kelson42, line 263: Do simpler, just call `createRenderer` with the appropriate value or expression. Variable should be instanciated only when:
* modified at some point
* use more than once

... ortherwise this is just a wast of resources.

kelson42 (changes requested):
Architecture looks good but a few details to fix to keep things lean.
Important now, is to get all the tests passing so we can make the final review.

src/renderers/renderer.builder.ts
VadimKovalenkoSNF: This brings us to the recent conversation about `auto` mode, where we have to choose `desktop` if no `forceRender` is specified and no `mobile` capability is present. Check this line of code [here](https://github.com/openzim/mwoffliner/blob/251b6bb93f737b107ae6f356c0870cac9f6036f4/src/util/saveArticles.ts#L257) where I implemented an `auto` mode as a fallback for `mobile`.

src/util/saveArticles.ts
VadimKovalenkoSNF: Done, pls check.

src/MediaWiki.ts
VadimKovalenkoSNF: Yes, this naming came from url builders, I've updated it where it was possible.

src/Downloader.ts
VadimKovalenkoSNF: I removed `Rest` wording for desktop and mobile.

VadimKovalenkoSNF:
Set this to the review. To secure tests I explicitly set `forceRender: 'WikimediaDesktop'` where it was necessary. Once https://github.com/openzim/mwoffliner/pull/1898 is approved, we can start to test mobile treatments along with other functionality without duplicating test files, etc.

kelson42:
@VadimKovalenkoSNF Great, do we have anything know not working fine specificily with this new end-point and which would need further work (and an other ticket)?

VadimKovalenkoSNF:
@kelson42 collapsible sections not working (don't mismatch this with collapsible infoboxes) + I want you to look at the HTML output and give me feedback about it ( mostly form a user perspective ). Note that `page/mobile-html` is different than `page/html` but I've tried to apply transformations to elements that are not valid for the output.

VadimKovalenkoSNF:
I noticed that images in gallery box don't render and `nopic` parameter doesn't work as expected.

kelson42:
This is essential that both bugs are fixed

VadimKovalenkoSNF:
@kelson42 , I created a separate test file `test/e2e/mobileRenderIntegrity.test.ts` to check zim integrity for mobile renderer. It looks like I still got error about external pcs script there. Steps to reproduce:

Download this patch and run:
`npm run test:pattern 'test/e2e/mobileRenderIntegrity.test.ts'`

Output that I received:
> [ERROR] Invalid external links found:
      https://meta.wikimedia.org/api/rest_v1/data/javascript/mobile/pcs is an external dependence in article A/Canada

And the error about invalid internal links:
> [ERROR] Invalid internal links found:
      The following links:
    - ../-/File:&quot;O_Canada&quot;,_performed_by_the_United_States_Third_Marine_Aircraft_Wing_Band.oga-bg.vtt
    (-/File:&quot;O_Canada&quot;,_performed_by_the_United_States_Third_Marine_Aircraft_Wing_Band.oga-bg.vtt) were not found in article A/Canada
    
I would appreciate your input about how to solve this. Is there any effort required for **zimcheck** itself to fix this. Or is this only a mwoffliner issue? Link rels seems to work as expected. List of external resources that needed for mobile renderer can be found here - https://en.wikipedia.org/api/rest_v1/page/mobile-html-offline-resources/Canada *
(* title 'Canada' or any other title is required for this API)

kelson42:
@VadimKovalenkoSNF All internal links pointing to articles which are not mirrored should be removed.

VadimKovalenkoSNF:
> [ERROR] Invalid external links found:
This one has been fixed, I mirrored mobile modules on the right way now.

> [ERROR] Invalid internal links found
And this one doesn't seem to be related to the mobile renderer. It works depending on the article title, I left a comment [here](https://github.com/openzim/mwoffliner/pull/1903/files#diff-3adb2679187282029a394bfb2d837d0f38b853236802aec736cc718767815482R51), maybe this should be highlighted in a separate ticket.

kelson42 (changes requested):
I have finally run with thie renderer and we still have serious problems:
* Many editing artifact which should be removed (to add description under the title, to edit wikidata entries, ....) See 
image
* None of the internal links work because somehow the onclick event is still caught

VadimKovalenkoSNF:
> Many editing artifact which should be removed (to add description under the title, to edit wikidata entries, ....) See

I removed them. If there is anything else for removal, I prepared a separate res/pcs/pcs_override_style.css where we can apply CSS overrides

> None of the internal links work because somehow the onclick event is still caught

I fixed this, pls check.

VadimKovalenkoSNF:
Tested mobile renderer output for **bm.wikipedia.org**:
**--format=nopic** - works as expected, **2 Mb**
**--format=nodet** - works OK, **60 Mb**
**--format=nodet, nopic** - works OK, **1.6 Mb**
**bm.wikipedia.org  as is** - works OK, but the size is **101,5 Mb**, twice bigger than WikimediaDesktop render, this needs to be fixed. I suspect the problem is with image optimization. Here is a screenshot with comparison between same images that come from desktop and mobile renders.

image-comparison

src/renderers/abstract.renderer.ts
```diff
@@ -485,8 +490,33 @@ export abstract class Renderer {
           styleDependenciesList.length !== 0
             ? styleDependenciesList.map((oneCssDep) => genHeaderCSSLink(config, oneCssDep, articleId, config.output.dirs.mediawiki)).join('\n')
             : '',
-        ),
-    )
+        )
```
kelson42, line 493: There is a factorisation problem here, this should not be in the abstract class, but in the mobile class.

src/mwoffliner.lib.ts
```diff
@@ -420,17 +427,31 @@ async function execute(argv: any) {
...
     logger.log(`Found [${jsModuleDependencies.size}] js module dependencies`)
     logger.log(`Found [${cssModuleDependencies.size}] style module dependencies`)
 
-    const allDependenciesWithType = [
+    let allDependenciesWithType = [
       { type: 'js', moduleList: Array.from(jsModuleDependencies) },
       { type: 'css', moduleList: Array.from(cssModuleDependencies) },
     ]
 
+    if (dump.opts.isMobileRenderer) {
```
kelson42, line 447: this should be part of the mobile renderer

src/mwoffliner.lib.ts
```diff
@@ -401,6 +404,10 @@ async function execute(argv: any) {
     logger.info('Copying Static Resource Files')
     await saveStaticFiles(config, zimCreator)
 
+    // TODO: refactor sequence, this only needed for mobile renderer
```
kelson42, line 407: this should be indeed part of the mobilerenderer

src/util/builders/url/base.director.ts
```diff
@@ -14,24 +14,38 @@ export default class BaseURLDirector {
...
 
-  buildDesktopRestApiURL(path?: string) {
+  buildDesktopApiUrl(path?: string) {
     return urlBuilder
       .setDomain(this.baseDomain)
       .setPath(path ?? 'api/rest_v1/page/html')
       .build(true, '/')
   }
 
+  buildMobileApiUrl(path?: string) {
```
kelson42, line 31: This is `WikimediaMobile`, naming of the end-point should be respected allong all the functions.

test/e2e/mobileRenderIntegrity.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  await execa('redis-cli flushall', { shell: true })
+  const outFiles = await mwoffliner.execute(parameters)
+
+  return outFiles
+}
+
+const commonTreatmentTest = async (articleList: string, mwUrl: string) => {
+  if (!zimcheckIsAvailable || !zimdumpIsAvailable) {
```
kelson42, line 35: Please make a test once if zimcheck and zimdump are available and make the test dies otherwise. We should not test further if these tools are available.

src/util/saveArticles.ts
```diff
@@ -348,6 +313,12 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
             for (const dep of _moduleDependencies.styleDependenciesList) {
               cssModuleDependencies.add(dep)
             }
+            for (const dep of _moduleDependencies.mobileJsDependenciesList) {
```
kelson42, line 316: This is `wikimediaMobileJsDependenciesList`

src/util/saveArticles.ts
```diff
@@ -249,32 +231,35 @@ export function getArticleUrl(downloader: Downloader, dump: Dump, articleId: str
 /*
  * Fetch Articles
  */
-export async function saveArticles(zimCreator: ZimCreator, downloader: Downloader, dump: Dump, forceRender = null) {
+export async function saveArticles(zimCreator: ZimCreator, downloader: Downloader, dump: Dump, hasWikimediaMobileApi: boolean, forceRender = null) {
   const jsModuleDependencies = new Set<string>()
   const cssModuleDependencies = new Set<string>()
+  const jsMobileModuleDependencies = new Set<string>()
+  const cssMobileModuleDependencies = new Set<string>()
```
kelson42, line 238: Why are they declared separatly? Why not pushing what needs to be pushed in `jsModuleDependencies` and `cssModuleDependencies` ?

src/util/saveArticles.ts
```diff
@@ -429,5 +400,7 @@ export async function saveArticles(zimCreator: ZimCreator, downloader: Downloade
   return {
     jsModuleDependencies,
     cssModuleDependencies,
+    jsMobileModuleDependencies,
+    cssMobileModuleDependencies,
```
kelson42, line 404: `jsModuleDependencies` and `cssModuleDependencies` should be updated, please don't create other variables to return.

src/util/misc.ts
```diff
@@ -205,6 +228,12 @@ export function genHeaderScript(config: Config, js: string, articleId: string, s
   const path = isNodeModule(js) ? normalizeModule(js) : js
   return `<script ${attributes} src="${upStr}${resourceNamespace}/${jsPath(path, subDirectory)}"></script>`
 }
+export function genPCSOverrideScript(js: string) {
+  return `<script src='../-/${js}.js'></script>`
+}
+export function genPCSCOverrideCSSLink(css: string) {
+  return `<link rel="stylesheet" href="../-/${css}.css" />`
+}
```
kelson42, line 236: This is very specific to the wikimediamobiler render, and should be only there.

kelson42 (changes requested):
Tested the ZIM and it looks good in general, but the code needs IMO to be factored properly and partly simplified.

src/util/saveArticles.ts
VadimKovalenkoSNF: Done, I concated this data to the **jsDependenciesList** and **styleDependenciesList** in src/Downloader.ts.

src/util/saveArticles.ts
VadimKovalenkoSNF: During my first iteration of PCS module implementation, I separated them into distinct variables to avoid side effects and reduce the complexity of the refactoring process. Now it is updated regarding your requirements.

test/e2e/mobileRenderIntegrity.test.ts
VadimKovalenkoSNF: Done.

src/renderers/abstract.renderer.ts
VadimKovalenkoSNF: I refactored renderer architecture, there are abstract **MobileRenderer** and abstract **DesktopRenderer** that keep logic for mobile and desktop templating.

src/util/misc.ts
VadimKovalenkoSNF: Done, I replaced this to the `src/renderers/abstractMobile.render.ts`

src/util/saveArticles.ts
VadimKovalenkoSNF: Updated

src/util/builders/url/base.director.ts
VadimKovalenkoSNF: Updated

src/mwoffliner.lib.ts
VadimKovalenkoSNF: I decided to remove this condition because I can't correctly check the render type at this stage of the scrapping process.

src/mwoffliner.lib.ts
VadimKovalenkoSNF: Static files handle only in src/mwoffliner.lib.ts, not in renders.

kelson42:
@VadimKovalenkoSNF Can you please rebase?

res/templates/page.html
```diff
@@ -1,31 +1,25 @@
...
       </div>
     </div>
-  </div>
-  __ARTICLE_CONFIGVARS_LIST__
-  __ARTICLE_JS_LIST__
-</body>
-
+    __ARTICLE_CONFIGVARS_LIST__
+    __ARTICLE_JS_LIST__
+    __PCS_JS_OVERRIDE__
```
kelson42, line 23: If you need to override Mediawiki JS, then make a generic solution/naming to do that, but don't make it specific in particular with naming which is very Mediawiki specific (not MWoffliner

src/Downloader.ts
```diff
@@ -685,7 +696,27 @@ class Downloader {
 
     jsConfigVars = jsConfigVars.replace('nosuchaction', 'view') // to replace the wgAction config that is set to 'nosuchaction' from api but should be 'view'
 
-    return { jsConfigVars, jsDependenciesList, styleDependenciesList }
```
kelson42, line 688: Why we have only one `getModuleDependencies()` considering that this is render specific?

src/renderers/abstractMobile.render.ts
```diff
@@ -0,0 +1,54 @@
...
+          ? mobileJsModuleDependencies.map((oneMobJsDep) => genHeaderScript(config, oneMobJsDep, articleId, config.output.dirs.mediawiki)).join('\n')
+          : '',
+      )
+      .replace(
+        '__CSS_LINKS_MOBILE__',
+        mobileCssModuleDependencies.length !== 0
+          ? mobileCssModuleDependencies.map((oneMobCssDep) => genHeaderCSSLink(config, oneMobCssDep, articleId, config.output.dirs.mediawiki)).join('\n')
+          : '',
+      )
+      .replace('__PCS_CSS_OVERRIDE__', this.genPCSCOverrideCSSLink(config.output.pcsCssResources[0]))
```
kelson42, line 48: Please use generic naming for the placeholder

kelson42 (changes requested):
Amost good, but IMO really problematic that the modules are not handled properly in the renderers. Same for "static" files!

res/templates/page.html
VadimKovalenkoSNF: Ok, I replaced it with MW_MOBILE naming.

src/Downloader.ts
VadimKovalenkoSNF: Downloader doesn’t know anything about the chosen render, so we download all modules (both mobiles and desktop) here. More information in https://github.com/openzim/mwoffliner/issues/1923

VadimKovalenkoSNF:
1. I refactored the image size representation. Now the mobile renderer shows images by their original size, with no specific treatments involved, only CSS overriding. We can scale them down if necessary (in the sake of reducing space of the zim file sacrificing image quality). But this is more like a product decision, I'm not aware about metrics we need / can to achieve by doing so.
2. As for `nopdf` param, as I mentioned in https://github.com/openzim/mwoffliner/issues/1916#issuecomment-1746417465, it acts weird for all renders. I need confirmation whether it is a bug.

kelson42:
@VadimKovalenkoSNF The treatments of medias was and should stay the very same for all end-poibts. This is a strong requirement.

VadimKovalenkoSNF:
> The treatments of medias, in particular the pictures, was and should stay the very same for all end-points. This is a strong requirement.

Got it. Mobile renderer media files are affected by the same treatments in `src/renderers/abstract.renderer.ts` as the current desktop. I didn't do any manipulations except lazy load images because in `page/mobile-html` they were replaced by placeholder span by default.

kelson42:
@VadimKovalenkoSNF Good, most important is wepb, which the option used all the time in the zimfarm.

src/renderers/abstractMobile.render.ts
kelson42: This is not what I meant. We override anyway many other CSS, so why you need something special here? You should have one place to rewrite CSS ... and only one... with a generic name like `_CSS_OVERRIDE_`.

src/Downloader.ts
kelson42: Yes, because `Downloader` should not be specific and problem only "generic" downloading routines. Now we have here very specific (depending from end-points) code in `Downloader` which is not the right place. But I guess this will be fixed in #1923.

res/templates/page.html
kelson42: This is not what I meant. There should be only one place to override CSS and not a placeholder for each usage. A reall good name should be something like `__CSS_OVERRIDE__` or `__JS_OVERRIDE__` ... and inside you put whatever you want... here, for this end-point, the PCS related stuff.

src/mwoffliner.lib.ts
kelson42: This is always the same kind of problem (like the CSS): code has to be factored properly. You have one place to save static files and only one, why not here. But the place to specify what are the static files to save that should be in the renderer. In addition it seems you save these static mobile files... all the time, even if they are not needed. which is not clean.

kelson42 (changes requested):
Code is still not factored properly.

src/renderers/abstractMobile.render.ts
```diff
@@ -0,0 +1,54 @@
...
+          ? mobileJsModuleDependencies.map((oneMobJsDep) => genHeaderScript(config, oneMobJsDep, articleId, config.output.dirs.mediawiki)).join('\n')
+          : '',
+      )
+      .replace(
+        '__CSS_LINKS_MOBILE__',
+        mobileCssModuleDependencies.length !== 0
+          ? mobileCssModuleDependencies.map((oneMobCssDep) => genHeaderCSSLink(config, oneMobCssDep, articleId, config.output.dirs.mediawiki)).join('\n')
+          : '',
+      )
+      .replace('__WM_MOBILE_CSS_OVERRIDE__', this.genWmMobileOverrideCSSLink(config.output.wmMobileCssResources[0]))
```
kelson42, line 48: Why here, suddently, the naming convention is not respected?! It shoudl be `genWikimediaMobileOverrideCSSLink()`?

kelson42 (changes requested):

kelson42:
@VadimKovalenkoSNF Don't forget to add `WikimediaMobile` in `RENDER_LIST`

kelson42:
@VadimKovalenkoSNF As far as I can see, there is also a size problem. With this rendere, the ZIM file is almost and still 10x bigger than it should. Please compare the maxi versions (basically `--format=novid`) in https://download.kiwix.org/zim/wikipedia/

src/renderers/abstractMobile.render.ts
VadimKovalenkoSNF: I did so because for desktop and mobile static resources should have different order on the page (otherwise, overriding won't work for mobile). I created a separate template page for mobile and desktop to resolve this issue.

res/templates/page.html
VadimKovalenkoSNF: Done by creating a separate template for desktop and mobile.

src/mwoffliner.lib.ts
VadimKovalenkoSNF: I replaced modules handling to renderers themself. Now abstract desktop and abstract mobile have the ability to filter out necessary modules that come from Downloader. In the execution process now should be the exact modules we need.

src/renderers/abstractMobile.render.ts
VadimKovalenkoSNF: I renamed though I consider the name `genWikimediaMobileOverrideCSSLink()` to be cumbersome.

VadimKovalenkoSNF:
@kelson42 , I've noticed that **ab** wiki doesn't work as expected with mobile render. Though it has the support of `page/mobile-html` ([prooflink](https://ab.wikipedia.org/api/rest_v1/#/Page%20content/getContent-mobile-html)), there is an example of this article ([desktop](https://ab.wikipedia.org/wiki/%D4%A4%D1%85%D1%8B%D0%BD%D2%B7%D0%BA%D3%99%D1%8B%D0%BD%D0%BC%D0%B7%D0%B0_17)) that returns empty string in `page/mobile-html` ([link](https://ab.wikipedia.org/api/rest_v1/page/mobile-html/%D4%A4%D1%85%D1%8B%D0%BD%D2%B7%D0%BA%D3%99%D1%8B%D0%BD%D0%BC%D0%B7%D0%B0_17?redirect=false)). That causes an error and break entire scrapping process. I'll make a shallow error handling without exiting the runtime process.

kelson42:
@VadimKovalenkoSNF either this a bug by us and it has to be fixed or its a bug on wikimedia side and it has to be reported. If the backend is not able to deliver the cobtent of an article, the whole scraper should stop.

VadimKovalenkoSNF:
I noticed that `page/html` is also not working for the [article](https://ab.wikipedia.org/api/rest_v1/page/html/%D4%A4%D1%85%D1%8B%D0%BD%D2%B7%D0%BA%D3%99%D1%8B%D0%BD%D0%BC%D0%B7%D0%B0_17) mentioned above, but mwoffliner doesn't exit the scrape process and returns the page with empty with an empty body.

abkhazian-wiki-article

VadimKovalenkoSNF:
>  Please compare the maxi versions (basically --format=novid)

I'm not sure that the problem is with `novid` format, I secured it in `test/e2e/mobileRenderFormatParams.test.ts`

kelson42:
@VadimKovalenkoSNF If I got that right, this PR still needs to be fixed to have a better static content handling.

kelson42 (changes requested):
@VadimKovalenkoSNF If I got that right, this PR still needs to be fixed to have a better static content handling.

VadimKovalenkoSNF:
> @VadimKovalenkoSNF If I got that right, this PR still needs to be fixed to have a better static content handling.

Yes, static file saving is in progress.

src/mwoffliner.lib.ts
```diff
@@ -211,7 +211,8 @@ async function execute(argv: any) {
 
   MediaWiki.apiCheckArticleId = mwMetaData.mainPage
   await MediaWiki.hasCoordinates(downloader)
-  await MediaWiki.hasWikimediaDesktopRestApi()
+  await MediaWiki.hasWikimediaDesktopApi()
+  const hasWikimediaMobileApi = await MediaWiki.hasWikimediaMobileApi()
```
kelson42, line 215: In the next few lines, we have twice `await downloader.setBaseUrls(forceRender)`, looks like to be useless. Once should be enough.

kelson42:
LGTM but the js/css handling really needs to be revamped, should be handled in the renders.

PR #1912, opened 2023-09-21
Add non-root / local install method

maxslug:
Here are some helpful instructions on getting the program up and running non-root.  Tested under ubuntu 18 and 22


maxslug:
Updated PR has two commits.

PR #1914, opened 2023-09-21
README.md - local install and local redis

maxslug:
Updated instructions w/ Local Redis install too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,52 @@ this:
 --optimisationCacheUrl="https://wasabisys.com/?bucketName=my-bucket&keyId=my-key-id&secretAccessKey=my-sac"
 ```
 
+### Local (non-root) Installs
+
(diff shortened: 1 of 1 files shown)


kelson42:
I don't really understand what this README brings more. Part of it is no realated to MWoffliner, par of it is already in the README.

maxslug:
It gives instructions on how to install as non-root, especially when you can't upgrade the system version of node js.    For us that don't program javascript, having easy to follow directions is never a bad thing.

benoit74:
@maxslug thank you, but unfortunately what this documentation cover is a rare edge case and we do not have resources to maintain this / check what you wrote is valid. We appreciate you shared this but it is way beyond what we can reasonably integrate in our repo. I suggest you find some online space to store this information outside this repo, no offense.

Issue #1911, opened 2023-09-19
Collapsed tables are not viewable


Ubuntu-2020:
https://en.wikipedia.org/wiki/Agriculture#Production
In this section of the above article there are four collapsed tables at once.
Table code: wikitable collapsible collapsed.

But in file ZIM it is still not possible to expand them:
https://library.kiwix.org/viewer#wikipedia_en_all_maxi_2023-08/A/Agriculture

Please do something with tables "wikitable collapsible collapsed" so that they become visible or so that they can be expanded.

kelson42:
For some readons we have indeed tables which are still not uncollapsable. Duplicate of #1411 IMO.

benoit74:
This is indeed a duplicate of #1491. On this specific page, issue is now gone because those tables are uncolapsed by default. But it means we do not have the `hide` button and can't collapse them, so issue is still here somehow.

PR #1934, opened 2023-10-25
Unfold collapsible tables for Desktop and Mobile renders

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1911


kelson42:
@VadimKovalenkoSNF What is the status here? You have hardcoded some CSS rules, but why these rules are not in the dependences scraped by MWoffliner?

benoit74:
This is not the proper solution, we do not want to uncollapse table but to add the show/hide button and allow tables to be collapsed / uncollapsed dynamically.

Issue #1937, opened 2023-10-30
[REGRESSION] Renderer tests should have a delay between each invocation.


VadimKovalenkoSNF:
After https://github.com/openzim/mwoffliner/pull/1933 ( **Apply test coverage for all endpoins** ) has been merged, a regression with exceeded rate limit appeared while checking VisualEditor capability. At the moment of speaking, this endpoint (https://en.wikipedia.org/w/api.php?action=visualeditor&mobileformat=html&format=json&paction=parse&formatversion=2&page=MediaWiki%3ASidebar) has the response for all clients:

```
{
"error": {
  "code": "parsoid-stash-rate-limit-error",
  "info": "Stashing failed because rate limit was exceeded. Please try again later.",
  "docref": "See https://en.wikipedia.org/w/api.php for API usage. Subscribe to the mediawiki-api-announce mailing list at &lt;https://lists.wikimedia.org/postorius/lists/mediawiki-api-announce.lists.wikimedia.org/&gt; for notice of API deprecations and breaking changes."
},
  "servedby": "mw1491"
}
``` 

There is no Retry-After header in the response and the response status itself is 200. A possible solution for this is to add a delay between tests.

More info about MW Action API rate limits can be found in https://phabricator.wikimedia.org/T172293#6344271 and https://www.mediawiki.org/wiki/API:Etiquette#Request_limit

VadimKovalenkoSNF:
Phab ticket is here - https://phabricator.wikimedia.org/T350117

MatmaRex:
_(I'm missing the context, I'm just here because of https://phabricator.wikimedia.org/T350117)_

I would suggest migrating your use of the action=visualeditor action API to the new MediaWiki REST API.

The visualeditor action API is considered internal, and while we don't make breaking changes for no reason (since others, just like you, have started using it), we do not test how changes affect third-parties. It is slower because it produces metadata needed by the visual editor, but probably not needed by you.

The new MediaWiki REST API can be used like this: <https://en.wikipedia.org/w/rest.php/v1/page/MediaWiki%3ASidebar/html>. It's stable, intended for external use, has simpler output, and is faster. The output is identical to the `.content` property in the visualeditor action API. It is also available on all third-party wikis running MediaWiki 1.35 or newer.

[...]

PR #1938, opened 2023-10-30
Add delay between render tests

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1937


VadimKovalenkoSNF:
Tests will fail till https://en.wikipedia.org/w/api.php?action=visualeditor&mobileformat=html&format=json&paction=parse&formatversion=2&page=MediaWiki%3ASidebar is unblocked.

kelson42 (changes requested):
Don't put test only related stuff in the software codebase. Create a `test/const.ts`.

I also don't understand why you don't play with the speed factor?

Secure CI is green (should be right?).

kelson42:
I guess the bug is going to be fixed upstream, so should we not close this PR?

VadimKovalenkoSNF:
> I guess the bug is going to be fixed upstream, so should we not close this PR?
@kelson42 , yes, please close it.

PR #1926, opened 2023-10-11
Implement MediaWiki REST API render

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1601

PR #1947, opened 2023-11-14
Implement Mediawiki REST API render

VadimKovalenkoSNF:
Fixes https://github.com/openzim/mwoffliner/issues/1601
Successor of https://github.com/openzim/mwoffliner/pull/1926


kelson42:
@VadimKovalenkoSNF This is the last PR, hope you can fix it and put to review today.

test/e2e/forceRender.test.ts
```diff
@@ -38,22 +38,4 @@ describe('forceRender', () => {
       console.log('Zimcheck not installed, skipping test')
     }
   })
-
-  test('Scrape article from bm.wikipedia.org should throw error when using VisualEditor render', async () => {
```
VadimKovalenkoSNF, line 42: Unfortunately, this set of tests should be removed - they caused issues in CI pipeline related to memory corruption. One of the examples of this problem can be found in these logs - https://github.com/openzim/mwoffliner/actions/runs/6854615228/job/18637919124#step:7:1589
I also faced with `terminate called after throwing an instance of 'Xapian::DatabaseLockError'
Aborted (core dumped)` and `Node error process completed with exit code 139`. Probably, that happened because we forced the application to throw an error here without the right clearing of the Redis state and/or dangling process.exit(1) invocation.

src/parameterList.ts
```diff
@@ -18,7 +18,8 @@ export const parameterDescriptions = {
   keepEmptyParagraphs: 'Keep all paragraphs, even empty ones.',
   mwWikiPath: 'Mediawiki wiki base path (per default "/wiki/")',
   mwActionApiPath: 'Mediawiki API path (per default "/w/api.php")',
-  mwRestApiPath: 'Mediawiki Rest API path (per default "/api/rest_v1")',
+  mwRestApiPath: 'Wikimedia Rest API path (per default "/api/rest_v1")',
+  mwMediaWikiRestApiPath: 'Mediawiki Rest API path (per default "w/rest.php/v1/page/")',
```
VadimKovalenkoSNF, line 22: This is probably not the best naming, but I have to follow the convention that `mw` stands for Mediawiki singleton in mwoffliner, and `MediaWikiRestApiPath` should be named like this because this is how this API was named originally.

test/unit/mwApiCapabilities.test.ts
```diff
@@ -0,0 +1,85 @@
+import MediaWiki from '../../src/MediaWiki.js'
+
+describe('Checking Mediawiki capabilities', () => {
```
VadimKovalenkoSNF, line 3: This test suit should have been introduced a while ago as a convenient way to check the capabilities of different wikis, as well as testing of the specific receipts. Before trying to scrape some wiki, this is a good place to start.

PR #1990, opened 2024-02-08
add class owid-frame to mwoffliner blacklist to prevent iframes from …

tim-moody:
…displaying offline

I followed the existing code in putting a comma after this new last item in the list, though I'm not sure why.


kelson42:
@tim-moody LGTM on the principle

tim-moody:
Well, for the moment mdwiki-cacher does not work with the dev version; trying to sort it out.

PR #2025, opened 2024-05-11
Fix main page name for wikis with base URL rewritten

### LINKED ISSUE #1995: Troubleshooting minecraftwiki_zh_all recipe
Note: This is only tested on MWoffliner v1.13.0 (since all openZIM scrapers are using this version). Both the code and the config between v1.13.0 and git main differs a lot. So this needs to be tested on git main.

The following description is mostly taken from [my comment](https://github.com/openzim/zim-requests/issues/755#issuecomment-1943160450) when troubleshooting the scrape for Minecraft Wiki (zh) (openzim/zim-requests#755).

---

The scraper reports `Unable to find appropriate API end-point to retrieve article HTML` when scraping Minecraft Wiki (zh). Here is a code analysis of MWoffliner v1.13.0.

Before the scrape starts, MWoffliner checks mobile REST API, desktop REST API, and VE REST API capabilities for a specific page (parameter `testArticleId`) in `Downloader.checkCapabilities`:

https://github.com/openzim/mwoffliner/blob/e9d4113536f0eebdaabe8cc26e25ccdeeca20e32/src/Downloader.ts#L243-L263

The default value `MediaWiki:Sidebar` is never used because the value of `mwMetaData.mainPage` is passed:

https://github.com/openzim/mwoffliner/blob/e9d4113536f0eebdaabe8cc26e25ccdeeca20e32/src/mwoffliner.lib.ts#L206

The value of `mwMetaData.mainPage` comes from API. The base URL is stripped and its last part is taken. (This is a **bad** idea because different wikis have different URL rewrites.)

https://github.com/openzim/mwoffliner/blob/e9d4113536f0eebdaabe8cc26e25ccdeeca20e32/src/MediaWiki.ts#L290-L325
https://github.com/openzim/mwoffliner/blob/e9d4113536f0eebdaabe8cc26e25ccdeeca20e32/src/MediaWiki.ts#L235-L279

This works for many wikis like [English Wikipedia](https://en.wikipedia.org/w/api.php?action=query&meta=siteinfo&format=json&siprop=general|namespaces|statistics|variables|category|wikidesc), but not for [Chinese Minecraft Wiki](https://zh.minecraft.wiki/api.php?action=query&meta=siteinfo&format=json&siprop=general|namespaces|statistics|variables|category|wikidesc). The reason is that MCW-zh has URL rewrite:

```jsonc
// Wikipedia-en
"base": "https://en.wikipedia.org/wiki/Main_Page",
// MCW-zh
"base": "https://zh.minecraft.wiki/",
```

There are two ways to fix this:
1. ⭐ Set `mwMetaData.mainPage` to `entries.mainpage`, which is already included in the API result. ([MediaWiki documentation](https://www.mediawiki.org/wiki/API:Siteinfo))
   ```patch
   -const mainPage = decodeURIComponent(entries.base.split('/').pop())
   +const mainPage = entries.mainpage
   ```
2. Use the default parameter for `Downloader.checkCapabilities`:
   ```patch
   -await downloader.checkCapabilities(mwMetaData.mainPage)
   +await downloader.checkCapabilities()
   ```

I have tested both, and both worked.

TripleCamera:
This is a very small change that has been tested on my computer, so I'll just use the web editor.

Fixes #1995.

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -418,7 +418,7 @@ class MediaWiki {
     }
 
     // Base will contain the default encoded article id for the wiki.
-    const mainPage = decodeURIComponent(entries.base.split('/').pop())
+    const mainPage = entries.mainpage
     const siteName = entries.sitename
 
     // Gather languages codes (en remove the 'dialect' part)
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -417,8 +417,7 @@ class MediaWiki {
       throw new Error(`Mediawiki version ${mwVersion} not supported should be >=${mwMinimalVersion}`)
     }
 
-    // Base will contain the default encoded article id for the wiki.
-    const mainPage = entries.mainpage
+    const mainPage = entries.mainpage.replace(/ /g, '_')
     const siteName = entries.sitename
 
     // Gather languages codes (en remove the 'dialect' part)


kelson42:
@TripleCamera Tahnk your for your PR and your patience, we will finally review your PR

audiodude:
Hi @TripleCamera, thanks for the PR. Can you look into the failing tests? Also, please add additional tests for the code you changed, or otherwise verify that the code is covered by existing tests.

Thanks!

TripleCamera:
> Hi @TripleCamera, thanks for the PR. Can you look into the failing tests? Also, please add additional tests for the code you changed, or otherwise verify that the code is covered by existing tests.
> 
> Thanks!

OK. I am running the tests on my computer.

First, I need to run the `saveArticles` test on git main (`npm test -- test/unit/saveArticles.test.ts`). Unfortunately, it fails. I am investigating about this.

```
 FAIL  test/unit/saveArticles.test.ts (238.36 s)
  ● saveArticles › --customFlavour using VisualEditor renderer

    thrown: "Exceeded timeout of 40000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      151 |     })
      152 |
    > 153 |     test(`--customFlavour using ${renderer} renderer`, async () => {
          |     ^
      154 |       const { downloader, dump } = await setupScrapeClasses({ format: 'nopic' }) // en wikipedia
      155 |       downloader.setUrlsDirectors(rendererInstance, rendererInstance)
      156 |       class CustomFlavour implements CustomProcessor {

      at test/unit/saveArticles.test.ts:153:5
      at test/unit/saveArticles.test.ts:18:1
```

After fixing this, I will test my patch.

XeroAlpha:
> > Hi @TripleCamera, thanks for the PR. Can you look into the failing tests? Also, please add additional tests for the code you changed, or otherwise verify that the code is covered by existing tests.
> > Thanks!
> 
> OK. I am running the tests on my computer.
> 
> First, I need to run the `saveArticles` test on git main (`npm test -- test/unit/saveArticles.test.ts`). Unfortunately, it fails. I am investigating about this.
> 
> ```
>  FAIL  test/unit/saveArticles.test.ts (238.36 s)
>   ● saveArticles › --customFlavour using VisualEditor renderer
> 
>     thrown: "Exceeded timeout of 40000 ms for a test.
>     Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
> 
>       151 |     })
>       152 |
>     > 153 |     test(`--customFlavour using ${renderer} renderer`, async () => {
>           |     ^
>       154 |       const { downloader, dump } = await setupScrapeClasses({ format: 'nopic' }) // en wikipedia
>       155 |       downloader.setUrlsDirectors(rendererInstance, rendererInstance)
>       156 |       class CustomFlavour implements CustomProcessor {
> 
>       at test/unit/saveArticles.test.ts:153:5
>       at test/unit/saveArticles.test.ts:18:1
> ```
> 
> After fixing this, I will test my patch.

It seems that you are in an area that network is restricted. You may need to use proxy.

XeroAlpha:
By the way, this pr is just a revert for [687b97ee](https://github.com/openzim/mwoffliner/commit/687b97ee15acdcac18420adba12f33fdcbd923be).

I think it is necessary to check this commit before merging this pr.

TripleCamera:
> It seems that you are in an area that network is restricted. You may need to use proxy.

Yes, I fixed this.

Besides, thanks @XeroAlpha: The command of running a single test file should be `npm run test:pattern 'test/unit/saveArticles.test.ts'`.

Now I can pass the test **without** the patch, and fail the test **with** the patch.

I printed out the value of `mainPage`:
 - Before: `decodeURIComponent(entries.base.split('/').pop())` &rarr; `Main_Page`
 - After: `entries.mainpage` &rarr; `Main Page`

So I should submit another patch that replaces spaces with underscores in `mainPage`.

> By the way, this pr is just a revert for [687b97ee](https://github.com/openzim/mwoffliner/commit/687b97ee15acdcac18420adba12f33fdcbd923be).
> 
> I think it is necessary to check this commit before merging this pr.

Indeed. 687b97ee belongs to #1197, which "should resolve #1180". I am getting confused:
1. Why is `spaceDelimiter` a variable? In MediaWiki, page names in URLs always have spaces replaced by underscores. I have read the issue description but still couldn't understand.
2. In #1197, why is the step of "replacing spaces with underscores" removed?
3. Originally, the value of `mainPage` was `entries.mainpage.replace(/ /g, self.spaceDelimiter)`. In an intermediate commit, this was changed to `entries.mainpage.replace(/ /g, '_')`. However, in the final commit, this was changed again to `decodeURIComponent(entries.base.split('/').pop())`. Why? The intermediate version is good enough.

@kelson42 Do you know the reason?

audiodude:
> 1. Why is `spaceDelimiter` a variable? In MediaWiki, page names in URLs always have spaces replaced by underscores. I have read the issue description but still couldn't understand.

Not sure what you mean here. I don't see any references to `spaceDelimiter` at HEAD.

> 2. In [Removing concept of spaceDelimiter from MWOffliner #1197](https://github.com/openzim/mwoffliner/pull/1197), why is the step of "replacing spaces with underscores" removed?

What do you mean by "step"? I believe it was successfully removed.

> 3. Originally, the value of `mainPage` was `entries.mainpage.replace(/ /g, self.spaceDelimiter)`. In an intermediate commit, this was changed to `entries.mainpage.replace(/ /g, '_')`. However, in the final commit, this was changed again to `decodeURIComponent(entries.base.split('/').pop())`. Why? The intermediate version is good enough.

According to https://www.mediawiki.org/wiki/API:Siteinfo :

> base: The absolute path to the main page. 1.8+

The example is:

```
"base": "https://en.wikipedia.org/wiki/Main_Page",
```

So this has the main page with the spaces already replaced. I believe the code was likely set up this way specifically so that spaces didn't have to be replaced. As you pointed out in #1995, MediaWiki sometimes reports that the "base" URL is not the "Main Page", which actually makes sense logically.

Basically, I think you just need to do space replacement on your current PR and we should be good. Thanks again!

TripleCamera:
> > 1. Why is `spaceDelimiter` a variable? In MediaWiki, page names in URLs always have spaces replaced by underscores. I have read the issue description but still couldn't understand.
> 
> Not sure what you mean here. I don't see any references to `spaceDelimiter` at HEAD.

I was talking about #1180. I couldn't understand kelson's description.

> > 2. In [Removing concept of spaceDelimiter from MWOffliner #1197](https://github.com/openzim/mwoffliner/pull/1197), why is the step of "replacing spaces with underscores" removed?
> 
> What do you mean by "step"? I believe it was successfully removed.

I inspected #1197 and couldn't understand why the "replace" step was removed, perhaps because the author thought they were unnecessary? Here is an example:
```diff
@@ -96,11 +94,11 @@ class MediaWiki {
       const entries = json.query[type];
       Object.keys(entries).forEach((key) => {
         const entry = entries[key];
-        const name = entry['*'].replace(/ /g, self.spaceDelimiter);
+        const name = entry['*'];
         const num = entry.id;
         const allowedSubpages = ('subpages' in entry);
         const isContent = !!(entry.content !== undefined || util.contains(addNamespaces, num));
-        const canonical = entry.canonical ? entry.canonical.replace(/ /g, self.spaceDelimiter) : '';
+        const canonical = entry.canonical ? entry.canonical : '';
         const details = { num, allowedSubpages, isContent };
         /* Namespaces in local language */
         self.namespaces[util.lcFirst(name)] = details;
```

---

> Basically, I think you just need to do space replacement on your current PR and we should be good. Thanks again!

Done. Now I can pass the tests on my computer. Please run Actions again.

audiodude:
> I inspected #1197 and couldn't understand why the "replace" step was removed, perhaps because the author thought they were unnecessary? Here is an example:
> 
> ```diff
> @@ -96,11 +94,11 @@ class MediaWiki {
>        const entries = json.query[type];
>        Object.keys(entries).forEach((key) => {
>          const entry = entries[key];
> -        const name = entry['*'].replace(/ /g, self.spaceDelimiter);
> +        const name = entry['*'];
>          const num = entry.id;
>          const allowedSubpages = ('subpages' in entry);
>          const isContent = !!(entry.content !== undefined || util.contains(addNamespaces, num));
> -        const canonical = entry.canonical ? entry.canonical.replace(/ /g, self.spaceDelimiter) : '';
> +        const canonical = entry.canonical ? entry.canonical : '';
>          const details = { num, allowedSubpages, isContent };
>          /* Namespaces in local language */
>          self.namespaces[util.lcFirst(name)] = details;
> ```
> 
> In my opinion, the author of the pull request simply needs to replace `self.spaceDelimiter` with `'_'`. However, they removed `.replace(/ /g, self.spaceDelimiter)`, which changed the behavior of the program.

I think you're right, this is probably a bug, but I assume it doesn't affect the output ZIM. Re-running CI now.

audiodude (approved):
LGTM

kelson42:
AFAIK we can not assume that articleIds replaces spaces by undescores, this depends how Mediwiki is configured.

kelson42:
Actually, I can not find the reference anymore...

kelson42:
@audiodude This operation of replacing space by underscore is very standard. Very surprised there is no primitive already there for that.

PR #2030, opened 2024-05-23
Update README.md

audiodude:
Remove garbage from the top and change the instructions for importing mwoffliner to an `import` statement from a `require`, because the library apparently switched to ESM at some point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # MWoffliner
-65;6800;1c
+
 MWoffliner is a tool for making a local offline HTML snapshot of any
 online [MediaWiki](https://mediawiki.org) instance. It goes through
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,9 +81,10 @@ this:
 ## API
 
 MWoffliner provides also an API and therefore can be used as a NodeJS
-library. Here a stub example:
+library. Here a stub example that could go in your index.mjs file:
```
kelson42, line 84: mjs?

audiodude: Yes, rather than explaining that Node defaults to common JS style imports (`require ('foo')) and the various ways to get your Javascript file to do ESM imports, it seems simpler and more terse to just mention .mjs in passing.

PR #2043, opened 2024-06-21
Render mobile images to standard size

### LINKED ISSUE #1925: Images from page/mobile-html endpoint are too big
WikimediaMobile API in https://github.com/openzim/mwoffliner/pull/1903 relies on [page/mobile-html](https://en.wikipedia.org/api/rest_v1/#/Page%20content/getContent-mobile-html) endpoint when scraping Wikipedia articles. Most of the images that come from mobile-html are 640px in width which is not appropriate for the scrape process, because of the drastic increase of the final zim file. Check this article's images as an example - https://bm.wikipedia.org/api/rest_v1/page/mobile-html/Bamak%C9%94
Pay attention to width value in src attribute for each image, e.g https://upload.wikimedia.org/wikipedia/commons/thumb/2/20/Bamako_bridge2.jpg/640px-Bamako_bridge2.jpg has **/640px** placed there by mediawiki mobileapps service.

Related ticket in Phabricator: https://phabricator.wikimedia.org/T348529

audiodude:
Fixes #1925

In https://phabricator.wikimedia.org/T349972 a solution was given to img `data-src` attributes hardcoding images that are too big. This was the provision of an attribute `data-data-file-original-src` which has the original src size of the image. By hacking this URL, we are able to provide images that are any size we want, in this case 140px is chosen arbitrarily.

Note the use of the `WikimediaMobileRenderer.INTERNAL` object in this PR. This is to provide test access to methods that are, and should remain private. Here, we only want to test the image transformations of the renderer with a fake document. The production implementation uses `this.INTERNAL.convertLazyLoadToImages` to reference the private method, and we do the same in the test. This is the pattern that we have agreed upon in our Typescript codebase at YouTube and I think it works fairly well, though it is a bit ugly. The benefits are clear though from looking at the unit test provided with this PR. It is short and concise and fast, tests only the method being changed, and doesn't require a network connection or setting up a long chain of mwoffliner configs and objects.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,7 @@
         "split-by-grapheme": "^1.0.1",
         "swig-templates": "^2.0.3",
         "typescript": "^4.9.4",
+        "url-join": "^5.0.0",
         "utf8-binary-cutter": "^0.9.2",
         "webp-hero": "0.0.2",
         "yargs": "^17.7.1"
@@ -18805,6 +18806,14 @@
       "resolved": "https://registry.npmjs.org/urix/-/urix-0.1.0.tgz",
       "integrity": "sha512-Am1ousAhSLBeB9cG/7k7r2R0zj50uDRlZHPGbazid5s9rlF1F/QKYObEKSIunSjIOkJZqwRRLpvewjEkM7pSqg==",
       "deprecated": "Please see https://github.com/lydell/urix#deprecated"
+    },
+    "node_modules/url-join": {
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/url-join/-/url-join-5.0.0.tgz",
+      "integrity": "sha512-n2huDr9h9yzd6exQVnH/jU5mr+Pfx08LRXXZhkLLetAMESRj+anQsTAh940iMrIetKAmry9coFuZQ2jY8/p3WA==",
+      "engines": {
+        "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+      }
     },
     "node_modules/url-parse-lax": {
       "version": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "split-by-grapheme": "^1.0.1",
     "swig-templates": "^2.0.3",
     "typescript": "^4.9.4",
+    "url-join": "^5.0.0",
     "utf8-binary-cutter": "^0.9.2",
     "webp-hero": "0.0.2",
     "yargs": "^17.7.1"
--- a/src/renderers/wikimedia-mobile.renderer.ts
+++ b/src/renderers/wikimedia-mobile.renderer.ts
@@ -1,4 +1,5 @@
 import * as domino from 'domino'
+import urlJoin from 'url-join'
 import * as logger from '../Logger.js'
 import { MobileRenderer } from './abstractMobile.render.js'
 import { getStrippedTitleFromHtml } from '../util/misc.js'
@@ -33,7 +34,7 @@ export class WikimediaMobileRenderer extends MobileRenderer {
         const mobileHTML = domino.createDocument(data)
         const finalHTMLMobile = await this.pipeMobileTransformations(
           mobileHTML,
-          this.convertLazyLoadToImages,
+          this.INTERNAL.convertLazyLoadToImages,
           this.removeEditContainer,
           this.removeHiddenClass,
           async (doc) => {
@@ -91,7 +92,7 @@ export class WikimediaMobileRenderer extends MobileRenderer {
     return doc
   }
 
-  private convertLazyLoadToImages(doc: DominoElement) {
+  private convertLazyLoadToImagesImpl(doc: DominoElement) {
     const protocol = 'https://'
     const spans = doc.querySelectorAll('.pcs-lazy-load-placeholder')
 
@@ -100,7 +101,14 @@ export class WikimediaMobileRenderer extends MobileRenderer {
       const img = doc.createElement('img') as DominoElement
 
       // Set the attributes for the img element based on the data attributes in the span
-      img.src = protocol + span.getAttribute('data-src')
+
+      // The data-data-file-original-src attribute is the URL of the image that was used in the original article.
+      // It is preferred over the data-src attribute, which is a "mobile" image that may be scaled up to 320px
+      // or 640px in order to be "full width" on mobile devices. However, if the mobile API didn't scale the
+      // image up, then the data-data-file-original-src attribute will be missing, and we should use the data-src.
+      // See https://github.com/openzim/mwoffliner/issues/1925.
+      const imgSrc = span.getAttribute('data-data-file-original-src') || span.getAttribute('data-src')
+      img.src = urlJoin(protocol, imgSrc)
       img.setAttribute('decoding', 'async')
       img.width = span.getAttribute('data-width')
       img.height = span.getAttribute('data-height')
@@ -147,5 +155,9 @@ export class WikimediaMobileRenderer extends MobileRenderer {
     })
 
     return doc
+  }
+
+  public readonly INTERNAL = {
+    convertLazyLoadToImages: this.convertLazyLoadToImagesImpl,
   }
 }
--- a/test/unit/renderers/mobile.renderer.test.ts
+++ b/test/unit/renderers/mobile.renderer.test.ts
@@ -96,5 +96,34 @@ describe('mobile renderer', () => {
       expect(imgs[0].src).toEqual('https://upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Bamako_et_fleuve_Niger.jpg/250px-Bamako_et_fleuve_Niger.jpg')
       expect(imgs[1].src).toEqual('https://upload.wikimedia.org/wikipedia/commons/thumb/2/20/Bamako_bridge2.jpg/250px-Bamako_bridge2.jpg')
     })
+
+    test('it uses the data-src when data-data-file-original-src is not available', async () => {
+      const test_window = domino.createWindow(
+        `
+        <span
+          class="mw-file-element gallery-img pcs-widen-image-override pcs-lazy-load-placeholder pcs-lazy-load-placeholder-pending"
+          style="width: 150px"
+          data-class="mw-file-element gallery-img pcs-widen-image-override"
+          data-src="//upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/150px-BMW.svg.png"
+          data-srcset="//upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/225px-BMW.svg.png 1.5x, //upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/300px-BMW.svg.png 2x"
+          data-width="150"
+          data-height="150"
+          data-alt="Logo used in vehicles since 1997"
+          data-data-file-width="1015"
+          data-data-file-height="1015"
+          ><span style="padding-top: 100%">
+        `,
+        'http://en.wikipedia.org/api/rest_v1/page/mobile-html/BMW',
+      )
+      const mobileRenderer = new WikimediaMobileRenderer()
+
+      const actual = mobileRenderer.INTERNAL.convertLazyLoadToImages(test_window.document)
+      const spans = actual.querySelectorAll('.pcs-lazy-load-placeholder')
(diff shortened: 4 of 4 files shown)


kelson42 (commented):
@audiodude The image is retrieved from the right place but I don't understand why you change the resolution. It's not our role to change the resolution.

It's seems you think the size given is wrong, but I'm not sure about the standardsyou refer too. I also don't understand hoe 140px width could be the right width!

audiodude:
Sorry, I guess I was confused about the requirements.

If you look at the images retrieved from the sample endpoint, which I included in the test, you see that the first has `data-width="320"` and:

```
data-src="//upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Bamako_et_fleuve_Niger.jpg/320px-Bamako_et_fleuve_Niger.jpg"
```

So the `data-src` already has the right size image. Same for the other image:

```
data-width="640"
data-src="//upload.wikimedia.org/wikipedia/commons/thumb/2/20/Bamako_bridge2.jpg/640px-Bamako_bridge2.jpg"
```

Since I read in the reports that 640px was "too big/wide", I thought I had to resize the images. 140px was just an arbitrary value I chose.

kelson42:
@audiodude Key approach (to see if we are doing good is) IMHO is to compare to what we get currently with the `mobile-section` end-point in production on the Zimfarm.

audiodude:
Yes looking at the HTML of https://bm.wikipedia.org/wiki/Bamak%C9%94 I see:

```
<img src="//upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Bamako_et_fleuve_Niger.jpg/250px-Bamako_et_fleuve_Niger.jpg" decoding="async" width="250" height="188" class="mw-file-element" srcset="//upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Bamako_et_fleuve_Niger.jpg/375px-Bamako_et_fleuve_Niger.jpg 1.5x, //upload.wikimedia.org/wikipedia/commons/thumb/8/8f/Bamako_et_fleuve_Niger.jpg/500px-Bamako_et_fleuve_Niger.jpg 2x" data-file-width="600" data-file-height="450">
```

Which has a width of 250, and that's what we see in `data-data-file-original-src`.

So maybe the solution is to just set the `.src` to this value and not have to transform any URLs?

kelson42:
> So maybe the solution is to just set the `.src` to this value and not have to transform any URLs?

If you read the issue (here and on Phabricator) this has been clearly what I requested... and what I expected this PR to do...

But it we do only this we will:
* Increase significantly the size of our ZIM files
* Might have ugly visual effect
* ... and an other side effect will be to recompute the whole cache of pictures....

I wonder how bad would these 3 points above be... and how important/urgent is this to fix.

My (bad) feeling is that it is urgent and we need to either
* Request the WMF to deliver "mobile friendly" width and height to recompute the right size URL (I don't like this as I said previously)
* Request the WMF to deliver the "mobile friendly" resolution URL (I don't like because this is not what I requested first)
* As what the WMF to deliver the logic used to compute the "mobile friendly" resolution to resize on the the fly our pictures (hate as well as sounds fragile)

What do you think?

audiodude:
I think there is a confusion about what "mobile-friendly" means. It seems the Mediawiki software is using 640px and 320px images as "mobile-friendly" because they will take up the full width of the phone, which arguably looks better when you're scrolling an article in vertical format I don't think there is any concept here, in the original software, that "mobile-friendly" means "saving bandwidth".

I don't think using the `data-data-file-original-src` will increase the size of the ZIMs. The Mediawiki code you linked to [here](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/thumbnail.js#L29) will never upscale an image past it's "original file size", but it will make it 640px or 320px if that file size is greater than those options. As pointed out, this is potentially greater than the size that the image was on the page (in the example above, 250px).

So I think using the original page image will:

1. Actually decrease ZIM size compared to using `data-src`
2. More closely match what the article looks like on the website (no impact to "visual effect")

As far as recomputing the cache of pictures, that is definitely a side effect, but seems necessary.

I just don't think WMF or the Mediawiki maintainers have any more idea what a "mobile-friendly" size is than we do, or at least they have a very different definition of it. That's why I just guessed 140px.

kelson42:
> I think there is a confusion about what "mobile-friendly" means. It seems the Mediawiki software is using 640px and 320px images as "mobile-friendly" because they will take up the full width of the phone, which arguably looks better when you're scrolling an article in vertical format. I don't think there is any concept here, in the original software, that "mobile-friendly" means "saving bandwidth".

Yes, the "mobile-section" new API we use is conceived for The Wikipedia mobile app. I believe the mobile app does on the fly image resising (probably depending of the device screen resolution).

> I don't think using the `data-data-file-original-src` will increase the size of the ZIMs. The Mediawiki code you linked to [here](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/thumbnail.js#L29) will never upscale an image past it's "original file size", but it will make it 640px or 320px if that file size is greater than those options. As pointed out, this is potentially greater than the size that the image was on the page (in the example above, 250px).

You might be right, but reference should be current generated ZIM files, which use an other API. That should be actually really easy to do.

> So I think using the original page image will:
> 
> 1. Actually decrease ZIM size compared to using `data-src`
> 2. More closely match what the article looks like on the website (no impact to "visual effect")
> 
> As far as recomputing the cache of pictures, that is definitely a side effect, but seems necessary.
> 
> I just don't think WMF or the Mediawiki maintainers have any more idea what a "mobile-friendly" size is than we do, or at least they have a very different definition of it. That's why I just guessed 140px.

Yes, at least their definition of "mobile-friendly" has significantly evolved between the "mobile-section" API and the "mobile-html". To me "mobile-friendly" means "smaller than desktop", somlike it is currently with the "mobile section" API.

audiodude:
> You might be right, but reference should be current generated ZIM files, which use an other API. That should be actually really easy to do.

So do I just compare https://download.kiwix.org/zim/wikipedia/wikipedia_bm_all_maxi_2024-06.zim with a new ZIM that I make from scratch using the updated code?

kelson42:
> > You might be right, but reference should be current generated ZIM files, which use an other API. That should be actually really easy to do.
> So do I just compare https://download.kiwix.org/zim/wikipedia/wikipedia_bm_all_maxi_2024-06.zim with a new ZIM that I make from scratch using the updated code?

Yes

> Is there a way to see the params that were used to make that ZIM? I don't see a recipe on https://farm.zimit.kiwix.org/recipes probably because I don't have permssions?

https://farm.openzim.org/recipes/wikipedia_bm_all

Important to consider is the --webp option.

kelson42:
@audiodude `--forceRender` might have to be used (actually this is something we need to check as well if the priorisations of API is correct)

audiodude:
Here is the command I used:

```
$ node ./lib/cli.js --webp --mwUrl="https://bm.wikipedia.org" --format="nopic:nopic" --format="novid:maxi" --osTmpDir="/dev/shm" --adminEmail="contact@kiwix.org" --outputDirectory="output" --customZimFavicon="https://drive.farm.openzim.org/wikipedia_all/favicon-48x48.png" --customZimDescription="offline version of Wikipedia in Bambara" --publisher="openZIM" --forceRender=WikimediaMobile
```

And here is the resulting file size:

```
$ du output/wikipedia_bm_all_maxi_2024-06.zim 
41700	output/wikipedia_bm_all_maxi_2024-06.zim
```

So no significant impact on file size using the Mobile Renderer with getting the image src from `data-data-file-original-src`

audiodude:
I've updated the code, PTAL

kelson42:
@audiodude Code LGTM, but CI is red. I will rebase.

audiodude:
> @audiodude Code LGTM, but CI is red. I will rebase.

It looks like the tests are flaky, one of them is timing out sometimes:

```
FAIL test/unit/mwApiCapabilities.test.ts (10.2 s)
  ● Checking Mediawiki capabilities › test capabilities of pokemon.fandom.com with RestApi receipt

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      67 |   })
      68 |
    > 69 |   test('test capabilities of pokemon.fandom.com with RestApi receipt', async () => {
         |   ^
      70 |     MediaWiki.base = 'https://pokemon.fandom.com/'
      71 |     MediaWiki.wikiPath = '/'
      72 |     MediaWiki.restApiPath = 'rest.php'

      at test/unit/mwApiCapabilities.test.ts:69:3
      at test/unit/mwApiCapabilities.test.ts:3:1
```

audiodude:
> > @audiodude Code LGTM, but CI is red. I will rebase.
> 
> It looks like the tests are flaky, one of them is timing out sometimes:
> 
> ```
> FAIL test/unit/mwApiCapabilities.test.ts (10.2 s)
>   ● Checking Mediawiki capabilities › test capabilities of pokemon.fandom.com with RestApi receipt
> 
>     thrown: "Exceeded timeout of 5000 ms for a test.
>     Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
> 
>       67 |   })
>       68 |
>     > 69 |   test('test capabilities of pokemon.fandom.com with RestApi receipt', async () => {
>          |   ^
>       70 |     MediaWiki.base = 'https://pokemon.fandom.com/'
>       71 |     MediaWiki.wikiPath = '/'
>       72 |     MediaWiki.restApiPath = 'rest.php'
> 
>       at test/unit/mwApiCapabilities.test.ts:69:3
>       at test/unit/mwApiCapabilities.test.ts:3:1
> ```

#2045

audiodude:
This failure might be an actual bug because it's failing zimcheck for a MobileRenderer scrape.

audiodude:
Looks like several of the images in the BMW article don't have `data-data-original-file-src`:

```
<span
                            class="mw-file-element gallery-img pcs-widen-image-override pcs-lazy-load-placeholder pcs-lazy-load-placeholder-pending"
                            style="width: 150px"
                            data-class="mw-file-element gallery-img pcs-widen-image-override"
                            data-src="//upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/150px-BMW.svg.png"
                            data-srcset="//upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/225px-BMW.svg.png 1.5x, //upload.wikimedia.org/wikipedia/commons/thumb/4/44/BMW.svg/300px-BMW.svg.png 2x"
                            data-width="150"
                            data-height="150"
                            data-alt="Logo used in vehicles since 1997"
                            data-data-file-width="1015"
                            data-data-file-height="1015"
                            ><span style="padding-top: 100%">
...

<span
                            class="mw-file-element gallery-img pcs-widen-image-override pcs-lazy-load-placeholder pcs-lazy-load-placeholder-pending"
                            style="width: 150px"
                            data-class="mw-file-element gallery-img pcs-widen-image-override"
                            data-src="//upload.wikimedia.org/wikipedia/commons/thumb/b/b2/Squircle_%28120290377%29.jpg/150px-Squircle_%28120290377%29.jpg"
                            data-srcset="//upload.wikimedia.org/wikipedia/commons/thumb/b/b2/Squircle_%28120290377%29.jpg/225px-Squircle_%28120290377%29.jpg 1.5x, //upload.wikimedia.org/wikipedia/commons/thumb/b/b2/Squircle_%28120290377%29.jpg/300px-Squircle_%28120290377%29.jpg 2x"
                            data-width="150"
                            data-height="147"
                            data-alt="The logo on a BMW car"
                            data-data-file-width="1642"
                            data-data-file-height="1614"
                            >
```

I guess we have to fall back to `src` in this case?

audiodude:
It looks like in the mobile API source code, that if the image didn't get resized, it doesn't have a `data-data-file-original-src` attribute. So in that case it should be fine to use the `.src`.

https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/thumbnail.js#L96

kelson42:
@audiodude 

> It looks like in the mobile API source code, that if the image didn't get resized, it doesn't have a `data-data-file-original-src` attribute. So in that case it should be fine to use the `.src`.
> 
> https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/thumbnail.js#L96

I guess this PR should be patch then!?

audiodude:
Yes just updated. Tests pass locally.

audiodude:
Force pushed to consolidate commits, PTAL

Issue #2057, opened 2024-07-07
Architecture: Unit tests are non-hermetic


audiodude:
Many (most?) of the unit tests actually make network requests to various MediaWiki projects like English Wikipedia. This leads them to be inherently flaky, because changes to data (not the API, just data) in these wikis can cause the tests to fail. See [here](https://medium.com/geekculture/hermetic-unit-testing-8c49276e3acd) for a simple and direct overview of the problem and the value of hermetic tests.

It is understood that mwoffliner needs to be tested against live wikis in order to prevent regressions. However, tests should be segregated so that any test which requires a live network connection is moved into e2e tests.

We should be able to test specific behaviors like "This utility method returns normalized titles" without downloading entire wiki pages.

Fixing this would also likely lead to architectural improvements, because we would be forced to decouple various modules more and use [dependency injection](https://www.jamesshore.com/v2/blog/2006/dependency-injection-demystified).

It would also make the unit tests more speedy, because the network requests would not be necessary.

E2E tests could still be run in CI and on demand, when necessary.

kelson42:
I can only agree with you, but not that easy to mockup MediWiki with different configurations. That said we have an easy way to deply MediaWiki with https://github.com/offspot/mediawiki-docker

audiodude:
Yes, one of the ways to make the unit tests hermetic would be to have our own MediaWiki installations.

But I was thinking more along the lines of https://github.com/openzim/mwoffliner/pull/2043/files

Where the unit test just tests one aspect of the code and does not connect to any MediaWiki installation at all. I believe unit tests should test individual aspects and functions, instead of "Download this whole list of articles and then test if this one portion of it worked", which seems to be the current architecture.

rgaudin:
## GSoC guidance

MWoffliner’s test suite currently relies heavily on live HTTP requests to Wikipedia and other MediaWiki sites, resulting in 30 minutes test runs that are flaky, resource-intensive, and unsuitable for CI/CD pipelines. This project aims to rewrite the test infrastructure to be hermetic—using mocked responses, local test fixtures, and in-memory MediaWiki instances where needed.

[...]

PR #2059, opened 2024-07-07
Workflows: Add scheduled e2e step and refactor ENV vars

### LINKED ISSUE #2058: e2e tests should run periodically and generate failure emails
The mwoffliner project has many e2e tests that run in CI and test that the project can actually properly process various live wikis and their data. This is great!

In fact, many of the unit tests themselves are actually e2e tests (see #2057)

However, since the external data itself could cause the tests to fail, and we expect this to happen fairly frequently, it would be useful to have the e2e tests run periodically, perhaps once per day (configurable). This would allow us to catch regressions early and not be "surprised" when trying to submit code and having e2e tests inexplicably failing.

Not sure if GitHub actions has this capability but it seems likely.

For example, this was a problem when trying to submit #2055, since it was blocked by the unrelated failure in #2056.

audiodude:
FIxes #2058 

This PR refactors the `ci.yml` file to introduce a workflow step that only runs when the workflow is triggered by cron/scheduled. If the workflow is not triggered by cron (eg by PR), the `test:e2e` target is not run. So, alternately, when the workflow is triggered by a PR, the normal tests, with coverage, are run instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,10 +5,17 @@ on:
   push:
     branches:
       - main
+  schedule:
+    - cron: '17 12 * * *' # Run every day at 12:17 UTC
+
+env:
+  REDIS: 'redis://127.0.0.1:6379'
+  S3_URL: ${{ secrets.S3_URL }}
+  ZIMCHECK_PATH: $( find .. -name zimcheck )
+  ZIMDUMP_PATH: $( find .. -name zimdump )
 
 jobs:
-  build:
-
+  ci-test:
     runs-on: ubuntu-22.04
 
     strategy:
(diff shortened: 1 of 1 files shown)


audiodude:
> @audiodude I'm good with this, but I see no value to remove testing e2e on PR. I think this is an error and unit testing coverage is really not that good.
> 
> If a PR CI fails because of one of the e2e tests, then we should fix the e2e test first, I'm even ready to accept if this is in the current PR. If the PR is urgent, then we can event exceptionally shortcut the CI.

If you look at the code:

```
      - name: Running all tests (w/coverage)
        if: ${{ github.event_name != 'schedule' }}
        run: npm run codecov
```

The target `npm run codecov` runs all tests, with coverage. So the behavior on PR is unchanged.

However, the decision to only run the e2e tests on a schedule is probably flawed due to #2057. We should probably run all tests on the schedule, because many (most?) of the "unit" tests are actually e2e tests.

audiodude:
I've updated the PR to run all tests on the schedule (without coverage). PTAL.

I've also renamed 'build' to 'test-ci' because it's not a build, it's running the CI tests.

Note: One thing I ran into when researching this PR is that schedule build notification emails will only get sent to me (or in the future, the last person who committed a change to `ci.yml`). If we want to send scheduled test failure emails more broadly, I believe we have to manually setup some kind of `actions/foo-some-guy/send-email` action and provide SMTP credentials and lists of email addresses in the GitHub secrets.

PR #2063, opened 2024-07-15
Clean wepb in s3

kelson42:
This is mostly clarification and documentation around the image (cache) mgmt.


kelson42:
@audiodude This PR has two objectives:
* Clarify a bit what is going on around image download/update (with the cache)
* Letting you having a look to it to check it looks good (before we refill the cache). I believe the block `mwResp.status === 304` could be done better but don't know how :(

kelson42:
@audiodude There is no way tomdo better for the second bukket point?

kelson42:
@audiodude Do you in particular could explain to me please what does this line? Does it copy only the headers or does it copy the body as well?
https://github.com/openzim/mwoffliner/pull/2063/files#diff-b37958e03ae0b3fd7818115e6506d5ac5e8e10df266bee9ce4d1606838d83bc2R564

audiodude:
> @audiodude Do you in particular could explain to me please what does this line? Does it copy only the headers or does it copy the body as well? https://github.com/openzim/mwoffliner/pull/2063/files#diff-b37958e03ae0b3fd7818115e6506d5ac5e8e10df266bee9ce4d1606838d83bc2R564

I believe it copies everything *but* the body.

```
(({ Body, ...o }) => o)(s3Resp)
```

So it's creating an anonymous function that takes an object, and destructures it into two parts, `Body` and `...o` which I believe is "everything else. Then it returns the everything else. This anonymous function is immediately called on the s3Resp object.

audiodude:
Just wondering, do you want me to correct all the little grammatical mistakes in the comments?

kelson42:
> Just wondering, do you want me to correct all the little grammatical mistakes in the comments?

You can, but I obviously don't expect that!

PR #2093, opened 2024-10-31
Increase S3 socket limit to 250

### LINKED ISSUE #2092: Too many requests are started in parallel
In https://farm.openzim.org/pipeline/53d49603-fd60-4933-b7f3-5b8130b29805/debug, the task is finally aborted due to skew between request time and server time.

What I see in the logs are many message like `@smithy/node-http-handler:WARN - socket usage at capacity=50 and 115 additional requests are enqueued.`.

To me this is the root cause of the problem, and it is a software bug. I understand this message as the software has started many HTTP requests and they are getting delayed. And hence the skew time between request time and server time from my PoV. Classic async issue in Node.JS from my PoV, where one can easily get bitten by all this asynchronous logic which has many advantages but some drawbacks.

And of course this problem occur randomly since it depends on your connectivity with the upstream. Travis connectivity to AWS is probably better than the connectivity from mwoffliner1 / wikimedia foundation datacenter.

I'll open another issue to track this problem.

audiodude:
Fixes #2092

This should increase the number of underlying sockets that the AWS S3 library uses. Given that the default is 50, and the error mentioned 113 requests waiting, 250 should be sufficient.

Tested on my local machine, I was able to create a bm ZIM without problem.

### DIFF
--- a/src/S3.ts
+++ b/src/S3.ts
@@ -3,6 +3,9 @@ import * as logger from './Logger.js'
 import { Readable } from 'stream'
 import { publicIpv4 } from 'public-ip'
 
+import * as https from 'https'
+import { NodeHttpHandler } from '@smithy/node-http-handler'
+
 interface BucketParams {
   Bucket: string
   Key: string
@@ -34,6 +37,11 @@ class S3 {
   }
 
   public async initialise() {
+    const requestHandler = new NodeHttpHandler({
+      httpsAgent: new https.Agent({
+        maxSockets: 250,
+      }),
+    })
     const s3UrlBase: any = new URL(this.url)
     this.s3Handler = new S3Client({
       credentials: {
@@ -43,6 +51,7 @@ class S3 {
       endpoint: s3UrlBase.href,
       forcePathStyle: s3UrlBase.protocol === 'http:',
       region: this.region,
+      requestHandler,
     })
 
     return this.bucketExists(this.bucketName)


benoit74:
For me this solution is just an escape, which does not really solve the real problem which seems to me to be rather "why do we start so many parallel connections to S3?". Increasing the number of parallel connections can only lead to performance issues at one point or another, and in no way allows the scraper to run faster. Is it so hard to limit the number of parallel connections to S3?

kelson42:
@benoit74 I'm not sure to understand what would be a better architecture. Do you suggest some kind of pooling there with S3 workers?

benoit74:
I think all parallel operations in scrapers should be based on a fixed size worker pool, yes.

What we have in most (all?) Python scrapers (but I might be completely off-topic) is that whenever we are doing things in parallel, we have a limited set of workers (typically 4, 10, ...) working of a queue of tasks. E.g. for storing pictures in a ZIM, we have one task per picture to either retrieve from S3 or download and optimize. And 4 to 10 workers, each working on one task at a time. While it does not allows to be very adaptive (depending on the IO or CPU nature of work to do, the optimal number of worker might change), it allows to avoid loosing time switching context.

I know that Node.JS paradigms are different and they favor asynchronous "by-design", but this has been meant for web servers where you have to process web request in parallel, this is not something you can control. And it is generally better to very slowly progress on all connections (to avoid timeouts) rather than queuing them and answering after the client timeout, having loosed your effort. Here in a scraper, we can totally choose our level of parallelism and we do not have any timeout on tasks to perform.

What we've seen in Python scraper is that passed a certain level of parallelism, the bigger the work pool the slower the application because it induces too much context switching at all levels.

audiodude:
> Is it so hard to limit the number of parallel connections to S3?

Unfortunately, yes, it is.

The mwoffliner is designed to process articles as soon as they are available. Part of this processing includes saving the images to S3, which is done simply by dispatching an async task to do so. There is no concept of an mwoffliner "worker".

Trying to manage the rate at which mwoffliner uploads images would require rewriting major parts of the system, using architecture and abstractions that we don't even have.

I recommend merging this to fix the immediate problem, and revisiting whether we want to rearchitect/rewrite mwoffliner.

kelson42:
> I recommend merging this to fix the immediate problem, and revisiting whether we want to rearchitect/rewrite mwoffliner.

I agree with this conclusion. If we have the proof that we face a bootleneckt here, we can reduce the number of parallelisation in general.

PR #2097, opened 2024-11-04
Remove inline styles from <section> tags returned from mobile-html

audiodude:
As stated in the comment, we just clobber the whole style tag. This might need to be revisited in the event that the endpoint starts returning something useful there.

### DIFF
--- a/src/renderers/wikimedia-mobile.renderer.ts
+++ b/src/renderers/wikimedia-mobile.renderer.ts
@@ -37,6 +37,7 @@ export class WikimediaMobileRenderer extends MobileRenderer {
           this.INTERNAL.convertLazyLoadToImages,
           this.removeEditContainer,
           this.removeHiddenClass,
+          this.INTERNAL.unhideSections,
           async (doc) => {
             const { finalHTML, subtitles, mediaDependencies } = await super.processHtml(
               doc.documentElement.outerHTML,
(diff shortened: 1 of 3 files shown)


Jaifroid:
@audiodude Thanks, I'll take a look!

src/renderers/wikimedia-mobile.renderer.ts
```diff
@@ -130,6 +131,16 @@ export class WikimediaMobileRenderer extends MobileRenderer {
     return doc
   }
 
+  private unhideSectionsImpl(doc: DominoElement) {
+    const sections = doc.querySelectorAll('section')
+    Array.from(sections).forEach((section: DominoElement) => {
+      // Domino doesn't allow us to easily manipulate specific styles. Rather than trying to parse
+      // the style attribute and remove display: none, we just clobber the whole thing.
+      section.style = ''
```
Jaifroid, line 139: Just wondering if you tried `section.style.display = ''`. Your comment suggests you think this might not work, and I don't know Domino, but it seems like a strange omission if Domino is designed to implement the DOM spec in NodeJS. I only suggest trying this (if you didn't already) because it might save headaches in the future in case some other attribute other than display is added to the style element. However, of course your code has the same practical effect given the cases we're currently aware of.

Jaifroid (approved):
Apart from the small comment about potentially trying a more targeted approach, this looks like it should do the job! I leave that up to you!

src/renderers/wikimedia-mobile.renderer.ts
audiodude: Yes unfortunately that was the first thing I tried. As you guessed, it's a Domino limitation. I read through the Domino source code to figure that out. They don't support properties of the `style` property. It's just a string.

PR #2101, opened 2024-11-14
Re-work image resizing

### LINKED ISSUE #1925: Images from page/mobile-html endpoint are too big

### LINKED ISSUE #2071: ZIM for ab_all_maxi has different sizes between 1.13 and 1.14
The ZIM scraped in July 2024 by 1.14 has a different size than the one scraped in June 2024 by 1.13:

```
wikipedia_ab_all_maxi_2024-06.zim             2024-06-17 02:21   26M   
wikipedia_ab_all_maxi_2024-07.zim             2024-07-22 20:56   36M
```

Oddly, this is the opposite problem as #2070. We don't yet know what the issue might be.

audiodude:
Fixes #1925
Fixes #2071

Re-work image sizing algorithm. It now enforces a maximum image width of 320px.

This value is chosen because it is a reasonable size on both mobile and desktop, but saves bandwidth compared to "full size" images (which was what the code might have been grabbing from `data-data-original-file-src` before). This should give us a nice tradeoff between efficiency and quality. We can always adjust this value later, of course, but the logic for retrieving and resizing the image URLs remains the same.

Of note, it takes the smallest of:

* The `src` attribute set on the `<span>` itself that will become the image
* The `data-data-original-file-src` attribute, of the original size the image was in the article
* 320px, provided it can look at one of the previous two URLs and find a URL that is hackable to set the image size.

Tests have been added/adjusted.

EDIT:

Additionally, based on feedback in this thread, I have modified the algorithm to refrain from scaling *any* dimension smaller than 320px. So for large panorama images such as in the enwiki Paris article, the image does not get destroyed because it is so wide (because scaling to a width of 320px leaves a height of 50px)

### DIFF
--- a/src/renderers/wikimedia-mobile.renderer.ts
+++ b/src/renderers/wikimedia-mobile.renderer.ts
@@ -112,7 +112,7 @@ export class WikimediaMobileRenderer extends MobileRenderer {
 
     // Calculate the ratio so we know if we're scaling down in the width or height dimension.
     const widthHeightRatio = preparedData.width / preparedData.height
-    let scaleUsingHeight = widthHeightRatio > 1.0
+    const scaleUsingHeight = widthHeightRatio > 1.0
 
     // The data-data-file-original-src attribute is the URL of the image that was used in the original article.
     // It is preferred over the data-src attribute, which is a "mobile" image that may be scaled up in order to
--- a/src/renderers/wikimedia-mobile.renderer.ts
+++ b/src/renderers/wikimedia-mobile.renderer.ts
@@ -8,7 +8,13 @@ import { RenderOpts, RenderOutput } from './abstract.renderer.js'
 type PipeFunction = (value: DominoElement) => DominoElement | Promise<DominoElement>
 
 const THUMB_WIDTH_REGEX = /\/(\d+)px-[^/]+$/
-const THUMB_MAX_WIDTH = 320
+const THUMB_MAX_DIMENSION = 320
+
+declare interface ImageMetadata {
+  src: string | null
+  width: number
+  height: number
+}
 
 // Represent 'https://{wikimedia-wiki}/api/rest_v1/page/mobile-html/'
 export class WikimediaMobileRenderer extends MobileRenderer {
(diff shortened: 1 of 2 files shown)


audiodude:
Fixes #1925. Fixes #2071.

audiodude:
With this change, we see the following:

```
14272	output/wikipedia_bm_all_maxi_2024-11.113.zim
16536	output/wikipedia_bm_all_maxi_2024-11.114.zim
```

Scraping of BM with these settings results in a ZIM that is about 2 MB, or 15% bigger.

audiodude:
By the way, I spent way too long looking through the [repo here](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/605640c5deb3dffb3da9033527c5df98fdb21156/lib/thumbnail.js) and even posted a [phabricator ticket](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/605640c5deb3dffb3da9033527c5df98fdb21156/lib/thumbnail.js), but my intuition is that they were just setting max sizes and we should do the same.

Jaifroid:
Just checking, the current max appears to be 264px (looking at full English Wikipedia). I quite like the larger 320px, and it is something users have been requesting... I suppose this needs weighing up carefully. 15% on ~100GB doesn't seem like too high a price to pay to me...

audiodude:
> Just checking, the current max appears to be 264px (looking at full English Wikipedia). I quite like the larger 320px, and it is something users have been requesting... I suppose this needs weighing up carefully. 15% on ~100GB doesn't seem like too high a price to pay to me...

How did you find that value? Just by finding some 264 width images and assuming there's nothing bigger?

Jaifroid:
> How did you find that value? Just by finding some 264 width images and assuming there's nothing bigger?

Oh sorry, I did that a bit fast, and having checked, I've found some at 280px down the right-hand side, so no, no guarantee that that's the max, more an average! There are some much wider images centred in some pages and we'll have to be careful not to clobber those. For example here in the "Paris" article:

image

audiodude:
Okay, when investigating the Paris panorama, I found a bug, which I fixed and added a test for above. Now the sizes look like this:

```
14272	output/wikipedia_bm_all_maxi_2024-11.113.zim
14532	output/wikipedia_bm_all_maxi_2024-11.114-280.zim
15364	output/wikipedia_bm_all_maxi_2024-11.114-320.zim
```

That's a 1.8% increase for 280px width images, and a 7.65% increase for 320px width images. These numbers have gotten smaller because we're catching more images and squishing them to our max size.

audiodude:
And of course, as you predicted @Jaifroid, we get this:

image

Looks like we need to add logic to only scale in the dominant dimension.

audiodude:
Okay, added logic for scaling the smallest dimension to 320px, while keeping the aspect ratio.

This, unsurprisingly, gives us the biggest ZIM yet (`.114A-320` is the current build)

```
14272	wikipedia_bm_all_maxi_2024-11.113.zim
14532	wikipedia_bm_all_maxi_2024-11.114-280.zim
15364	wikipedia_bm_all_maxi_2024-11.114-320.zim
15640	wikipedia_bm_all_maxi_2024-11.114A-320.zim
```

And

image

kelson42:
@audiodude To me your approach is going to fail, this is an heuristic to approach the result of Mediawiki. If we don´t have the data with the proper value (which seems the case) here, we should retake the logic from Mediawiki, otherwise we will have a hight discrepancy.

audiodude:
> @audiodude To me your approach is going to fail, this is an heuristic to approach the result of Mediawiki. If we don´t have the data with the proper value (which seems the case) here, we should retake the logic from Mediawiki, otherwise we will have a hight discrepancy.

Thanks for the response @kelson42. I think "fail" is a strong word. Fail in what sense? Fail to produce an identical ZIM as compared to 1.13? Clearly. But I think it succeeds in creating reasonably sized images for scraped MediaWikis.

I did spend time trying to find the code in Page Content Service for the "mobile section" endpoint image resizing, but it was hard to track down. I don't see why it's important to do *exactly* the same thing in the new version of mwoffliner.

kelson42:
> > @audiodude To me your approach is going to fail, this is an heuristic to approach the result of Mediawiki. If we don´t have the data with the proper value (which seems the case) here, we should retake the logic from Mediawiki, otherwise we will have a hight discrepancy.
> 
> Thanks for the response @kelson42. I think "fail" is a strong word. Fail in what sense? Fail to produce an identical ZIM as compared to 1.13? Clearly. But I think it succeeds in creating reasonably sized images for scraped MediaWikis.
> 
> I did spend time trying to find the code in Page Content Service for the "mobile section" endpoint image resizing, but it was hard to track down. I don't see why it's important to do _exactly_ the same thing in the new version of mwoffliner.

As long as we don't original, we will have a hard time to handle all the edge cases. This is what I would like to avoid and why I believe this approach will probably fail.

Thad been said, we hae to move forward and I will open a ticket to follow the "other path", so we can - at least for the moment - benefit of this fix and move forward. Thx for the PR.

kelson42:
See https://github.com/openzim/mwoffliner/issues/2107 for the follow-up issue.

PR #2110, opened 2024-12-04
Relocate inline <body> scripts to their own tag in the <head>

### LINKED ISSUE #2096: REGRESSION: Remove new inline JS from ZIMs produced by dev (at least 1.14.0) to comply with restrictive CSPs
This is a regression of #1578 and its fix may depend on a fix to #2074. However, it is useful to revisit this issue, as a lot of work was done in the past to eradicate inline JS from mwOffliner ZIMs. The issue is that where a restrictive CSP is imposed that bans inline JS as opposed to attached JS (a real-life example of this is in Chrome extensions, so this negatively affects Kiwix JS Browser Extension), no inline JS can be run (as opposed to JS in attached scripts).

In the latest ZIMs scraped from the mobile-html endpoint using dev 1.14.0 (at least), e.g. in `wikipedia_bm_all_nopic_2024-10.zim`, I believe the functionality to unhide hidden sections depends on the following piece of inline JS:

Image

As this is blocked by the CSP directive enforced in Chromium extensions, it seriously affects the usability of such ZIMs in Chromium extensions running in ServiceWorkerLocal mode.

This is a regression because, as mentioned, we did a lot of work previously to remove inline JS from all Kiwix scrapers (see the list here https://github.com/kiwix/kiwix-js/issues/865), so we should be very careful not to re-introduce them in development. If it is in fact needed, then the script containing the desired function to run should attach itself internally to an appropriate browser event, and not be run from the HTML.

The fix to this issue may simply be to fix #2074 to ensure all sections are open by default. Then I believe it won't be necessary to run the inline JS here, unless it is doing other things that are desirable (though they should not be essential to be able to read the ZIM in browsers that don't run any JS from the ZIM).

audiodude:
Fixes #2096 

In the abstract Renderer, we create a method, collectInlineJs, which grabs any `<script>` tags that are in the body of the article, takes the actual JavaScript inside, and collects it into a string. The original `<body>` `<script>` tags are removed in the process. This string is passed to the callback of `processHtml`, where the concrete Renderers convert the intermediate `<body>` HTML into a full HTML document, and put in the `<head>`.

We introduce a new template, `inline_js_to_head_wrapper.html`, which contains a single script tag, with the body scripts wrapped in a `DOMContentLoaded` listener. This means that the scripts will still execute, in supported environments, after the page has loaded. It is unclear the proper function of these scripts, but the thought is that we might as well try to run them since we're already including scripts such as `pcs.js` in the page.

### DIFF
--- a/res/templates/inline_js_to_head_wrapper.html
+++ b/res/templates/inline_js_to_head_wrapper.html
@@ -1,0 +1,8 @@
+<script>
+(function () {
+  // Scripts taken from inline JS, concatenated, and moved here.
+  document.addEventListener('DOMContentLoaded', function () {
+    {{ jsContent | safe }}
+  })
+})()
+</script>
--- a/res/templates/pageWikimediaDesktop.html
+++ b/res/templates/pageWikimediaDesktop.html
@@ -1,10 +1,10 @@
-<!DOCTYPE html>
-<html class='client-js'>
+<!doctype html>
+<html class="client-js">
   <head>
-    <meta charset="UTF-8"/>
+    <meta charset="UTF-8" />
     <title></title>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0"/> __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_LIST__
-    __CSS_LINKS__ __JS_SCRIPTS__
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_LIST__ __CSS_LINKS__ __JS_SCRIPTS__ __RELOCATED_INLINE_JS__
   </head>
   <body class="mediawiki mw-hide-empty-elt ns-0 ns-subject stable skin-minerva action-view animations">
     <div id="mw-mf-viewport" class="feature-header-v2">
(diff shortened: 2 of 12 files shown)


Jaifroid:
@audiodude Just to clarify, the idea is that they are collected and put into a separate js file which is then called on `DOMContentLoaded`? The issue being that in certain environments we can only run scripts that are attached as files, rather than raw JS code included in `<script>` tags in the document (whether in the head or the body). If that is what you're doing in the PR, then it should solve the problem.

Jaifroid:
I'll take a look at the code asap.

audiodude:
> @audiodude Just to clarify, the idea is that they are collected and put into a separate js file which is then called on `DOMContentLoaded`? The issue being that in certain environments we can only run scripts that are attached as files, rather than raw JS code included in `<script>` tags in the document (whether in the head or the body). If that is what you're doing in the PR, then it should solve the problem.

Nope that's not it. I guess I misunderstood the requirements, so thanks for looking!

This collects all of the inline `<body>` scripts and shoves them in a `<script>` tag in the head. I don't see any reasonable way (though it is technically possible) to collect them into their own file and load that. So I think we should pare back this PR to just remove them altogether.

Jaifroid:
Ah OK. It used to be the case that mwOffliner could include its own script files (overrides). It would probably be complicated to do. If these scripts are not doing anything useful now that all sections are open anyway (their function was to open the sections), then removing them would be the simpler option.

audiodude:
> Ah OK. It used to be the case that mwOffliner could include its own script files (overrides). It would probably be complicated to do.

mwoffliner can, and still does, that. However, it's for generic scripts that apply to all pages. We would have to assume every page has `onBodyStart()` and `onBodyEnd()` only, and include a script that calls those, using the current method of just attaching a single script file to every page, while stripping them from the body. This is possible, but not resilient to new scripts being added to the body, which would be stripped but not have their functionality replaced.

> If these scripts are not doing anything useful now that all sections are open anyway (their function was to open the sections), then removing them would be the simpler option.

Yes, as I said on the bug, [the README for these scripts](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/48022ecd51e52ffac665a1a59ac5cf1acef7ec39/pagelib/docs/pcs/pcs.md) say they are for lazy loading images, "themeing" (whatever that is), and obviously for unhiding sections (though the README doesn't mention that).

PR #2111, opened 2024-12-05
Remove inline scripts from scraped articles

### LINKED ISSUE #2096: REGRESSION: Remove new inline JS from ZIMs produced by dev (at least 1.14.0) to comply with restrictive CSPs

audiodude:
Fixes #2096 

Simply find and remove all script tags from scraped HTML when rendering the article. This doesn't affect "module dependencies", because they are calculated before and re-added later.

### DIFF
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -698,6 +698,14 @@ export abstract class Renderer {
       }
     })
 
+    /*
+     * Because of CSP, some ZIM reader environments do not allow inline JS. See issues/2096.
+     */
+    const scripts = Array.from(parsoidDoc.getElementsByTagName('script')) as DominoElement[]
+    for (const script of scripts) {
+      script.parentNode.removeChild(script)
+    }
+
     /* Force display of element with that CSS class */
     filtersConfig.cssClassDisplayList.map((classname: string) => {
       const nodes: DominoElement[] = Array.from(parsoidDoc.getElementsByClassName(classname))
--- a/test/unit/saveArticles.test.ts
+++ b/test/unit/saveArticles.test.ts
@@ -212,6 +212,36 @@ describe('saveArticles', () => {
       // Prague was correctly post-processed
       expect(PragueDocument.querySelector('#POST_PROCESSOR')).toBeDefined()
     })
+
+    test('Removes inline JS', async () => {
+      const { downloader, dump } = await setupScrapeClasses({ mwUrl: 'https://en.wikipedia.org' }) // en wikipedia
+      downloader.setUrlsDirectors(rendererInstance, rendererInstance)
+      const articleId = 'Potato'
+      const articleUrl = downloader.getArticleUrl(articleId)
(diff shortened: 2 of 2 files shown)


src/renderers/abstract.renderer.ts
```diff
@@ -698,6 +698,14 @@ export abstract class Renderer {
       }
     })
 
+    /*
+     * Because of CSP, some ZIM reader environments do not allow inline JS. See issues/2096.
+     */
+    const scripts = Array.from(parsoidDoc.getElementsByTagName('script')) as DominoElement[]
```
Jaifroid, line 704: I guess this will also remove non-inline scripts, i.e. scripts with `src`? These are not blocked by CSP. I mean, I have never seen an essential script in Wikimedia archives, but I'm a little nervous about removing them all. I guess with sufficient testing, we might be able to say that in current Wikimedia ZIMs there are no undesirable effects... And we could modify this if there turned out to be some pages that require scripts for specific features, i.e. a release-and-fix approach...

Jaifroid (approved):
Other than the caveat mentioned, and if the risk is acceptable, the code looks fine to me!

src/renderers/abstract.renderer.ts
audiodude: Good catch! However, the way mwoffliner works is that it only takes the `<body>` tag from the returned HTML. It processes other tags to figure out the "JS module dependencies" for the article, but downloads those separately and re-inserts links to the in-ZIM versions of those scripts later.

So while this code technically wipes out all `<script>` tags in the document, in practice any that were not in the `<body>` were going to be thrown out anyways.

PR #2120, opened 2024-12-26
Remove S3 upload concurrency

### LINKED ISSUE #2118: Cache error while uploading object: RequestTimeTooSkewed
From https://farm.openzim.org/pipeline/845ea49e-0d80-4d92-8127-1afce890c276/debug

benoit74:
Fix #2118

Changes:
- revert increase of sockets from #2093, this was not needed in the past, and I don't expect this to be needed if we do not push too many concurrent requests (which we shouldn't)
- await upload of image to S3: while not awaiting the upload could help save some crawling time, it cannot be efficiently handled without proper queue size management (otherwise we end-up with current situation where too many uploads are queued at once). Implementing proper queue management is not worth it

Nota: we might want to open an issue to support async S3 upload with proper queue management on the medium/long term. I'm not convinced at all this is worth to implement, most (all?) Python scrapers work pretty well without it and I do not expect much benefit (i.e. price to pay is too high)

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -588,7 +588,7 @@ class Downloader {
           // Check for the ETag and upload to cache
           const etag = this.removeEtagWeakPrefix(mwResp.headers.etag)
           if (etag) {
-            this.s3.uploadBlob(stripHttpFromUrl(url), mwResp.data, etag, mwResp.headers['content-type'], this.webp ? 'webp' : '1')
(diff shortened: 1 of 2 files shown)


kelson42:
@benoit74 Thank you very much for the PR. I will let @audiodude full judge this PR.

kelson42:
@rgaudin My experimentations on Linux with a 2x DPR screen:
* Qt6 does not fix anything in itself
* Your patch proposal does not work for me.

kelson42:
@audiodude Can you please review this PR?

kelson42 (approved):
LGTM

audiodude (commented):
LGTM

Issue #2090, opened 2024-10-05
Why alecexposed.org is complaining about `No available desktop renderer.`?


benoit74:
Zimfarm recipe: https://farm.openzim.org/recipes/alec_exposed

The [last task](https://farm.openzim.org/pipeline/f694890e-12b6-4970-a020-b2432073940c) (ran with `dev` right now, see close to a 1.14) fails with `No available desktop renderer.`.

I don't get why, because all [prerequisites](https://github.com/openzim/mwoffliner/wiki/Frequently-Asked-Questions#can-all-mediawiki-instances-be-scraped) are ok:
- Mediawiki version is 1.30 (see https://www.alecexposed.org/wiki/Special:Version)
- ActionParse API seems to be working

Working: https://www.alecexposed.org/w/api.php?action=parse&format=json&prop=text|modules|jsconfigvars|headhtml&formatversion=2&page=ALEC_Bills

Working with warning `Unrecognized parameter: parsoid`: https://www.alecexposed.org/w/api.php?action=parse&format=json&prop=text|modules|jsconfigvars|headhtml&formatversion=2&page=ALEC_Bills&parsoid=1

audiodude:
I added some logging and I'm getting the following:

```
====== Error checking API availability: https://www.alecexposed.org/api/rest_v1/page/html/ALEC_Exposed
====== Error checking API availability: https://www.alecexposed.org/api/rest_v1/page/mobile-html/ALEC_Exposed
====== Error checking API availability: https://www.alecexposed.org/w/rest.php/v1/page/ALEC_Exposed/html
```

In fact all of those URLs are 404s.

So I'm not sure if the site is misconfigured and the endpoints are not enabled, or if mwoffliner is misconfigured and we're not looking for them in the right places.

As far as "ActionParse API", I don't see any references to that in the mwoffliner code and I'm not sure it's an API we work with.

benoit74:
Regarding "ActionParse API", I was refering to the name listed [here](https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points). 

REST API is indeed not available on this website, so it is normal these calls are returning 404.

But why none of the logs you get are referring to something like `https://www.alecexposed.org/w/api.php?action=parse&format=json&...`? Is there another place in the codebase to check?

audiodude:
[...]

PR #2125, opened 2025-01-03
Add logs for API checks

audiodude:
This will help debugging when we get a message along the lines of "No available [foo] renderer".

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -175,7 +175,9 @@ class MediaWiki {
   public async hasWikimediaDesktopApi(): Promise<boolean> {
     if (this.#hasWikimediaDesktopApi === null) {
       this.wikimediaDesktopUrlDirector = new WikimediaDesktopURLDirector(this.wikimediaDesktopApiUrl.href)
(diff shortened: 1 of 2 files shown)


audiodude:
See #2090

benoit74:
I love this. Given these checks are done only once per scrape, the 4 additional log lines are "very cheap" and having the exact URL used helps a lot to ensure there is no problem of misconfiguration of the scraper URLs.

My only suggestion is that you should probably use `logger.log` rather than `console.log` so that it is properly filtered by the log levels system.

audiodude:
> My only suggestion is that you should probably use `logger.log` rather than `console.log` so that it is properly filtered by the log levels system.

Thanks, done.

kelson42:
@audiodude Could we use also this PR to log as well what API end-point(s) is/are chosen at the end?

audiodude:
Great idea, done.

audiodude:
Also changed the code to only build the URL once, PTAL

kelson42:
@audiodude I guess this is ready to review!?

benoit74 (approved):
LGTM, please consider replacing `Checking` with `Checked` in the messages since the check is complete once log is done

Issue #1921, opened 2023-10-04
Release 1.14.0


kelson42:
~~Blocked by https://github.com/openzim/zimfarm/issues/990~~

- [ ] Check that dependencies have been updated to latest version
- [ ] Check Usage (in `README.md`) and FAQ are still up-to-date
- [x] Ensure the CI is green on git `main`
- [x] Update Github milestone to match the issues that will be released 
- [x] Close Github milestone
- [x] Update the Changelog so that it is in line with the content of Github milestone
- [x] Update version in `package.json` to `x.x.x`
- [x] Update the Changelog so that it is in line with the content of Github milestone
- [x] Commit and push these changes to tag on Github named `vx.x.x` (`v` prefix is mandatory)
- [x] Create a new release tag on git
- [x] Create a Github release, pointing to the tag, with the Changelog of this release, and publish the Github release (this will trigger the CI, if the CI fails and you have to push a minor fix which does not justify to create a new version, you will have to delete the release and re-create it from scratch)
- [x] Check that version is published as a Github release at https://github.com/openzim/mwoffliner/releases and tagged `latest`
- [x] Review and approve release workflow at https://github.com/openzim/mwoffliner/actions
- [x] Check that version is published on NpmJs at https://www.npmjs.com/package/mwoffliner
- [x] Check that version is published on Github Container Registry at https://ghcr.io/openzim/mwoffliner and tagged `latest`
- [x] Create a new Github milestone with the next minor version incrementaly
- [x] Create a new Github issue attached to this milestone with this checklist inside
- [x] Create new `Unreleased` section in Changelog (placeholder for future entries)
- [x] Adjust version in `package.json` to `x.y.z+1-dev0` (next **patch** version incrementaly)
- [x] Commit and push these changes to `main` Github branch
- [x] Inform rgaudin/benoit74 that a new release is ready to use so that they will update Zimfarm recipes
- [ ] Advertise the new scraper release on #scrapers Slack channel
- [x] If needed, open a PR on Zimfarm to add support for new CLI parameters of interest / remove old ones / update recipes (or ask rgaudin/benoit74 for help)

kelson42:
@benoit74 I tried to published, but somehow the upload to npmjs.org fails, see https://github.com/openzim/mwoffliner/actions/runs/12735058997/job/35493358938. To me it looks like a crendential (token) problem, but not sure. Could you please have a look on monday?

benoit74:
[...]

PR #2129, opened 2025-01-12
Publish container image on release event

kelson42:
By publishing `1.14.0` I saw that this workflow what not OK. I suspect we will have to release an `1.14.1` to get everything done properly.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,11 +1,8 @@
 name: Docker
 
 on:
-  push:
-    branches:
-      - main
-    tags:
-      - v*
+  release:
+    types: [published]
 
 jobs:
   build-and-push:


audiodude:
Don't you still want the "on...push..." sections, because that's what publishes the `:latest` and `:1.14` images?

kelson42:
It didn't because I don't want to start tags with "v".

benoit74:
Problem is simply that https://github.com/openzim/mwoffliner/issues/1921 is incorrect / incomplete, you need to create a `v1.14.0` tag, not a `1.14.0`, it has been so since multiple releases:

image

I will fix this situation, no need for `1.14.1`, code is not changing in `main` since yesterday.

benoit74:
I've updated https://github.com/openzim/mwoffliner/issues/1921#issue-1925408497 so that it is in-line with other scrapers practices and level of details

benoit74:
This is done differently now, to be aligned with practices on other scrapers.

kelson42:
@benoit74 I clearly didn't wanted to put at tag `v...` this is uselessly complicated and error prone. This rule to put a `v` before is just an heritage and IMHO should not be perpetuated. None of the software (maybe beside scrapers) uses this.

PR #2148, opened 2025-01-30
Pre-install all Node.JS dependencies to make image smaller/faster

benoit74:
Since Node.JS dependencies rarely need to be changed during successive builds of the Docker image, it is way better to first install only dependencies without mwoffliner code, and then install mwoffliner itself. Any change to the codebase will hence only update few things without having to re-download again all dependencies.

Edit: to be merged to main once https://github.com/openzim/mwoffliner/pull/2143 is merged

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,14 +11,15 @@ RUN apt-get update && \
 # Install mwoffliner
 WORKDIR /tmp/mwoffliner
 COPY *.json ./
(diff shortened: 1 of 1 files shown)


kelson42:
@benoit74 Great PR. I will test it this WE, but I would really appreciate @audiodude expertise in the review work.

kelson42 (approved):
LGTM

Issue #1972, opened 2024-01-12
Thumbnails captions texts are getting overflowed


Shreyassp002:
From **Bollywood Offline** app of the playstore version:
While Scrolling through the Home Screen of the app some of the texts are getting overflowed.

Model Name : Poco C55
Android Version : 13

Screenshots : 
1)
![1705082620724](https://github.com/openzim/mwoffliner/assets/96625037/b1975329-7f80-4775-93ea-a16a57841519)

2)
![1705082620698](https://github.com/openzim/mwoffliner/assets/96625037/c955464a-e2d2-4dff-a590-c7a1322daf47)

saurav-sinha-3141:
Hey @Shreyassp002 , just wanted to know if the issue still persists. If not well and good but if it still exists can you please tell the steps to reproduce this error?

madhav2348:
@Shreyassp002  , is this issue still persists

madhav2348:
hey @kelson42 can i work on this

benoit74:
@madhav2348 sure. Please start by confirming the issue is still present on production ZIM, and then build up a test case so that you can quickly diagnose the problem (typically a list of 20 articles or so, with some which have the problem).

PR #2169, opened 2025-02-18
Fix width of thumbnails caption to avoid overflow

benoit74:
Fix https://github.com/openzim/mwoffliner/issues/1972

| Before | After |
|--|--|
| Screenshot 2025-02-18 at 14 27 13 | Screenshot 2025-02-18 at 14 33 38 |


kelson42:
@benoit74 Would not be a good idea to add ellipsis in case it is impossible to display properly (width to small)?

PR #2189, opened 2025-03-19
Store test ZIMs in CI storage so that we can check for actual regressions when file counts change

audiodude:
Would love to figure out a way to confirm, on-wiki, if these file numbers are correct or if the test is actually uncovering an issue. For now I've simply bumped them blindly to get CI to pass again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ env:
   S3_URL: ${{ secrets.S3_URL }}
   ZIMCHECK_PATH: $( find .. -name zimcheck )
   ZIMDUMP_PATH: $( find .. -name zimdump )
+  KEEP_ZIMS: 1
 
 jobs:
   ci-test:
@@ -53,6 +54,15 @@ jobs:
       - name: Running scheduled all tests (no coverage)
         if: ${{ github.event_name == 'schedule' }}
         run: npm run test-without-coverage
+
+      - name: Uploading ZIM artifacts
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          path: mwo-test-*/*.zim
+          # keep scheduled (daily artifacts for 7 days, others only 1 day)
+          retention-days: ${{ github.event_name == 'schedule' && 7 || 1 }}
+          compression-level: 0
 
       - name: Mailgun Action
         if: ${{ failure() && github.event_name == 'schedule' }}
--- a/test/e2e/apiPathParamsSanitizing.e2e.test.ts
+++ b/test/e2e/apiPathParamsSanitizing.e2e.test.ts
@@ -17,7 +17,7 @@ const parameters = {
   mwWikiPath: sanitizeWikiPath('/wiki'),
 }
 
-await testAllRenders(parameters, async (outFiles) => {
+await testAllRenders('api-path-params', parameters, async (outFiles) => {
   describe(`e2e test for api url params for en.wikipedia.org for ${outFiles[0]?.renderer} renderer`, () => {
     test('Mediawiki actionApiPath option sanitized', () => {
       expect(outFiles[0].mwMetaData.actionApiPath).toBe('/w/api.php')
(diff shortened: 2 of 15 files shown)


audiodude:
More failures now with file numbers not being right. Without a way to independently verify these, I'm starting to wonder if this is an actual issue?


```
FAIL test/e2e/wikisource.e2e.test.ts (38.826 s)
  wikisource
    ✕ Wikisource List for WikimediaDesktop renderer (30 ms)
    ✕ Wikisource List for VisualEditor renderer (25 ms)

  ● wikisource › Wikisource List for WikimediaDesktop renderer

    expect(received).toBeGreaterThanOrEqual(expected)

    Expected: >= 13
    Received:    7

      30 |                 console.log(dump.status.files.fail)
      31 |                 // nopic has enough files
    > 32 |                 expect(dump.status.files.success).toBeGreaterThanOrEqual(13)
         |                                                   ^
      33 |                 // nopic has enough redirects
      34 |                 expect(dump.status.redirects.written).toBeGreaterThanOrEqual(16)
      35 |                 // nopic has enough articles

      at Object.<anonymous> (test/e2e/wikisource.e2e.test.ts:32:51)

  ● wikisource › Wikisource List for VisualEditor renderer

    expect(received).toBeGreaterThanOrEqual(expected)

    Expected: >= 13
    Received:    7

      52 |               if (dump.nopic) {
      53 |                 // nopic has enough files
    > 54 |                 expect(dump.status.files.success).toBeGreaterThanOrEqual(13)
         |                                                   ^
      55 |                 // nopic has enough redirects
      56 |                 expect(dump.status.redirects.written).toBeGreaterThanOrEqual(16)
      57 |                 // nopic has enough articles

      at Object.<anonymous> (test/e2e/wikisource.e2e.test.ts:54:51)
```

benoit74:
Very good question, AFAIK it is impossible to see these numbers in the UI. 

To me the only way to answer that question is to:
- have a look at which files are in last ZIM of library.kiwix.org and not in these tests (if feasible ...)
- run same crawl as in the test locally and check whether the ZIM work as expected

To me, these tests are only a warning "something has changed, you should have a manual look". Because we can have two CSS merged into a single one file with exactly the same functionality. Or we could retrieve CSS files with only bullshit inside, still have the proper number of files but have a completely broken ZIM.

audiodude:
The scrapes that these tests do do not correspond to any public ZIM.

It sounds like we need to upload the test ZIMs themselves as artifacts somewhere, so that when these tests fail, we can manually inspect (as you said). Otherwise, we're just guessing.

benoit74:
Indeed ... no need to keep them for long, but at least having previous "ok" version would help a lot

benoit74:
I've added CI configuration (+ small tests code changes) to upload e2e tests ZIMs as artifacts ; waiting for completion to confirm size and hence retention period

benoit74:
I've:
- added more changes in numbers
- modified the CI store e2e ZIMs as a CI artifact
  - each artifact is about 250M, and we pay (few cents per GB.month) for this storage, so we should be cautious
  - daily scheduled runs are kept for 7 days so that we have a chance to go back to a "previously working" artifact in case of regression, 7 days is deemed sufficient, let's see
  - other artifacts (push to main + pull requests) are kept only for 1 day (we can rebuild them on-demand, and we can produce many of them
- ⚠️ modified computation of ZIM name / filename from an article list
  - we used the underscore `_` as separator between words while we should aim for dash `-` since underscores are reserved for name parts in the convention (https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention)
  - we kept to many special chars ; it caused an issue in CI, but is also again not aligned with our convention

@kelson42 @audiodude feedback is welcomed

benoit74 (approved):
LGTM

audiodude:
I'm not sure it makes sense or is a good idea to change the ZIM naming convention in this PR, which otherwise only deals with updating broken e2e/CI tests and storing CI ZIMs.

benoit74:
Let's move this to another PR then

benoit74:
Ready for review again (after https://github.com/openzim/mwoffliner/pull/2196)

audiodude:
LGTM (can't give actual approval because this is technically my PR)

PR #2193, opened 2025-03-24
Upgrade node libzim to 2.0.3

### LINKED ISSUE #1576: Use latest node-libzim
… and bring all the improvments of libzim7, see changelog:
https://github.com/openzim/libzim/blob/master/ChangeLog

This depends on https://github.com/openzim/node-libzim/issues/69

audiodude:
Fixes #1576

DO NOT MERGE.

This is a WIP to upgrade the version of `node-libzim` used by mwoffliner. The dependency in the package.json has been updated and all call sites have been migrated to the new API.

However, the code doesn't function. It segfaults while trying to write the ZIM.

So far, my belief is that this is some kind of concurrency issue, since `node-libzim` does not provide any thread safety guarantees. It seems logical to me that we are trying to write to a file descriptor in multiple threads, which is causing the segfault.

The test added here (https://github.com/openzim/mwoffliner/pull/2193/files#diff-11a316e0e8dcb3cf8faa6ececd70b5e1185706783132e453f70a4ad73d29283f) attempts to both provide a working example of writing to the ZIM, as well as several "degenerate" examples which should be subject to similar segfaulting as we see when running the main code. However, all of the test code in that file passes.

I have also found that if I comment out [this code](https://github.com/openzim/mwoffliner/blob/main/src/mwoffliner.lib.ts#L439-L452) and [this code](https://github.com/openzim/mwoffliner/blob/main/src/mwoffliner.lib.ts#L427), the scrape completes successfully, but it might just be the case that no files are being written to the ZIM and that's why it succeeds.

Either way, this PR is *NOT* ready to merge, and doesn't fully fix the issue because it causes segfaults.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,17 @@
 {
   "name": "mwoffliner",
-  "version": "1.14.0",
+  "version": "1.14.1-dev0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "mwoffliner",
-      "version": "1.14.0",
+      "version": "1.14.1-dev0",
       "license": "GPL-3.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.374.0",
         "@ladjs/country-language": "^1.0.3",
-        "@openzim/libzim": "2.4.4",
+        "@openzim/libzim": "3.2.0",
         "@types/async": "^3.2.18",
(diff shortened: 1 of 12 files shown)


benoit74:
Superseded by #2194, no offense

audiodude:
None taken, love to see the progress!

PR #2194, opened 2025-03-25
Upgrade to latest node libzim 3.2.3

### LINKED ISSUE #1576: Use latest node-libzim

benoit74:
Fix #1576

This is the next step on https://github.com/openzim/mwoffliner/pull/2193

I created a different branch / PR because I had to force-push to rebase on `main` latest changes and was worried to break / loose something in the process.

Most of the solution around segfaults is in https://github.com/openzim/mwoffliner/pull/2194/commits/22283a6f582b26ecddf31624c5f68fe84777980e : add a mutex to ensure only one thread add an item to the ZIM at a given time (the call to libzim returns quite fast once libzim has checked it has everything needed to process ; libzim handles completion of operation in an async manner, which we hopefully do not need to track)

I then had to fix many stuff around URLs which were wrong do to the removal of the namespace in the URLs.

This must be merged after ~~Finally, I backported changes of~~ https://github.com/openzim/mwoffliner/pull/2189 is merged ~~to make CI pass as well (will be removed once this other PR is merged).~~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
         "imagemin-optipng": "^8.0.0",
         "imagemin-pngquant": "^9.0.2",
         "imagemin-webp": "^7.0.0",
+        "jest": "^29.7.0",
         "md5": "^2.3.0",
         "merge": "^2.1.1",
         "mkdirp": "^2.1.6",
@@ -79,7 +80,6 @@
         "eslint-plugin-prefer-arrow": "^1.2.3",
         "eslint-plugin-prettier": "^4.2.1",
         "file-type": "^18.2.1",
-        "jest": "^29.3.1",
         "nyc": "^15.1.0",
         "prettier": "2.8.7",
         "replace": "^1.2.2",
@@ -105,7 +105,6 @@
       "version": "2.2.1",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
       "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
-      "dev": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.0",
         "@jridgewell/trace-mapping": "^0.3.9"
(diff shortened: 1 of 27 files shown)


audiodude:
Does this mean I was right that it was a concurrency issue?

I didn't realize that a mutex could so easily be implemented in TS.

Great work!

benoit74:
> Does this mean I was right that it was a concurrency issue?

Yes sir, you were 100% right, implementing mutex was the only thing I really needed to make it work.

Then tons of other stuff was still broken due to of namespaces in Redis and URLs and other stuff around this, but that's mostly "details" compared to the segfault issue.

> I didn't realize that a mutex could so easily be implemented in TS.

I don't know what it really implies under the hood tbh, but the abstraction is indeed straightforward. From my experience, there is great libs for mutex / synchronization / semaphores in most languages nowadays. And it is not like we are optimizing for high performance computing, so I didn't look at details to save few milliseconds where we already spend seconds in HTTP calls.

benoit74:
@audiodude could you develop a bit around https://github.com/openzim/mwoffliner/pull/2194/commits/3ed8cf4f526d678d20bb189480d146375f7509df ? I feel like it is OK, but I don't have much justification on whether it is mandatory, an optimization, a nice-to-have, ...

I'm sorry but I finally totally amended your commit to simplify Git history, we do not have your name anymore as a certified author of the first steps of this migration ... but it was just too much of a mess otherwise (or too painful to rewrite)

kelson42:
> Does this mean I was right that it was a concurrency issue?
> Yes sir, you were 100% right, implementing mutex was the only thing I really needed to make it work.

If possible this should be built in node-libzim (actually IMHO it should be part of libzim itself, but this is another story).

benoit74:
> If possible this should be built in node-libzim (actually IMHO it should be part of libzim itself, but this is another story).

I've opened https://github.com/openzim/node-libzim/issues/155 for the matter, no need to dig into it for now to merge this PR.

audiodude:
> @audiodude could you develop a bit around [3ed8cf4](https://github.com/openzim/mwoffliner/commit/3ed8cf4f526d678d20bb189480d146375f7509df) ? I feel like it is OK, but I don't have much justification on whether it is mandatory, an optimization, a nice-to-have, ...
> 
> I'm sorry but I finally totally amended your commit to simplify Git history, we do not have your name anymore as a certified author of the first steps of this migration ... but it was just too much of a mess otherwise (or too painful to rewrite)

Don't worry about the attribution, I don't mind.

I believe `nodenext` is the recommended value for modern projects. I'm not sure if it was strictly necessary for importing node-libzim, but it might have been based on the changes that library has undergone. Feel free to revert and see if the tests pass?

audiodude (approved):
LGTM

PR #2196, opened 2025-03-28
Fix the filename computation of article list

benoit74:
When the computation of ZIM filename is based on article list, and when the article list contains real article titles (instead of file / URL), the resulting ZIM name does not necessarily respect our convention:

- we used the underscore _ as separator between words while we should aim for dash since underscores are reserved for name parts in the convention (https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention)
- we kept too many special chars, only lowercase alphanum is allowed

Note that this is a temporary fix, mid-term solution is probably https://github.com/openzim/mwoffliner/issues/2197 but there is just too much to analyze in this issue, this simple fix is enough for the short term

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ env:
   S3_URL: ${{ secrets.S3_URL }}
   ZIMCHECK_PATH: $( find .. -name zimcheck )
   ZIMDUMP_PATH: $( find .. -name zimdump )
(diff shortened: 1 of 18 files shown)


benoit74:
Failing tests are the ones fixed in https://github.com/openzim/mwoffliner/pull/2189

audiodude (approved):
LGTM

PR #2204, opened 2025-04-01
Improve error logging

benoit74:
Current logging strategy had many flaws:
- `cleanupAxiosError` was not systematically called, resulting in some very long error message when an AxiosError was raised up to the top of the stack
- the stack trace was generally not available
- the error message was often displayed twice
- the stack trace was impossible to use because it reported JS line numbers instead of TS line numbers

This has been fixed:
- use source maps to display TS line numbers in stack trace
- use `cleanupAxiosError` at the top of the stack
- simplify logging of errors to avoid duplicates
- stop transforming error into JSON, this is handled by Node logging better than on our own

It also removes a silly `debugger` statement, probably left from a debugging session


kelson42 (changes requested):
CI fails

src/mwoffliner.lib.ts
```diff
@@ -358,7 +358,6 @@ async function execute(argv: any) {
         await doDump(dump)
         await filesToDownloadXPath.flush()
       } catch (err) {
-        debugger
         throw err
```
audiodude, line 361: Agree with CodeFactor. `catch(err) { throw err }` is a no-op

audiodude (approved):
LGTM

benoit74:
> CI fails

It doesn't fails, it has an sporadic issue while downloading a package.

> Agree with CodeFactor. catch(err) { throw err } is a no-op

Indeed, found other occurrences I cleaned-up as well.

benoit74:
Finally CI passed ... I had to disable openstreetmap tests which were failing because wiki is in bad shape (not totally down, but everything is broken)

kelson42 (approved):
LGTM

PR #2205, opened 2025-04-01
Refactor modules download

benoit74:
~~⚠️ to be merged only once https://github.com/openzim/mwoffliner/pull/2204 is merged, hence the unusual base branch ⚠️~~

In preparation for the introduction of ActionParse API, this PR refactors the downloading of module dependencies.

Before that PR, downloading of module dependencies was handled above the renderer. In fact, this should be done differently based on the renderer used. It was visible even before the introduction of ActionParse API since we had to filter the list of modules afterwards at render stage, since this is a decision done by the renderer. This will be mandatory with ActionParse API since we will download both article content and article modules in a single HTTP call.

Changes:
- renderers now have a second abstract method: `download` which must download both article data (mostly parsoid HTML, but depending on the API there could be more and format is varying) and list of module dependencies (styles, scripts and jsConfigVars)
- this made it obvious that `webp` argument is in fact not needed in many interfaces
- simplify a bit the code of `Downloader.getModulesDependencies` (only semantics)
- extract logic to compute `jsConfigVars` from head HTML from`Downloader.getModulesDependencies` to static `Downloader.extractJsConfigVars`
  - we will need this logic in ActionParse renderer `download` method, because all other renderers continue to use the "default" `Downloader.getModulesDependencies` method, but not future ActionParse renderer
- new interfaces `DownloadOpts` and `DownloadRes` for type checking
- add an error when renderer passed is unknown (not supposed to happen because cleaned-up at CLI level but this could be a bug)
- remove `continuation` fallbacks, not needed

### DIFF
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -11,6 +11,7 @@ import { Dump } from '../Dump.js'
 import { rewriteUrlsOfDoc } from '../util/rewriteUrls.js'
 import { footerTemplate } from '../Templates.js'
 import { getFullUrl, getMediaBase, getRelativeFilePath, interpolateTranslationString, encodeArticleIdForZimHtmlUrl, getStaticFiles } from '../util/misc.js'
+import Downloader from 'src/Downloader.js'
 
 type renderType = 'auto' | 'desktop' | 'mobile' | 'specific'
 type renderName = 'VisualEditor' | 'WikimediaDesktop' | 'WikimediaMobile' | 'RestApi'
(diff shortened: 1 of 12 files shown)


src/renderers/abstract.renderer.ts
```diff
@@ -763,5 +775,6 @@ export abstract class Renderer {
     return doc.documentElement.outerHTML
   }
 
+  abstract download(downloadOpts: DownloadOpts): Promise<DownloadRes>
```
audiodude, line 778: Why are we using this pattern of one parameter that is a typed object, rather than just using normal typed parameters? Ie:

```
abstract download(downloader: Downloader, articleUrl: string, articleDetail: ArticleDetail)
```

src/renderers/visual-editor.renderer.ts
```diff
@@ -15,6 +15,7 @@ export class VisualEditorRenderer extends DesktopRenderer {
   private async retrieveHtml(renderOpts: RenderOpts): Promise<any> {
     const { data, articleId, articleDetail, isMainPage } = renderOpts
 
+    /* istanbul ignore if */
```
audiodude, line 18: What's this for?

test/unit/renderers/article.renderer.test.ts
audiodude, line 1: Yeah I never understood the point of the leading underscores, thanks for removing!

audiodude (approved):
LGTM

src/renderers/visual-editor.renderer.ts
benoit74: Specify that next line is not covered by tests and this is expected.

kelson42:
@benoit74 Can you please rebase (and fix conflicts)

kelson42 (commented):
@benoit74 Can we remove/fix the "articleDetails" Codefactor warning?

src/renderers/abstract.renderer.ts
kelson42: I have no strong opinions on this

PR #2217, opened 2025-04-02
Upgrade to Node 22 and upgrade all dependencies except eslint

### LINKED ISSUE #2132: Upgrade to Node.JS 22
Current LTS version of Node.JS is 22. We are currently on 18, which is approaching end of maintenance end of May.

We should upgrade to 22.

benoit74:
Fix #2132

Changes:
- upgrade to Node.JS 22 and mention we support only it in the README so that it is more obvious
- upgrade all Node.JS dependencies 
- upgrade to ESLint 9 with minimal set of plugins
  - only use recommended rules instead of custom sets we had before (not our expertise to know which rules are meaningful or not)
  - disable `@typescript-eslint/no-explicit-any`, there are just too many
  - see https://github.com/openzim/mwoffliner/issues/2218 for the rest of the work
- drop intermediate docker image with only custom redis since it was never updated by the CI
- fix some casing typos in `tsconfig.json`
- add mandatory `"isolatedModules": true` in `tsconfig.json` so that TS compiles without issue
- fix bug linked to axios not accepting URL with protocol anymore (e.g. `//upload.wikimedia.org/wikipedia/commons/thumb/6/67/London_Skyline_%28125508655%29.jpeg/330px-London_Skyline_%28125508655%29.jpeg` is not working while it was ; assuming the URL comes from the wiki and hence using same protocol)


kelson42:
@benoit74 May we meege this before actionparse?

benoit74:
Yes, no issue at all.

PR #2223, opened 2025-04-03
Fix silent build errors

### LINKED ISSUE #2130: Fix silent CI errors
See e.g. https://github.com/openzim/mwoffliner/actions/runs/12744447456/job/35516375224 (or any more recent `CI` job).

```
> mwoffliner@1.14.0 build
> ./dev/build.sh

Building at [Mon Jan 13 09:40:41 UTC 2025]
Error: node_modules/@types/imagemin-gifsicle/index.d.ts(7,24): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("imagemin")' call instead.
Error: node_modules/@types/imagemin-webp/index.d.ts(7,24): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("imagemin")' call instead.
Error: node_modules/split-by-grapheme/index.ts(2,15): error TS2339: Property 'Segmenter' does not exist on type 'typeof Intl'.
Error: node_modules/split-by-grapheme/index.ts(3,16): error TS2339: Property 'Segmenter' does not exist on type 'typeof Intl'.
Build Complete at [Mon Jan 13 09:40:49 UTC 2025]
```

I think two things needs to be fixed:
- we shouldn't have these errors
- we should have silent errors in the build process (errors should fail the build)

benoit74:
Fix #2130

Changes: see commits message

### DIFF
--- a/dev/build.sh
+++ b/dev/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 echo "Building at [$(date)]"
 npx tsc -p tsconfig.build.json
 # Remove tslint disable
--- a/test/unit/util/metaData.test.ts
+++ b/test/unit/util/metaData.test.ts
@@ -1,0 +1,18 @@
+import { byGrapheme } from 'split-by-grapheme'
+
+describe('Test metadata utilities', () => {
+  describe('Count graphemes', () => {
+    test('simple latin', async () => {
+      expect('title'.split(byGrapheme).length).toBe(5)
+    })
+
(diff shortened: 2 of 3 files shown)


kelson42 (approved):
LGTM

PR #2228, opened 2025-04-08
Fix npm install and enhance CI

benoit74:
Currently, `main` branch is broken, CI is failing.

The problem has been introduced in https://github.com/openzim/mwoffliner/pull/2223 and concerns only the Docker image build.

The fact that we do not test Docker image at all is an issue, we should at least try to build and start the image, this is introduced in this PR (and aligned with Python scraper "standard")

The Docker build issue is that build cannot complete in "early" dependencies installation step because it misses the cli.ts script. This PR proposes to create a fake cli.ts so that installation of dependencies can complete properly.

This PR also updates CI images to `ubuntu-24.04` since `ubuntu-22.04` is going to be removed soon.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   ci-test:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


kelson42 (approved):
LGTM even if I don't understand why we have commits related to memory issue in this apR

benoit74:
> even if I don't understand why we have commits related to memory issue in this apR

But the other PR on memory issue was based on this PR branch so that CI can pass properly. I should have mentioned it and you should not have merged the other PR before this one. Not a big deal.

benoit74:
Let me fix that situation before merging

PR #2229, opened 2025-04-08
Clearer status around supported Node version

### LINKED ISSUE #2227: We should be clearer about which version of Nodejs is supported
* `README.md` says 16.x+
* CI says 22.x only
* on my computer it seems to run as well with 18.x
* Nothing is forced in `package.json` AFAIK (can we do that?)

benoit74:
Fix #2227

- have only one version mentioned in README.md (sic)
- drop non-existent 'engine-strict' property of package.json
- add .npmrc with engine-strict rule to enforce by default the Node.JS version (applies only to developers, and even them can drop this file if they don't see fit ; users will only see a warning)

This seems to be the "best" we can do in a Node.JS context

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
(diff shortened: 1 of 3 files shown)


benoit74:
Please wait a bit for reviewing, I just realize other stuff is wrong in READM

benoit74:
Ready for review again, sorry

kelson42 (approved):
LGTM

PR #1728, opened 2023-01-12
Update and refactor node-redis

uriesk:
- update redis to current v4
- put the store into an own class
- don't fetch article details from redis again on getArticle, hand them down
- write interfaces for RedisKvs and RedisStore in types.d.ts and stop importing the classes everywhere
- split `RedisKvs.exists` into `.exists` and `.existsMany`
- rewrite the mapping functions of `RedisKvs` to not use `.reduce`
- `--redis` parameter is either redis URL or path to unix socket (v4 changed its constructor, have to now detect by `startsWith('/') || startsWith('./')` if its a path)
- Make sure that stores are flushed on exit, print errors on launch when stores from other runs are found (but continue)
- `getArticleIdsByNS`:
    - batch redis commands together and await them in the next iteration
    - detect when it freezes
    - log time
- `saveArticles`:
    - again batch more redis commands together (getMany for FileDetail and setMany instead of Set)
    - detect when it freezes
    - log time
- `getModuleDependencies`:
    - no silent process.exit(1) on error

Since that one test in the test/jest folder was just about RedisKvs.iterateItem:

- mock-redis is abandoned and no replacement for redisv4 exists, so make our own mock
- move the test to the unit tests and remove it from the github workflow

PR #2231, opened 2025-04-08
Avoid reducing redirects directly to an Object to avoid memory issue

### LINKED ISSUE #2142: Out of memory issues
Since few days, lots of WP1 tasks are now ending with OOM even for quite reasonably sized selections. See https://farm.zimit.kiwix.org for sample

benoit74:
Fix #2142 

Reducing the list of redirects directly to an Object caused abnormal memory consumption. While the exact root cause is unknown, reducing first to a Map and then transforming into a Object removes the memory issue. Since Object keys are redirect titles which might be quite big and there might be many in some occasion, we might have memory fragmentation issues or something else.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,7 @@
 Unreleased:
 
+* FIX: Avoid reducing redirects directly to an Object to avoid memory issue (@benoit74 #2142)
+
 1.14.2:
 * FIX: Ignore pages without a single revision / revid (@benoit74 #2091)
 * CHANGED: Set default log level to `log` to sufficiently verbose logs by default (@benoit74 #2121)
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -64,13 +64,11 @@ export async function getArticlesByIds(articleIds: string[], downloader: Downloa
         for (const [articleId, articleDetail] of Object.entries(mwArticleDetails)) {
           if (articleDetail.redirects && articleDetail.redirects.length) {
             await redirectsXId.setMany(
-              articleDetail.redirects.reduce((acc, redirect) => {
-                const rId = redirect.title
-                return {
-                  ...acc,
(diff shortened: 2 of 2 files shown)


kelson42:
@benoit74 Do you know when this bug has been introduced?
 
I would appreciate if:
* You check the code base for similar construct (and fix if necessary)
* Put an explanation as comment beside this code so we secure nobody is tempted to reintroduce the "bad" code pattern

Otherwise I'm fine with the PR. I guess it works like you said even this is still a pity we don't understand why we had a memory leak.

benoit74:
> Do you know when this bug has been introduced?

The bug I fixed is in `getArticlesByIds`. Code has been like this since at least 1.5.0.

Code was also similar in `getArticlesByNS` some time ago but it has been changed with https://github.com/openzim/mwoffliner/commit/8461cf03bf542fb9a72b03e812ba1becd37387e1 / https://github.com/openzim/mwoffliner/pull/1728 (released in v1.12.0) for a version which does not exhibit the issue. Commit is huge and PR does not specifically mention the problem. To me it is more accidental that the code has been changed to something which does not exhibit the issue.

I've now looked at the codebase and realized we have a "more common" way to write the code while still not having the memory issue. I wonder if the problem might come from the "unnecessary" `...acc`.  I pushed a new commit with this change instead of previous one.

Looking at the codebase, I found one other `reduce` function with same "weird" code. And many other with the "new" one. Fixed as well in same commit.

I didn't added a comment because it was hard to tell what we should not do, especially since we've not totally understood the issue + it is not the standard "pattern" used in reduce functions.

benoit74:
Nota: there is a non-negligible chance the bug is linked to a Node.JS VM rather than our own code, so it might be more recent than 1.5.0 ... don't know ... and mostly impossible to assess.

kelson42 (approved):
LGTM, thx for this catch!

Issue #2154, opened 2025-02-10
Downloader(s) should be a singleton(s)


kelson42:
to avoid passing everywhere the object

madhav2348:
@kelson42 can you give little more details on this issue

benoit74:
Currently we create a Downloader object and pass this object around in every function which needs access to this Downloader.

We want to replace this with a single Downloader singleton which could be directly imported when needed, not requiring passing the object around. Exactly like Mediawiki object is a singleton currently.

We currently have in MediaWiki.ts:

```
const mw = MediaWiki.getInstance()
export default mw as MediaWiki
```

We want to have in Downloader.ts something like:

```
const downloader = Downloader.getInstance()
export default downloader as Downloader
```

And use this Downloader singleton everywhere.

Shaikh-Mudassir:
Hey I would like to work on this issue, and solve it!
can someone please assign me this?
@benoit74

ronVK18:
Hey , I also like to work on the issue, can someone please assign me this ?
@benoit74

benoit74:
@Shaikh-Mudassir is currently in charge, no need to duplicate effort

@Shaikh-Mudassir are you making progress?

PR #2233, opened 2025-04-10
Refactor: create a Downloader singleton to stop passing object around

arjitcodes:
fix: #2154 
Create a Downloader singleton to stop passing object around

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,6 @@
 Unreleased:
 
+* UPDATE: Create Downloader singleton to stop passing object around (@arjitcodes #2154)
 * FIX: Ignore pages without a single revision / revid - continued (@benoit74 #2234)
 * FIX: Set proper mimetype for ZIM illustration(s) metadata (@benoit74 #1974)
 * FIX: Avoid reducing redirects directly to an Object to avoid memory issue (@benoit74 #2142)
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -78,38 +78,69 @@ type URLDirector = WikimediaDesktopURLDirector | WikimediaMobileURLDirector | Vi
  * Downloader is a class providing content retrieval functionalities for both Mediawiki and S3 remote instances.
  */
 class Downloader {
+  private static instance: Downloader
+  public static getInstance() {
+    if (!Downloader.instance) {
+      Downloader.instance = new Downloader()
+    }
+    return Downloader.instance
+  }
   public loginCookie = ''
-  public readonly speed: number
+  private _speed: number
   public cssDependenceUrls: KVS<boolean> = {}
-  public readonly webp: boolean = false
-  public readonly requestTimeout: number
-  public readonly basicRequestOptions: AxiosRequestConfig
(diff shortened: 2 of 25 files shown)


benoit74:
@arjitcodes please fix build issue

arjitcodes:
@benoit74 build issue is fixed now

benoit74:
CI still failing for new reasons, please fix.

arjitcodes:
@benoit74 the new issue is also fixed now. Could you run the workflow again so I can check if there are any other problems?

benoit74:
Many test errors now. 

Please consider running tests locally to confirm everything is fine, it is a mess to run CI over and over again to "hopefully" fix a test (we have 18 to fix). 

For this you need locally a `redis-cli` command + a running `redis` server + `zim-tools`. For some tests you also need S3 credentials / bucket (see my previous [comment](ttps://github.com/openzim/mwoffliner/pull/2202#issuecomment-2765686209)). I know it is a bit cumbersome to setup on first test, but probably way less than iterating over and over on the CI (unless you know exactly what the bug is).

arjitcodes:
@benoit74 tests fixed now

benoit74:
CI passed 🥳 I will review this on Monday

kelson42:
@arjitcodes Please don't merge main branch to feature branch. This make a mess in the git revision log. Use rebase instead.

src/Downloader.ts
```diff
@@ -165,7 +201,43 @@ class Downloader {
       responseType: 'stream',
       method: 'GET',
     }
-  }
+ }
+
+  public reset() {
```
benoit74, line 206: Why do we need both `isinit` and `reset`? wouldn't just allow to call `init` again be sufficient?

Besides tests, in the scraper code I see that you always call reset before calling init. Do I miss something?

benoit74 (changes requested):
Few more remarks / questions

src/Downloader.ts
arjitcodes: Yeah, good point! `isInit` was there to avoid accidental re-inits, mainly outside tests. But since only tests need re-init, it makes sense to just allow init to be called multiple times. I’ll clean it up — thanks!

benoit74: And you should indeed ensure we call init only once. But this is probably pretty easy by inspecting code. Having a method which avoids calling init twice, and bypassing it by calling reset everytime just before the function which allowed to safeguard makes no sense to me.

benoit74 (changes requested):
Mostly OK, please:
- rebase on latest main commit
- add an entry in changelog
- squash all your commits

Changelog
```diff
@@ -1,5 +1,6 @@
 Unreleased:
 
+* UPDATE: Create Downloader singleton to stop passing object around (@arjitcodes #2233)
```
benoit74, line 3: Please use issue number, not PR

benoit74 (approved):
LGTM, I will take care of merge right now.

PR #2236, opened 2025-04-11
Also rewrite Mediawiki CSS modules

### LINKED ISSUE #2214: Style modules retrieved by ActionParse url are not properly rewritten
CSS modules retrieved might contain URLs. Currently, these URLs are not rewritten and we do not retrieve the corresponding image.

Consequence on https://github.com/openzim/mwoffliner/pull/2203 is that some important assets are missing, e.g. "external link" icons.

We need to retrieve these CSS URLs and retrieve corresponding files.

benoit74:
Fix #2214 

Before this commit, only CSS fetched from main page was rewritten.

With this commit, we also rewrite CSS modules retrieved from the Mediawiki for any article.

Changes:
- extract method responsible to rewrite CSS
- also call this method when downloading/saving CSS modules 

Nota: I'm not sure what the id at the end of SVG url used in test is used for. This is maybe used for caches refresh and might be regularly rotated ... if so, then we will have to harden this test, not a big deal but no hurry to do it if this ID stays the same "forever". Anyway, skin might change at some point and not use same SVG at any point in time.

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -448,7 +448,7 @@ async function execute(argv: any) {
         return pmap(
           moduleList,
           (oneModule) => {
-            return downloadAndSaveModule(zimCreator, downloader, dump, oneModule, type as any)
+            return downloadAndSaveModule(zimCreator, downloader, oneModule, type as any)
           },
           { concurrency: downloader.speed },
         )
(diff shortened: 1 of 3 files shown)


kelson42:
> Nota: I'm not sure what the id at the end of SVG url used in test is used for. This is maybe used for caches refresh and might be regularly rotated ... if so, then we will have to harden this test, not a big deal but no hurry to do it if this ID stays the same "forever". Anyway, skin might change at some point and not use same SVG at any point in time.

Sure at 99$ this is for [cache-busting](https://www.keycdn.com/support/what-is-cache-busting) like you suggested. I don't think we will have to handle this.

kelson42:
Seems to work for me, going to merge.

PR #2248, opened 2025-04-17
Check early for availability APIs + add check for module API

### LINKED ISSUE #2246: Check that module API (load.php) is available asap
There are no check at the beginning of the scrape regarding available of the module API (`load.php`). This API is however mandatory to retrieve modules content. It is always available because used by the "real" mediawiki UI, but it might need custom `--mwModulePath` setting to adapt to local setups.

Prior https://github.com/openzim/mwoffliner/pull/2243, we were using the module API very early in the scrape. 

It is not the case anymore, and now when `--mwModulePath` is incorrect, scraper fails only after having retrieved all articles which is way too late.

We must check this API availability asap in scraper logioc.

benoit74:
Fix #2246 

Changes:
- create renderers early in scraper logic so that we fail early should we not be able to create them because of missing API
  - before this change, renderers are created when we try to get and save articles, which is quite late since we've already spent time downloading articles list (potentially from namespaces, which could take some time)
- store these renderers in a singleton
- add a check for module API which is needed by all renderers

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -4,6 +4,7 @@ Unreleased:
 * FIX: Ignore pages without a single revision / revid - continued (@benoit74 #2234)
 * FIX: Set proper mimetype for ZIM illustration(s) metadata (@benoit74 #1974)
 * FIX: Avoid reducing redirects directly to an Object to avoid memory issue (@benoit74 #2142)
+* NEW: Check early for availability of load.php module API (@benoit74 #2246)
 
 1.14.2:
 * FIX: Ignore pages without a single revision / revid (@benoit74 #2091)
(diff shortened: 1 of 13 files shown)


kelson42:
@benoit74 There is no way to do that earlier within all validations tests?

benoit74:
@kelson42  indeed, sorry for missing that, got confused by `process.on` instructions.

Check moved earlier + rebased on main

benoit74:
@kelson42 ready for a new review. As discussed I moved these checks in a more "standard" place. As suspected, it forced me to realize that checks of other APIs were done too late, i.e. after the list of articles have been retrieved from the namespaces or from passed article lists. I refactored the whole thing with a new singleton holding renderers as soon as they are created, and allowing to create them early.

kelson42 (approved):
LGTM

PR #2270, opened 2025-04-28
Keep `typeof` HTML attribute and stop adding custom thumbiner HTML around figures

### LINKED ISSUE #2254: [ActionParse] All images are surended by a one pixel grey border
See for example here at https://dev.library.kiwix.org/viewer#wikiquote_zh_all_maxi_2025-04/%E7%BE%85%E8%B3%93%E5%BE%B7%E6%8B%89%E7%B4%8D%E7%89%B9%C2%B7%E6%B3%B0%E6%88%88%E7%88%BE

Image

Whereas online https://zh.wikiquote.org/wiki/%E7%BE%85%E8%B3%93%E5%BE%B7%E6%8B%89%E7%B4%8D%E7%89%B9%C2%B7%E6%B3%B0%E6%88%88%E7%88%BE

Image

It's no the worse bug, but this is quite visible because it applies to all images and change slightly the layout surrounding the image. I also kind of suspect this is a symoptom of something which is more fundamentaly wrong. Therefore I'm a bit concerned.

benoit74:
Fix #2254

The `typeof` HTML attribute must be kept because it is used for some CSS rules.

image

Previous scraper logic was adding custom HTML around figures and spans. This does not seem to be necessary anymore with `ActionParse` API, and even leads to visual glitches.

Both changes are done only for `ActionParse` API since it seems they were needed with other renderers (or at least they "combined" well with ugly CSS which was injected)

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -10,6 +10,7 @@ Unreleased:
 * FIX: CSS dependencies are not downloaded/pushed to the ZIM on all but first requested flavor (@benoit74 #2249)
 * FIX: Links to articles through index.php are not detected properly (@benoit74 #2260)
 * FIX: Properly sanitize mw* settings (@benoit74 #2269)
+* CHANGED: Keep `typeof` HTML attribute and stop adding custom thumbiner HTML around images when ActionParse renderer is used (@benoit74 #2254)
 
 1.14.2:
 * FIX: Ignore pages without a single revision / revid (@benoit74 #2091)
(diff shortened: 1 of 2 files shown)


kelson42:
@benoit74 i would really prefer that you remove the part after the "return" . I believe we have 50% chance that these insteuctions are deprecated. Would really prefer you check we other API backend that thimbs are sstill rendered properly.

benoit74:
Removing the part after the "return" does not work at all with other APIs, these changes are needed by them. Don't want to loose time investigating something which might be history in few weeks.

kelson42 (approved):
LGTM, please rebase an merge

PR #2274, opened 2025-04-28
Use only minimal Vector (legacy and 2022) templates to avoid side-effects from CSS reserved spaces

### LINKED ISSUE #2259: [Actionparse] Strange skin for rationalwiki latest scrape
All pages of Rationawiki actionparse based scrape are really good, but suffer of stranges footer and header (I guess from the skin). See for https://dev.library.kiwix.org/viewer#rationalwiki_en_all_maxi_2025-04/RationalWiki

Large padding at top

Image

and kind of empty box at the bottom

Image

benoit74:
Fix #2259 

When pushing vector (legacy and 2022) templates into the scraper, I wondered whether it might help or not to keep siblings of the HTML tree nodes we are really using to render the article content. 

I did not had any strong argument to push in any direction (keep them or remove them) at that time and hence considered it would be easier to keep them until something proves that we should remove them, rather than removing them and then having to reverse-engineer again the Mediawiki template (which is kinda a pain in the ass).

We are now at a point where issue #2259 proves that keeping these siblings have bad side-effects, since there is some CSS rules which kinda "reserve space" for them. Let's remove them then. Might be that we will need them somewhere in the future when we will add support for the content of these siblings, but then they will not be empty anymore and it will make much more sense.

Problem seems to be more visible on vector legacy template than on vector-2022, but I preferred to apply same approach on both templates for consistency.

Random rationalwiki page after the change:

Screenshot 2025-04-28 at 11 29 10

Screenshot 2025-04-28 at 11 29 20

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -12,6 +12,7 @@ Unreleased:
 * FIX: Properly sanitize mw* settings (@benoit74 #2269)
 * CHANGED: Keep `typeof` HTML attribute and stop adding custom thumbiner HTML around images when ActionParse renderer is used (@benoit74 #2254)
 * FIX: Stop removing all images which should not be editable (@benoit74 #2255)
+* FIX: Use only minimal Vector (legacy and 2022) templates to avoid side-effects from CSS reserved spaces (@benoit74 #2259)
 
 1.14.2:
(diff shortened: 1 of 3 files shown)


kelson42 (approved):
LGTM, please rebase and merge

PR #2290, opened 2025-05-08
Fix issues when article is moved with redirect left behind

### LINKED ISSUE #2278: Fix issues when article is moved with redirect left behind
```
Error: Impossible to add C/Alejandro_González_y_Robleto
  dirent's title to add is : Alejandro González y Robleto
  existing dirent's title is : Alejandro González y Robleto

    at <anonymous> (/tmp/mwoffliner/src/mwoffliner.lib.ts:474:22)
    at Immediate.workerFunc (/tmp/mwoffliner/src/util/RedisKvs.ts:124:17)
    at process.processImmediate (node:internal/timers:487:21)

```

From https://farm.openzim.org/pipeline/66a197ea-6bbd-4cc2-ab0f-8e803d593db0/debug

benoit74:
Fix #2278 

Changes:
- change ActionParse query to follow redirects
- benefit from ActionParse redirects result to detect when a page has been moved since our listing of articles and include redirect from new location (where we will have nothing) to old location
- detect when page is moved during the listing as well, and recover properly (do not fetch new page and add a redirect from new location to old one)
- when writing redirects to the ZIM, ensure that we are not redirecting to self, not redirecting from a path for which we already have an article, and not redirecting to a path which is not an article
- fix issue (only impacting tests) where URL directors where not reinitialized now that Downloader is a singleton (we were first URLDirector init in any test, and hence using bad URLs throughout most tests)

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -17,6 +17,7 @@ Unreleased:
 * CHANGED: Log ZIM location at the end of the scrape for conveniency (@benoit74 #2297)
 * FIX: ActionParse: enhance fixed style management + add external link icon in footer (@benoit74 #2276 #2283)
 * FIX: WikimediaMobile: fix wrong CSS/JS links when article has a slash in its title (@benoit74 #2293)
+* FIX: Fix issues when article is moved with redirect left behind (@benoit74 #2278)
 
 1.14.2:
 * FIX: Ignore pages without a single revision / revid (@benoit74 #2091)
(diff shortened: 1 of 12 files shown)


kelson42:
@benoit74 A bit difficult to 100% follow what tou have done, but is that a correct understanding that this PR creates - whatever the end-point used - a redirection in the opposite direction of ehat is is in the online wiki (when upstream an article is transformed in a redirect between article titles retrieval and contents retrieval)?

benoit74:
>  is that a correct understanding that this PR creates - whatever the end-point used - a redirection in the opposite direction of the one which is on online wiki (when upstream an article is transformed in a redirect between article titles retrieval and contents retrieval)?

Yes, but not under all conditions, it does so only if a redirect was left behind (other situation is handled in the other PR) and :
- for all APIs when article was moved during the listing and the new article was fetched in this listing as well
- for ActionParse API when article is moved after the listing, while fetching article contents, before retrieving the moved article (because we have sufficient info only with this API + it was complex / impossible to implement in a generic manner)

Issue #2213, opened 2025-04-01
Allow scraper user to choose skin to use


benoit74:
Mediawikis can have multiple skins. On some occasion, one might want to use a specific skin rather than the one selected by default detection mechanism introduced in https://github.com/openzim/mwoffliner/pull/2203

We probably need a CLI parameter, but then we need to check this skin is really available...

benoit74:
Note that this is also required to properly test `vector-legacy` skin in e2e tests. As of today, this skin is not tested because most wikis have been updated and now have both Vector skins available. And by default we prefer `vector-2022`.

When implementing this feature, we should add an e2e test (on en.wikipedia.org for instance, i.e. in `en.e2e.test.ts`)  to test both `vector` and `vector-2022` skins

benoit74:
I confirm this is needed for https://github.com/openzim/zim-requests/issues/418 where there are many potential skins. Or we need to give Minerva higher precedence over Vector, but I'm not sure it will be future-proof

PR #2293, opened 2025-05-08
ActionParse: enhance fixed style management + add external link icon in footer

### LINKED ISSUE #2276: Move custom CSS out of skin HTML template
Currently, both vector skins have some custom CSS embedded in the HTML template file (res/templates/pageVector2022.html and res/templates/pageVectorLegacy.html).

Should we move this CSS to a standalone file?

@kelson42, arguments on why we must do this are welcomed.

On my end, I do not have any real arguments besides this would make processing logic more complex (we will have to manage both an HTML template + associated CSS file). And I do not expect huge gain from this move, but I might miss something.

### LINKED ISSUE #2283: [actionparse] (external link) icon is not applied properly in the footer
The footer of articles (with link to upstream article) is ignored (no external link visual on these links. 

Image

Par exemple avec https://dev.library.kiwix.org/viewer#wikipedia_fr_medicine-app_maxi_2025-05/Trouble_de_la_personnalit%C3%A9_borderline

benoit74:
Fix #2276
Fix #2283 

Changes:
- move custom CSS for footer in dedicated CSS file
- add SVG icon for these external links
- remove stuff around text direction because this is already applied automatically at mostly the top of the document (we already fetch the information from Mediawiki metadata, no need to ask translators for this info)
- move custom CSS linked to the skin in dedicated CSS files

Screenshot 2025-05-08 at 14 27 36

Screenshot 2025-05-08 at 14 30 04

NOTA: the "external link" icon in arabic is not perfectly positioned (should be on the left rather than on the right), but this seems to be a general issue both at ar.wikipedia.org and in our Wikipedia ZIM in arabic, in the whole documents, so I do not mind much about this.


benoit74:
I modified the `en.e2e.test` to use a page with a `/` (slash) and a ` ` (space). This better reflect the real difficulties we might encounter in the wild.

Note that however this meant we test only one skin (the one on en.wikipedia.org, vector-2022 as of today). 

I'm not sure how to test vector-legacy skin, since this skin is less in use and going to disappear. This will be easier once https://github.com/openzim/mwoffliner/issues/2213 is implemented (I've left a comment over there to not forget to add these test cases).

Note that I had to fix a bug in `WikimediaMobile` API as well. And this still has to be rebased, but I kept original commit for easier review.

benoit74:
Github refused to run CI until checks have been resolved ... this is new? Anyway, I force pushed ...

kelson42 (approved):
LGTM

PR #2295, opened 2025-05-08
Replace urlparser deprecated API with WHATWG URL API

### LINKED ISSUE #2285: (node:14) [DEP0170] DeprecationWarning: The URL xxx is invalid. Future versions of Node.js will throw an error.
We regularly have message like `(node:14) [DEP0170] DeprecationWarning: The URL xxx is invalid. Future versions of Node.js will throw an error.`

See e.g. https://farm.openzim.org/pipeline/dcdbd892-f2ac-4153-a188-1f621212939f/debug

URLs are really invalid. We should check if this is going to be fine once Node.js will really throw an error. I'm not sure we have sufficient information yet to know where there warnings are raised.

benoit74:
Fix #2285


kelson42 (approved):
LGTM although would be extremly surprised if this change does not need a few additional fixes to be stabelised. Touchy!

PR #2296, opened 2025-05-08
Add prettier to CI, so that we fail early on formatting issues

benoit74:
So far, the CI does not checks code formatting with prettier. 

This means that we regularly have commits which reformat something totally unrelated to the change simply because previous committer did not formatted files, and current author did.

This PR adds prettier (in check mode, to fail on error) to the CI.

This PR also adds required pre-commit configuration and instruction, so that users wishing to can check their code automatically before pushing it to the repo (prettier plugin is archived and supposedly broken but it still works fine ATM, waiting for the community to restart work on it should it break).

NOTA : pre-commit is a Python tool, coming mostly from the Python community, but it works very well on Node.JS projects and has mostly no requirements beside a Python runtime

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,13 @@ jobs:
       - name: Installing other dependencies
         run: |
           npm install
+          npm i -g prettier
           npm i -g eslint
           npm i -g codecov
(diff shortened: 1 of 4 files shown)


kelson42 (approved):
Don't have checked everything in detail, but supportive of the principle and implementation LGTM

PR #2300, opened 2025-05-11
Replace supported Node.js 22 by 24

kelson42:
* Replace Node.js 22 by 24 to the CI
* Replace Node.js 22 by 24 as supported engine

Depends on https://github.com/openzim/node-libzim/pull/162


benoit74 (changes requested):
PR title is wrong, it is not what is inside this PR.

We agreed that we will support only one Node.JS major version (or Python) on scrapers since these are end-products. So this PR should drop support of Node 22. And this should be implemented only once node-libzim supports Node 24 and we agree this version is stable enough (which probably is since it is now GA).

kelson42:
> We agreed that we will support only one Node.JS major version

I'm against applying this like this here. You mostly decide this for Python scrapers and I didn't oppose but I don't think this is a good policy because it creates artificially incompatibilities with the stack of developer. To me as long as this does not create a burden, I don't see any reason to stop supporting an older Node.js.

benoit74:
The CI also failed on `main` branch, so it is unrelated to this PR (most probably linked to an upstream change): https://github.com/openzim/mwoffliner/actions/runs/14955761328/job/42011187351

> To me as long as this does not create a burden, I don't see any reason to stop supporting an older Node.js.

It does create an associated burden de-facto: 
- two CI to wait for: more computing time needed and twice more chances of having a flaky test fail
- always a question when a bug appears: always need to mention/ask which Node.JS was used, ask yourself if it might be linked to the Node.JS version, ...

In today world, have multiple Node/Python versions in parallel on a single workstation is a nobrainer, so I don't see where it is a real burden for developers.

Since we are the main (mostly only) users of the scrapers, since we do provide a Docker image built on a single Node version, since I'm in charge of all scrapers and since the Python policy proved to cause 0 reported issue since two years in the wild, I'm strongly in favor of applying this same policy on all scrapers.

kelson42:
Put as draft to "avoid" merging

benoit74 (approved):
LGTM

PR #2309, opened 2025-05-16
Split CI and fix coverage report issues

### LINKED ISSUE #2060: Fix code coverage reporting
In #2055 I got a notification that the line I changed was missing coverage:

image

However when I added a log statement [here](https://github.com/openzim/mwoffliner/blob/4e572ac34b71695bb34b9966928c30c58e0f42d8/src/mwoffliner.lib.ts#L436) I could see that the code was being executed by [this test](https://github.com/openzim/mwoffliner/blob/4e572ac34b71695bb34b9966928c30c58e0f42d8/test/unit/webpAndRedirection.test.ts#L21)

Perhaps e2e tests are not being counted for coverage?

benoit74:
Fix #2060 

Split CI into distinct e2e-tests, unit-tests and code-quality jobs so that it is both faster, clearer and more complete.

Fix coverage issues linked to e2e tests overwriting unit tests coverage. It is mandatory to produce coverage results in distinct directories and then merge them.


benoit74:
I had a look at the results and it seems pretty good:
- looks like we've solved the problem of coverage not including stuff which is covered
- looks like codecov for some reasons better detects code which is only partially covered (e.g. condition only partially tested)

Overall feedback is also obviously way clearer now that stuff is split.

Issue #2310, opened 2025-05-17
Wikipedia articles scraped with ActionParse API contain scripts that attempt to load external content


Jaifroid:
There are a bunch of attempts (blocked by CORS in the JS apps) to load scripts from links that look like `https://en.wikipedia.org/w/load.php?debug=2&lang=en&modules=mediawiki.String&skin=vector-2022&version=ztntf`. The Ray Charles actionparse test ZIM attempts to load no fewer than 23 external resources or modules. You can see some of them in this screenshot. Same for KJS Browser Extension. I suppose these external script links should be rewritten or removed at scrape time to ensure no external content is accessed. I don't think we can or should rely on CORS (which may not be applied in all contexts) to do the work of removing external content for us.

![Image](https://github.com/user-attachments/assets/7cd5a3bd-8626-45be-a388-4fac0b53887e)

Jaifroid:
Further info: These loads originate in the included script `startup.js`, where a couple of external URLs are hardcoded:

![Image](https://github.com/user-attachments/assets/01afb748-51c9-4d35-aada-5cd394b28328)

kelson42:
@Jaifroid Thank you! On my side my question would be: if confirmed why the CI (zimcheck) didn't detect this serious problem?

Jaifroid:
@kelson42 I'm working from the Ray Charles test ZIM that @benoit74 kindly provided in #2263. That was around 3 weeks ago, so the issue would need to be verified in relation to code changes since then. I searched but couldn't find any open or closed issue indicating this or a similar issue has been noticed and addressed yet.

kelson42:
@Jaifroid I just have created a new ZIM https://farm.openzim.org/pipeline/4ddb96e7-c2c4-49e2-8d24-49a0d7c3fbf4

Jaifroid:
@kelson42 Thank you very much for making this ZIM so quickly! I'm still seeing this issue in console ☹️:

![Image](https://github.com/user-attachments/assets/e731d596-2570-4f61-8447-fef85aef5b42)

All these calls are blocked by CORS in line 1067 of the `addScript` function defined in `startup.js` (see screenshot below). This takes the URLs for the modules to load (which are composed of the prefix in line 2601 shown in [screenshot above](https://github.com/openzim/mwoffliner/issues/2310#issuecomment-2888283629) plus the module suffix, and then tries to inject them into the DOM as script elements using `document.head.appendChild( script );`. This is then blocked by CORS.

[...]

PR #2319, opened 2025-05-20
Do not add JS scripts/modules in the ZIM with ActionParser renderer for now

benoit74:
This is a "containment" change for #2310, not the long term fix.

See https://github.com/openzim/mwoffliner/issues/2310#issuecomment-2894221966 for details


kelson42 (approved):
OK, please rebase and merge

Issue #2318, opened 2025-05-20
terminate called after throwing an instance of 'Xapian::InvalidArgumentError'


benoit74:
See https://farm.openzim.org/pipeline/c09c6bed-2afe-44d7-bcab-69c081dc340c from https://farm.openzim.org/recipes/wikipedia_ar_all

Looks like we have a low level issue. I hope this is not a "random" one, but linked to some article content.

I've just restarted the recipe with `info` logs so that we have a chance to know on which article this happens.

benoit74:
New task failed again.

Last log lines are:

```
[log] [2025-05-21T02:16:59.496Z] Progress downloading articles [639870/1263308] [50.7%]
...
[info] [2025-05-21T02:18:29.823Z] Worker [8] processing batch of article ids [["نظرية_رابطة_التكافؤ","+8 more +","قائمة_مدن_أوكلاهوما"]]
[info] [2025-05-21T02:18:29.823Z] Getting article [نظرية_رابطة_التكافؤ] from https://ar.wikipedia.org/w/api.php?action=parse&format=json&prop=modules%7Cjsconfigvars%7Cheadhtml%7Ctext&parsoid=1&page=%D9%86%D8%B8%D8%B1%D9%8A%D8%A9_%D8%B1%D8%A7%D8%A8%D8%B7%D8%A9_%D8%A7%D9%84%D8%AA%D9%83%D8%A7%D9%81%D8%A4&useskin=vector-2022&redirects=1
[info] [2025-05-21T02:18:29.823Z] Getting JSON from [https://ar.wikipedia.org/w/api.php?action=parse&format=json&prop=modules%7Cjsconfigvars%7Cheadhtml%7Ctext&parsoid=1&page=%D9%86%D8%B8%D8%B1%D9%8A%D8%A9_%D8%B1%D8%A7%D8%A8%D8%B7%D8%A9_%D8%A7%D9%84%D8%AA%D9%83%D8%A7%D9%81%D8%A4&useskin=vector-2022&redirects=1]
[...]

PR #2320, opened 2025-05-21
Truncate article titles to 245 bytes

benoit74:
This is an interim bug fix for #2318 until upstream issue settles (or maybe a definitive fix should we discover this is just a documentation issue)

Changes:
- truncate article titles to 245 bytes in an efficient manner


benoit74:
Sorry, I just realized that this PR contained code intended for another PR. Fixed.

kelson42 (approved):
Very ugly but I get we have no real choice for the moment if we want that ZIM to finally pass!

PR #2323, opened 2025-05-21
Handle empty HTTP response

### LINKED ISSUE #2322: Handle unknown 500 error and replace with placeholder
For https://github.com/openzim/zim-requests/issues/1282, we have a bad 500 error on https://www.pokepedia.fr/api.php?action=parse&format=json&prop=modules%7Cjsconfigvars%7Cheadhtml%7Ctext&parsoid=1&page=Liste_des_Pok%C3%A9mon_dans_l%27ordre_du_Pok%C3%A9dex_Myrtille&useskin=vector-2022&redirects=1

The HTTP return code is 500 but the response is empty (content-length=0).

Looks like we should catch such 500 errors when response is empty and replace them with a placeholder since this is not a transient error.

benoit74:
Fix #2322


kelson42 (approved):
LGTM

PR #2371, opened 2025-06-19
Fix site info usage and fix illustration resizing

### LINKED ISSUE #2368: Illustration_48x48@1 is not 48x48 when source image is not square
When source image is not square, `Illustration_48x48@1` is not a 48x48 image.

See for instance https://farm.openzim.org/recipes/wikipedia_en_knots which uses https://upload.wikimedia.org/wikipedia/commons/thumb/1/16/Grief_knot.svg/330px-Grief_knot.svg.png as icon.

Resulting image is at https://dev.library.kiwix.org/raw/wikipedia_en_knots_nopic_2025-06/meta/Illustration_48x48@1 (currently ...) and is a 48x28 PNG.

I believe this should (must) be a 48x48 PNG, with transparent background.

benoit74:
Fix #

Fix #2368

Simply changed resize method + specified a white (most commonly "invisible") and transparent background. See doc at https://sharp.pixelplumbing.com/api-resize/

Consequence on kiwix-serve:

| Before | After |
|--|--|
| Screenshot 2025-06-19 at 14 09 12 | Screenshot 2025-06-19 at 14 09 32 |


kelson42:
> do not retrieve additional siprop (site info properties) since we need only general one at this stage (don't know why we requested so many metadata, I double checked multiple times and it is unused ... probably a leftover from previous scraper behaviors)

Have you checked that the retrieval of latitute and longitude still work fine. Might be relate.

benoit74:
> Have you checked that the retrieval of latitute and longitude still work fine. Might be relate.

No I did not, I'm not sure what you are speaking about.

If you speak about the fact that we add a meta tag in article HTML `<head>` with lat/lot, then this is still working:

image

PR #2375, opened 2025-06-20
Retrieve CSS from CSS-only gadgets automatically

### LINKED ISSUE #2212: Properly include styles from CSS-only gadgets
With https://github.com/openzim/mwoffliner/pull/2203, we've introduced basic / simplified set of "core" JS/CSS modules.

While `startup` JS module is not a surprise, we do not yet totally understand why it is necessary to "force" the retrieval of `ext.gadget.infobox,wikiMenuStyles` for proper infobox rendering.

Test page : `Brian_May` on hu.wikipedia.org

benoit74:
Fix #2212 

Until https://phabricator.wikimedia.org/T161278 is solved, we implement this workaround to automatically add CSS-only gadget modules to appropriate  pages. This is done in a way which is compatible with T161278 being solved, i.e. the scraper will still behave correctly once this issue is solved upstream.

Changes:
- at scraper begin, retrieve list of all gadgets and store it in memory
- use this list to add more style modules to download / add to the article HTML based on article properties
- these new modules are then automatically downloaded just like any style module
- remove "hacks" which are not needed anymore now that everything is automatically retrieved (only in ActionParse renderer, details of other hacks are too unknown

Other more or less related changed:
- stop passing `debug=true` when fetching modules, this causes bigger than necessary CSS/JS
- stop passing `lang=en` when fetching modules, this is wrong on all non-EN wikis ; using default is assumed to be more correct
- stop passing weird `version=&*` which seems unnecessary and only causing confusion
- request `info` when fetching list of articles by NS, so that we can have their `contentmodel`

Nota: inclusion of a gadget on a specific article page can be restricted based on article namespace and contentmodel. Code has been adapted to take this into account, but this has not been tested in the wild with a real gadget really applying only to one namespace or one module, since I don't know where to find them. Should a bug appear, the fix will anyway be minimal. In most cases, gadgets apply to all namespace and all contentmodel, and this works fine.

### DIFF
--- a/res/templates/pageVector2022.html
+++ b/res/templates/pageVector2022.html
@@ -1,13 +1,21 @@
-<!DOCTYPE html>
-<html class="client-js" __ARTICLE_LANG_DIR__>
+<!doctype html>
+<html class="client-nojs" __ARTICLE_LANG_DIR__>
   <head>
-    <meta charset="UTF-8"/>
+    <meta charset="UTF-8" />
     <title></title>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0"/> __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_LIST__
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css">
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector-2022.css">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_BEFORE_META__
+    <meta name="ResourceLoaderDynamicStyles" content="" />
+    __ARTICLE_CSS_AFTER_META__
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__mw/site.styles.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__mw/noscript.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector-2022.css" />
   </head>
-  <body class="skin--responsive skin-vector skin-vector-search-vue mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject mw-editable skin-vector-2022 action-view uls-dialog-sticky-hide" cz-shortcut-listen="true">
+  <body
+    class="skin--responsive skin-vector skin-vector-search-vue mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject mw-editable skin-vector-2022 action-view uls-dialog-sticky-hide"
+    cz-shortcut-listen="true"
+  >
     <div class="mw-page-container">
       <div class="mw-page-container-inner">
         <div class="mw-content-container">
@@ -24,8 +32,7 @@
           </main>
         </div>
       </div>
-    </div>    
-    __ARTICLE_CONFIGVARS_LIST__
-    __ARTICLE_JS_LIST__
+    </div>
+    __ARTICLE_CONFIGVARS_LIST__ __ARTICLE_JS_LIST__
   </body>
 </html>
--- a/res/templates/pageVectorLegacy.html
+++ b/res/templates/pageVectorLegacy.html
@@ -1,12 +1,16 @@
-<!DOCTYPE html>
-<html class="client-js" __ARTICLE_LANG_DIR__>
+<!doctype html>
+<html class="client-nojs" __ARTICLE_LANG_DIR__>
   <head>
     <meta charset="UTF-8" />
     <title></title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_LIST__
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css">
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector.css">
+    __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_BEFORE_META__
+    <meta name="ResourceLoaderDynamicStyles" content="" />
+    __ARTICLE_CSS_AFTER_META__
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__mw/site.styles.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__mw/noscript.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css" />
+    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector.css" />
   </head>
 
   <body
--- a/src/Gadgets.ts
+++ b/src/Gadgets.ts
@@ -1,0 +1,108 @@
+import Downloader from './Downloader.js'
+import * as logger from './Logger.js'
+import MediaWiki from './MediaWiki.js'
+
+class Gadgets {
+  private static instance: Gadgets
+  private gadgets: Gadget[] | undefined
+
+  public static getInstance() {
+    if (!Gadgets.instance) {
+      Gadgets.instance = new Gadgets()
+    }
+    return Gadgets.instance
+  }
+
+  public async fetchGadgets() {
+    const gadgetsUrl = Downloader.apiUrlDirector.buildGadgetsUrl()
+    const gadgetsResponse = await Downloader.getJSON<GadgetQueryResult>(gadgetsUrl)
+    if (!gadgetsResponse.batchcomplete) {
+      throw new Error(`Error while fetching gadgets with ${gadgetsUrl}, batch seems to be incomplete, scraper does not expect/support pagination here`)
+    }
+    this.gadgets = gadgetsResponse.query?.gadgets
+    logger.info(this.gadgets === undefined ? 'Gadgets are not supported on this wiki' : `${this.gadgets.length} gadgets retrieved`)
+  }
+
+  /*
+   * Get list of gadgets which are not listed on action=parse results
+   */
+  public getActiveGadgetsByType(articleDetail: ArticleDetail) {
+    const cssGadgets = []
+    const jsGadgets = []
+
+    const activeGadgets = (this.gadgets || []).filter((gadget) => {
+      const settings = gadget.metadata.settings
+      if (settings.skins && settings.skins.length && !settings.skins.includes(MediaWiki.skin)) return false
+      if (settings.actions && settings.actions.length && !settings.actions.includes('view')) return false
+      if (settings.namespaces && settings.namespaces.length && !settings.namespaces.includes(articleDetail.ns || 0)) return false
+      if (settings.contentModels && settings.contentModels.length && !settings.contentModels.includes(articleDetail.contentmodel || 'wikitext')) return false
+      return true
+    })
+
+    activeGadgets.map((gadget) => {
+      const module = gadget.metadata.module
+      if (module.peers && module.peers.length) {
+        // Only JS Gadgets can have peers
+        cssGadgets.concat(module.peers)
+        return jsGadgets.push(gadget.id)
+      }
+      if (module.scripts && module.scripts.length) return jsGadgets.push(gadget.id)
+      if (module.dependencies && module.dependencies.length) return jsGadgets.push(gadget.id)
+      if (module.datas && module.datas.length) return jsGadgets.push(gadget.id)
+      if (module.messages && module.messages.length) return jsGadgets.push(gadget.id)
+      // Anything left now is a CSS-only Gagdet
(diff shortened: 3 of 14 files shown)


src/Gadgets.ts
```diff
@@ -0,0 +1,90 @@
...
+    return Gadgets.instance
+  }
+
+  public async fetchGadgets() {
+    const gadgetsUrl = Downloader.apiUrlDirector.buildGadgetsUrl()
+    const gadgetsResponse = await Downloader.getJSON<GadgetQueryResult>(gadgetsUrl)
+    if (!gadgetsResponse.batchcomplete) {
+      throw new Error(`Error while fetching gadgets with ${gadgetsUrl}, batch seems to be incomplete, scraper does not expect/support pagination here`)
+    }
+    this.gadgets = gadgetsResponse.query.gadgets
```
Markus-Rost: Gadgets is an extension, it might not be installed on all wikis. Expect this request to return only warnings about unknown parameters, without a `gadgetsResponse.query` object:
```json
{
  "batchcomplete": true,
  "warnings": {
    "main": {
      "warnings": "Unrecognized parameters: gaallowedonly, gaenabledonly, gaprop."
    },
    "query": {
      "warnings": "Unrecognized value for parameter \"list\": gadgets"
    }
  }
}
```

src/renderers/action-parse.renderer.ts
```diff
@@ -116,6 +113,11 @@ export class ActionParseRenderer extends Renderer {
       elem.remove()
     })
 
+    // Add CSS-only gadgets which are used on this article
+    Gadgets.getCssOnlyGadgets(articleDetail).map((gadgetId) => {
+      moduleDependencies.styleDependenciesList.push(`ext.gadget.${gadgetId}`)
```
Markus-Rost, line 132: CSS-only gadgets should not be added directly to the normal style dependencies to ensure the correct load order. The gadgets are loaded after the modules listed in action=parse and styles included in JS modules, but before the general `site.styles` module. Ideally the gadget list also gets sorted alphabetically. 

The resulting HTML should look like this:
```html
<link rel="stylesheet" href="/load.php?modules=${styleDependenciesList.join('|')}&only=styles&skin=vector">
<script async="" src="/load.php?modules=startup&only=scripts&raw=1&skin=vector"></script>
<meta name="ResourceLoaderDynamicStyles" content="">
<link rel="stylesheet" href="/load.php?modules=ext.gadget.${cssGadgetsList.join(',')}&only=styles&skin=vector">
<link rel="stylesheet" href="/load.php?modules=site.styles&only=styles&skin=vector">
```

src/Gadgets.ts
```diff
@@ -0,0 +1,90 @@
...
+    return (this.gadgets || [])
+      .filter((gadget) => {
+        const module = gadget.metadata.module
+        const settings = gadget.metadata.settings
+        if (module.scripts.length || module.datas.length || module.dependencies.length || module.peers.length || module.messages.length) return false
+        if (!module.styles.length) return false
+        if (settings.skins.length && !settings.skins.includes(MediaWiki.skin)) return false
+        if (settings.actions.length && !settings.actions.includes('view')) return false
+        if (settings.namespaces.length && !settings.namespaces.includes(articleDetail.ns || 0)) return false
+        if (settings.contentModels.length && !settings.contentModels.includes(articleDetail.contentmodel || 'wikitext')) return false
```
Markus-Rost: actions, namespaces and contentModels are all fairly new options for gagdets. Wikis running older version of MediaWiki might not have those keys yet.

src/Gadgets.ts
```diff
@@ -0,0 +1,90 @@
...
+    const gadgetsUrl = Downloader.apiUrlDirector.buildGadgetsUrl()
+    const gadgetsResponse = await Downloader.getJSON<GadgetQueryResult>(gadgetsUrl)
+    if (!gadgetsResponse.batchcomplete) {
+      throw new Error(`Error while fetching gadgets with ${gadgetsUrl}, batch seems to be incomplete, scraper does not expect/support pagination here`)
+    }
+    this.gadgets = gadgetsResponse.query.gadgets
+    logger.info(`${this.gadgets.length} gadgets retrieved`)
+  }
+
+  public getCssOnlyGadgets(articleDetail: ArticleDetail) {
```
Markus-Rost: This doesn't account for `peers` dependencies as described in https://github.com/openzim/mwoffliner/issues/2212#issuecomment-2978175587 yet.

Markus-Rost (changes requested):

src/util/builders/url/api.director.ts
```diff
@@ -55,4 +55,11 @@ export default class ApiURLDirector {
       .setQueryParams({ action: 'parse', format: 'json', prop: 'modules|jsconfigvars|headhtml', formatversion: '2', page: articleId })
       .build()
   }
+
+  buildGadgetsUrl() {
```
Markus-Rost, line 59: The API queries `buildGadgetsUrl`, `buildSiteInfoURL` and `buildNamespacesURL` are all only fetched once in `mwoffliner.lib.ts`, those three requests could be combined into a single one.

src/Gadgets.ts
Markus-Rost: Actually, as long as we don't load JS due to #2310, we can just ignore the peers as well. Only support for JS modules is added, we will need to support peers as well

src/renderers/action-parse.renderer.ts
Markus-Rost: As long as we don't load JS modules due to #2310, we should also add `<link rel="stylesheet" href="/load.php?modules=noscript&only=styles&skin=vector">` after `site.styles`

src/Gadgets.ts
benoit74: Fixed, thank you

src/Gadgets.ts
benoit74: Let's add them, I don't think it will do any harm, and I'm pretty sure we have high chances to forget the peers once we add support fo JS modules.

Markus-Rost: Considering JS gadgets aren't returned by action=parse either (same upstream cause), we can just check the settings for all gadgets and then just split the resulting list into CSS and JS gadgets. Any enabled JS gadgets with peers add those to the list of CSS gadgets to load.

src/Gadgets.ts
benoit74: Fixed, I assumed all might be missing for simplicity.

src/renderers/action-parse.renderer.ts
benoit74: Few questions:
- what is `ResourceLoaderDynamicStyles` meta?
- is it really important to have regular styles dependencies before JS startup?
- currently we have one file per module (or CSS-only gadget), so that we can handle any combination since we do not generate them on-the-fly ; it is both simpler to handle and more storage efficient ; is there any downside to this approach?

What I would prefer to do is simply to implement a sorting function so that modules are always in proper order. We add all modules to the list, and just at rendering time we sort them in proper order. WDYT?

src/util/builders/url/api.director.ts
benoit74: I've pushed the matter to a distinct issue: https://github.com/openzim/mwoffliner/issues/2385

Both for clarity of review and because I'm not yet sure about the exact impact of this move. And anyway, it is not really an urgent change from my perspective, more a kind of optimization

src/renderers/action-parse.renderer.ts
Markus-Rost: - Any CSS included in JS modules are placed as `<style>` elements before the `ResourceLoaderDynamicStyles` marker when the module is loaded (through `startup`)
- No, but they need to be before the meta marker to ensure the correct order. 
- No, doing it that way should be fine.

src/Gadgets.ts
benoit74: This is what I did indeed.

src/Gadgets.ts
Markus-Rost: As the query only received warnings with no data returned, the `query` key is already missing (see the JSON above). So your current check if gadgets is installed would still throw a `TypeError: Cannot read properties of undefined (reading 'gadgets')`

(After #2385 we would always have the `query` key (because siteinfo always exists) and only the `gadgets` key missing.)

src/Gadgets.ts
```diff
@@ -0,0 +1,110 @@
...
+    // Build list of CSS-only gadgets
+    const cssOnlyGadgets = this._getActiveGadgets(articleDetail)
+      .filter((gadget) => {
+        const module = gadget.metadata.module
+        return module.styles.length && !module.scripts.length && !module.datas.length && !module.dependencies.length && !module.messages.length
+      })
+      .map((gadget) => gadget.id)
+
+    return [...cssOnlyGadgets, ...peersGadgets]
+  }
```
Markus-Rost: We can just go once through all the active gadgets and split everything between CSS and JS gadgets
```suggestion
  public getAdditionalGadgets(articleDetail: ArticleDetail) {
    const cssGadgets = []
    const jsGadgets = []
    this._getActiveGadgets(articleDetail)
      .map((gadget) => {
        const module = gadget.metadata.module
        if (module.peers && module.peers.length) {
          // Only JS Gadgets can have peers
          cssGadgets.concat(module.peers)
          return jsGadgets.push(gadget.id)
        }
        if (module.scripts && module.scripts.length) return jsGadgets.push(gadget.id)
        if (module.dependencies && module.dependencies.length) return jsGadgets.push(gadget.id)
        if (module.datas && module.datas.length) return jsGadgets.push(gadget.id)
        if (module.messages && module.messages.length) return jsGadgets.push(gadget.id)
        // Anything left now is a CSS-only Gagdet
        if (module.styles && module.styles.length) return cssGadgets.push(gadget.id)
      })

    return {cssGadgets, jsGadgets}
  }
```
This way we also have the JS gadgets ready once we start loading JS modules as well:
```ts
    const {cssGadgets, jsGadgets} = Gadgets.getAdditionalGadgets(articleDetail)
    cssGadgets.sort().map((gadgetId) => {
      moduleDependencies.styleDependenciesList.push(`ext.gadget.${gadgetId}`)
    })
    /*jsGadgets.map((gadgetId) => {
      moduleDependencies.jsDependenciesList.push(`ext.gadget.${gadgetId}`)
    })*/
```

src/renderers/action-parse.renderer.ts
benoit74: Fixed

src/Gadgets.ts
```diff
@@ -0,0 +1,110 @@
...
+    return Gadgets.instance
+  }
+
+  public async fetchGadgets() {
+    const gadgetsUrl = Downloader.apiUrlDirector.buildGadgetsUrl()
+    const gadgetsResponse = await Downloader.getJSON<GadgetQueryResult>(gadgetsUrl)
+    if (!gadgetsResponse.batchcomplete) {
+      throw new Error(`Error while fetching gadgets with ${gadgetsUrl}, batch seems to be incomplete, scraper does not expect/support pagination here`)
+    }
+    this.gadgets = gadgetsResponse.query.gadgets
```
Markus-Rost: As the query only received warnings with no data returned, the `query` key is already missing (see the JSON above). So your current check if gadgets is installed would still throw a `TypeError: Cannot read properties of undefined (reading 'gadgets')`

```suggestion
    this.gadgets = gadgetsResponse.query && gadgetsResponse.query.gadgets
```

(After #2385 we would always have the `query` key (because siteinfo always exists) and only the `gadgets` key missing.)

Markus-Rost (changes requested):

src/renderers/action-parse.renderer.ts
```diff
@@ -34,17 +31,28 @@ export class ActionParseRenderer extends Renderer {
...
+      .join('\n    ')
 
     const htmlTemplateString = htmlTemplateCode()
       .replace(/__ARTICLE_LANG_DIR__/g, articleLangDir)
       .replace('__ARTICLE_CANONICAL_LINK__', genCanonicalLink(config, MediaWiki.webUrl.href, articleId))
       .replace('__ARTICLE_CONFIGVARS_LIST__', articleConfigVarsList)
       .replace('__ARTICLE_JS_LIST__', articleJsList)
-      .replace('__ARTICLE_CSS_LIST__', articleCssList)
+      .replace('__ARTICLE_CSS_FROM_JS_LIST__', articleCssFromJsList)
+      .replace('__ARTICLE_CSS_ONLY_LIST__', articleCssOnlyList)
```
Markus-Rost: This naming is confusing `articleCssFromJsList` are also just CSS modules, not related to JS. While `articleCssOnlyList` is specifically CSS-only gadgets

res/templates/pageVector2022.html
```diff
@@ -1,13 +1,21 @@
-<!DOCTYPE html>
+<!doctype html>
 <html class="client-js" __ARTICLE_LANG_DIR__>
```
Markus-Rost:
```suggestion
<html class="client-nojs" __ARTICLE_LANG_DIR__>
```

Class should be `client-nojs` until #2310 gets fixed and we add a script element to change the class.

res/templates/pageVectorLegacy.html
```diff
@@ -1,12 +1,16 @@
-<!DOCTYPE html>
+<!doctype html>
 <html class="client-js" __ARTICLE_LANG_DIR__>
```
Markus-Rost:
```suggestion
<html class="client-nojs" __ARTICLE_LANG_DIR__>
```

Class should be `client-nojs` until #2310 gets fixed and we add a script element to change the class.

Markus-Rost (approved):
Looks good to me

PR #2378, opened 2025-06-22
Better RTL support

### LINKED ISSUE #1895: In the Hebrew Wikipedia, the labels and values in the infobox are aligned incorrectly
From kiwix-android created by [amire80](https://github.com/amire80): kiwix/kiwix-android#3321  

To test, open the article אנה קארנינה in the Hebrew Wikipedia ZIM. Look at the infobox: The labels and the values are aligned to the left. They are supposed to be aligned to the right. The rest of the article works well in terms of RTL alignment.

**Screenshots**

Screenshot_20230520_141521_Kiwix

**Environment**
- Version of Kiwix Android : 3.4.5 build 7230405
- Device : Samsung Galaxy S20 FE
- OS version : 13

### LINKED ISSUE #2367: Titles of articles for RTL languages displayed left-aligned instead of right-aligned
For example, in Arabic https://browse.library.kiwix.org/viewer#wikipedia_ar_all_nopic_2025-06

Image

And in Hebrew https://browse.library.kiwix.org/viewer#wikipedia_he_all_nopic_2025-06

Image

Markus-Rost:
Set the proper `lang=` and `dir=` attributes as well as the correct `mw-content-ltr/rtl` class in the HTML. 

Load modules using the actual default language of the wiki (this includes the parameter changes from #2375)

Fix #2367
Fix #1895

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -7,6 +7,7 @@ Unreleased:
 * FIX: Fix/simplify retrieval of siteinfo metadata (@benoit74 #2328)
 * FIX: Ensure illustration metadata is really 48x48 even if source image is not square (@benoit74 #2368)
 * FIX: Also rewrite inline <style> tags found in article HTML (@benoit74 #1807)
+* FIX: Better RTL support (@Markus-Rost #1865 #2367)
 
 1.15.1:
 * FIX: Do not fail when all articles retrieved have no revision (@benoit74 #2346)
--- a/res/templates/pageVector2022.html
+++ b/res/templates/pageVector2022.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html class='client-js'>
+<html class="client-js" __ARTICLE_LANG_DIR__>
   <head>
     <meta charset="UTF-8"/>
     <title></title>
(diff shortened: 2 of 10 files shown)


Markus-Rost:
I noticed that `abstract.renderer.ts` does a few class changes for images based on `textDir`. I didn't touch them yet as I wasn't sure about the exact purpose of the functions, but if they are just intended to change the default left/right floating, then the addition of the proper `mw-content-rtl` class will do this natively, making the function obsolete.

benoit74:
> I noticed that abstract.renderer.ts does a few class changes for images based on textDir.

These have been deactivated for ActionParse renderer (notice the `return` when `this.constructor.name === 'ActionParseRenderer'`. They indeed look like a legacy trying to handle something the wrong way (or the best way that was possible before ActionParse and everything else was available). No need to care for that.

benoit74:
I confirm this fixes the issue on https://he.wikipedia.org/wiki/%D7%90%D7%A0%D7%94_%D7%A7%D7%90%D7%A8%D7%A0%D7%99%D7%A0%D7%94

image

I will push force a commit just to fix the Changelog

benoit74:
CI is failing due to secrets not being passed to the workflow because we are running from an external repo (see https://github.com/kiwix/overview/issues/121). Nothing has been touched here, so I'm quite sure we did not broke anything. Anyway it will trigger something on `main` daily schedule run.

benoit74 (approved):

PR #2364, opened 2025-06-17
Make ActionParse renderer respect custom display titles

Markus-Rost:
Fix #2330

PR #2383, opened 2025-06-23
Support section redirects

### LINKED ISSUE #2382: Support section redirects
Wikis have a lot of redirects pointing to specific sections of the target page. However right now all our redirects only point to the top of the target article.

https://en.wikipedia.org/wiki/!= redirects to https://en.wikipedia.org/wiki/Equals_sign#Not_equal

However https://browse.library.kiwix.org/viewer#wikipedia_en_all_maxi_2024-01/A/!= only redirects to https://browse.library.kiwix.org/viewer#wikipedia_en_all_maxi_2024-01/A/Equals_sign instead of https://browse.library.kiwix.org/viewer#wikipedia_en_all_maxi_2024-01/A/Equals_sign#Not_equal

Markus-Rost:
Fix #2382

This should probably have some tests, but I don't know how to add them. Also libzim itself has no tests for redirects with url fragments in the target path, so I don't know if this actually works as intended.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -8,6 +8,7 @@ Unreleased:
 * FIX: Ensure illustration metadata is really 48x48 even if source image is not square (@benoit74 #2368)
 * FIX: Also rewrite inline <style> tags found in article HTML (@benoit74 #1807)
 * FIX: Better RTL support (@Markus-Rost #1865 #2367)
+* FIX: Support section redirects (@Markus-Rost #2382)
 
 1.15.1:
 * FIX: Do not fail when all articles retrieved have no revision (@benoit74 #2346)
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -460,7 +460,7 @@ class Downloader {
       // from new location to original location. Note that only ActionParse API gives proper redirects info.
       for (const redirect of redirects) {
         if (!(await RedisStore.articleDetailXId.exists(redirect.to)) && !(await RedisStore.redirectsXId.exists(redirect.to))) {
-          RedisStore.redirectsXId.set(redirect.to, { targetId: redirect.from, title: redirect.to })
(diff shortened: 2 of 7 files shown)


benoit74 (changes requested):
Redirects with fragments are not supported by libzim (yet?).

As mentioned in https://github.com/openzim/mwoffliner/issues/938#issuecomment-524567886, you must do it with normal HTML redirect, i.e. when you have a fragment, instead of creating a redirect you create a "normal" entry with this kind of content:

```
<html>
  <head>
    <title>Geography of Nairobi</title>
    <meta
      http-equiv="refresh"
      content="0;URL='./Nairobi#Geography'"
    />
  </head>
  <body></body>
</html>
```

benoit74:
Regarding tests, I can help or you might get inspired by https://github.com/openzim/mwoffliner/pull/2364 tests. Not sure it is "easy" to test both the selection by NS and by Id, maybe the test by NS can be added to the `bm.e2e.test.ts`, I believe this is the single test suite we have which fetch the whole NS.

benoit74:
@Markus-Rost do you intend to add more tests on this with the pointer I mentioned, or do you need help from me?

Markus-Rost:
Please help out with the tests as I'm not sure what you mean with the different selections. Also I won't get to implementing the HTML redirect before I'm on vacation next week, so feel free to implement that as well if you want to.

benoit74:
> Please help out with the tests 

I will

> I'm not sure what you mean with the different selections

I meant that there is in fact two scenario to test both code path `getArticlesByIds` and `getArticlesByNS`. First one is used when we build what we call a "selection" (of the Mediawiki) by passing an `--articleList` value. Second one is used when we ZIM the whole Mediawiki (`--articleList` is not used)

> Also I won't get to implementing the HTML redirect before I'm on vacation next week, so feel free to implement that as well if you want to.

I will too

benoit74:
Superseeded by https://github.com/openzim/mwoffliner/pull/2413

PR #2392, opened 2025-06-26
Use wiki default parser instead of forcing Parsoid with ActionParse

### LINKED ISSUE #2391: Don't force Parsoid with ActionParse
We are currently forcing the use of [Parsoid](https://www.mediawiki.org/wiki/Parsoid) in ActionParse using `parsoid=1`, but Parsoid is not the default parser for most wikis. While Parsoid is used on a lot of wikis through the VisualEditor, it is not yet used for normal read views. This means we are likely to miss the parser cache on most requests, adding unnecessary strain on the server. Some wikis might also be using extensions that are not compatible with the Parsoid parser yet. Parsoid is only planned to become the default MW parser by 1.47 (current release is 1.43). We should be using the default parser instead, there should be no issues with using the legacy parser for us because the `parsoid` API option only exists since MW 1.41 anyway. 

In #2351 it was mentioned that @cscott has requested we use the Parsoid parser and while I don't know the details of that discussion, I assume it mostly relates to Wikimedia wikis as Parsoid is being tested there for read views (see [Parsoid/Parser Unification](https://www.mediawiki.org/wiki/Parsoid/Parser_Unification#What_is_our_deployment_plan_/_strategy?)). To solve this issue we should instead set the `usearticle=1` option (added in MW 1.43) which will use the parser options as if it was a normal read view on the wiki. Specifically this will use Parsoid for any Wikimedia wikis with Parsoid HTML read views enabled, while using the legacy parser on all other wikis.

Markus-Rost:
Fix #2391


kelson42:
@Markus-Rost Thank you for your PR, I only start to understand the potential of it.

Can please try to rebase on main branch in place of merging main to your feature branch? This allows to have a cleaner git revision log.

I have invited you to join the team so you can directly create PR in our repo... and request reviewer yourself when your PR is ready.

Markus-Rost:
Thanks, I'll do that in the future.

Looks like I introduced some issues while trying to fix the tests, would be great if someone else could take a look at them.

test/unit/saveArticles.test.ts
```diff
@@ -72,10 +72,8 @@ describe('saveArticles', () => {
 
       const articleDoc = domino.createDocument(result[0].html)
 
-      const sections = Array.from(articleDoc.querySelectorAll('section'))
-      const leadSection = sections[0]
-      expect(sections.length).toEqual(1)
-      expect(leadSection.getAttribute('data-mw-section-id')).toEqual('0')
+      const sections = Array.from(articleDoc.querySelectorAll('.mw-heading'))
```
benoit74: Same remark, rename variable + add more tests

test/e2e/formatParams.test.ts
```diff
@@ -38,10 +38,9 @@ await testAllRenders('format-params-nodet', { ...parameters, format: 'nodet', ar
       const articleFromDump = await zimdump(`show --url BMW ${outFiles[0].outFile}`)
       const articleDoc = domino.createDocument(articleFromDump)
 
-      const sectionsElements = Array.from(articleDoc.querySelectorAll('section'))
+      const sectionsElements = Array.from(articleDoc.querySelectorAll('.mw-heading'))
```
benoit74: Rename `sectionsElements` to `headings`, I got very confused by this test.

Plus I think we should also assert that it contains some content + the infobox. Could be done for instance by checking there is multiple `<p>` and a table with `infobox` class.

src/util/saveArticles.ts
```diff
@@ -129,7 +129,8 @@ async function getAllArticlesToKeep(articleDetailXId: RKVS<ArticleDetail>, dump:
       try {
         const isMainPage = dump.isMainPage(articleId)
         const renderer = isMainPage ? mainPageRenderer : articlesRenderer
-        const articleUrl = isMainPage ? Downloader.getMainPageUrl(articleId) : Downloader.getArticleUrl(articleId)
+        const leadSectonId = dump.nodet ? config.filters.leadSectonId : ''
```
benoit74: Typo (everywhere): `leadSectonId` => `leadSectionId`

benoit74 (changes requested):
Some pretty small changes found.

But on a more higher level, I'm a bit puzzled by the fact that the kind of dump influence the query we have to make. 

While I do believe it makes sense to ask the parser to provide lead section rather than cleaning this on our own if this is too tricky, I don't get how it used to work before this PR. I would expect to have some changes linked to a cleanup of the previous way of doing the `nodet` version of an article.

Plus I'm also a bit sad that the dump type has an impact on the query we have to make, because it means we cannot factorize dumps "in general", we have some dumps which do need special queries. I had this idea to refactor the scraper to avoid doing one ZIM / dump at a time which is a significant waste of resources (if we build 3 dumps / ZIMs, we do 3 time the same query - or mostly the same now), but this PR tend to prove it is going to be a bit harder to implement.

Finally, on a even higher level, I don't get why we do not have anymore sections when querying en.wikipedia.org in tests. From your explanations in https://github.com/openzim/mwoffliner/issues/2391, I understood that `usearticle=1` would use the parser configured for read views on the given wiki, and that all Wikimedia wikis are using Parsoid for read views. I would hence expect to still have sections on en.wikipedia.org tests. The fact that we do not have any section seems to indicate that it is not Parsoid which is returning the response in this test. Did I missed something?

src/util/saveArticles.ts
Markus-Rost: oh wow, that's an old typo even

Markus-Rost:
> While I do believe it makes sense to ask the parser to provide lead section rather than cleaning this on our own if this is too tricky, I don't get how it used to work before this PR. I would expect to have some changes linked to a cleanup of the previous way of doing the `nodet` version of an article.

With Parsoid each page section has a `<section>` tag. Previously `nodet` just removed all those tags except for the first one. This doesn't work without Parsoid however, we would have to look for the first section header and remove everything afterwards. This would be a lot trickier, specially if the section heading is inside a table or other element. I didn't remove the only code for Parsoid because it is still being used by the other renderers.
https://github.com/openzim/mwoffliner/blob/600b9d34d419f09074d89ef0b650474885c344ab/src/renderers/abstract.renderer.ts#L648-L658


> Finally, on a even higher level, I don't get why we do not have anymore sections when querying en.wikipedia.org in tests. From your explanations in #2391, I understood that `usearticle=1` would use the parser configured for read views on the given wiki, and that all Wikimedia wikis are using Parsoid for read views. I would hence expect to still have sections on en.wikipedia.org tests. The fact that we do not have any section seems to indicate that it is not Parsoid which is returning the response in this test. Did I missed something?

en.wikipedia.org does not yet use Parsoid. Looking at the [deployment plan](https://www.mediawiki.org/wiki/Parsoid/Parser_Unification#What_is_our_deployment_plan_/_strategy?), everything besides Wikipedia (and Chinese wikis) are currently using Parsoid, with low-traffic Wikipedias being planned to switch next. This is in line with Wikimedias usual rollout strategy for experimental features, with them being tested on the smaller wikis first to catch any issues before enabling them on the large Wikipedias. en.wikipedia.org is generally the last Wikimedia wiki to enable new features.

src/util/saveArticles.ts
Markus-Rost: Fixed everywhere

test/e2e/formatParams.test.ts
Markus-Rost: Done

test/unit/saveArticles.test.ts
Markus-Rost: Done

benoit74:
OK, got it. Thank you for the details.

I believe we must then duplicate some tests to target both legacy parser and Parsoid when renderer is ActionParse. E.g. the `Check nodet article for en.wikipedia.org` should probably be in fact be ran on both en.wikipedia.org and on en.wikivoyage.org with ActionParse renderer. And we could assert from the returned HTML whether Parsoid was used or not, and check that all modifications are still done as expected.

And en.wikipedia.org is probably even probably not the wisest choice to test legacy parser, since it will probably be among the first remaining ones to migrate to Parsoid. An old unmaintained (but highly available) wiki would probably make even more sense, to be sure it will still work on legacy parser for quite a long time.

WDYT?

kelson42:
Maybe the best idea would be to have two different renderers... even if at the end they are the same to 99%?

Markus-Rost:
> Maybe the best idea would be to have two different renderers... even if at the end they are the same to 99%?

Wouldn't that defeat the point of automatically picking the best option for the wiki? We would need to manually set Parsoid or legacy for each wiki, keeping track of when they switch.

benoit74:
> Maybe the best idea would be to have two different renderers... even if at the end they are the same to 99%?

That's appealing because we would make this even more explicit, but at the same time a single wiki supports a single renderer ... which is kinda weird. For now we have multiple renderers and most wikimedia ones support all of them, and we use these wikis to test scraper behavior. Not sure the added work is really worth it. I don't know.

Markus-Rost:
Regarding tests, I agree that having tests for both parsoid and the legacy parser is a good idea. The current `nodet` test already works for both, as only ActionParse is using the legacy parser on en.wikipedia while the other renderers use Parsoid.

Having an old wiki to do tests against would also be great to ensure we actually support back to MW 1.27

benoit74:
@Markus-Rost do you intend to implement more tests, or should I review this "as-is" (and keep more tests as a nice-to-have we can live without?)

Markus-Rost:
Please review as-is.

We are already testing both parsers through the different renderers on en.wikipedia.org. I think that is good enough for now, though we will definitely need proper testing for both by the time en.wp switches to Parsoid. For that we will need to look for a wiki that will stay on the legacy parser first though.

benoit74 (approved):
LGTM

@kelson42 do you wanna review this as well or shall I merge this? Note that CI failing is only linked to the fact that the branch is from a fork.

kelson42:
@benoit74 Once rebased, good to merge for me

benoit74:
Superseeded by https://github.com/openzim/mwoffliner/pull/2405 (same content, different repo).

PR #2397, opened 2025-06-29
Use correct license and creator name in the footer

### LINKED ISSUE #2344: Revisit logic computing the `creator` used in all pages footer
Current logic to compute the `creator` used in all pages footer (Wikipedia, Wikisource, ...) is at https://github.com/openzim/mwoffliner/blob/b14021b2dae8fc2b0e32f14963baa727dffd6010/src/MediaWiki.ts#L412-L430

This is significantly wrong in multiple occasions.

First it is mostly based on hostname, and hence mostly only in English. This means that if the footer has been i18n and we are scraping a non-english mediawiki, we will still get the name in English in the middle of an i18n phrase. For instance, should `as` translations of the footer phrase be available

For instance, https://fr.wiktionary.org is called `Wiktionnaire` in French. But footer does not take this into account:

Image

Second, it does not take into account that mediawiki owner might have used a hostname which is significantly different that the nice-looking site name (for multitude of reasons).

I propose that we prefer to use the `sitename` metadata available in all wikis.

See e.g. https://fr.wiktionary.org/w/api.php?action=query&format=json&formatversion=2&meta=siteinfo&siprop=general|namespaces|statistics|variables|category|wikidesc

Image

The only downside I see is that if this `sitename` is already i18n but we miss this language in our scraper, then we will have a local name in the middle of an english phrase. Still probably more readable by most users than English-only phrase.

### LINKED ISSUE #2345: Licensing terms in page footer are at best imprecise, or even plainfully wrong
Currently the page footer of **all** mwoffliner ZIMs says (in English)

Image

(with a link to https://creativecommons.org/licenses/by-sa/4.0/)

But if we look at https://foundation.wikimedia.org/wiki/Policy:Terms_of_Use#7g, we will get:

> If the text content was imported from another source, it is possible that the content is licensed under a compatible CC BY-SA license but not GFDL (as described in "Importing text," above). In that case, you agree to comply with the compatible CC BY-SA license and do not have the option to relicense it under GFDL. To determine the license that applies to the content that you seek to reuse or redistribute, you should review the page footer, page history, and discussion page. 

In other words, for reuse (which we do), we might get CC BY-SA 4.0 license, or anything else compatible with it (public domain, other compatible license, ...) and there is absolutely no way to know this in an automated manner. So we have a significant chance that current phrase is at best imprecise, or even plainfully wrong.

And there are very higher chances it will be completely wrong for all non-Wikimedia mediawikis (this issue arises from https://github.com/openzim/zim-requests/issues/710#issuecomment-2967386302) where license might be "anything".

I don't know exactly how we should fix this, but I know I would definitely prefer to have no licensing information at all rather than a wrong one.

Markus-Rost:
Fix #2344
Fix #2345


translation/en.json
```diff
@@ -1,7 +1,7 @@
 {
-	"DISCLAIMER": "This article is issued from ${creator}. The text is licensed under ${license}. Additional terms may apply for the media files.",
+	"DISCLAIMER": "This article is issued from ${creator}. The text is available under ${license} unless otherwise noted. Additional terms may apply for the media files.",
 	"LAST_EDITED_ON": "Last edited on ${date}",
-	"LICENSE_NAME": "Creative Commons - Attribution - Sharealike",
+	"LICENSE_UNKNOWN": "an unknown license",
```
benoit74, line 4: You must:
- remove all `LICENSE_NAME` from all other locale JSON files (this is not needed anymore)
- add `LICENSE_UNKNOWN` to `qqq.json` with an explanation about the meaning of this key (`qqq.json` is a "magic" file used by TranslateWiki to load help text for translators

benoit74 (changes requested):
Thank you, see comment for small fix required.

translation/en.json
Markus-Rost: Of course I forgot about `qqq` 😅 Fixed

Issue #2396, opened 2025-06-27
Create a general skin template instead of supporting all skins


Markus-Rost:
We currently have a lot of open issues to support specific skins, some of them even just minor variants of an already support skin. But most skin specific elements are outside the content area while in the ZIM we only really care about the page content which barely changed with the different skins. So instead of creating templates for all possible skins, we should instead create a basic template skin which we use for all skins which aren't directly supported. As long as we set the correct skin body class (#2356), this should allow us to work with most skins without any issues.

Especially in relation to #2358 this would allow us to automatically use a wikis default skin to get the page content and place it in the basic template. This would mean most [skin support issues](https://github.com/openzim/mwoffliner/issues?q=is%3Aissue%20state%3Aopen%20label%3Askin) are no longer blocking the creation of ZIM files for wikis with custom skins.

benoit74:
Thanks a lot again @Markus-Rost for confirming what became to be a strong intuition on my side. Your experience with mediawiki and skin is once again very valuable here. 

So far, it looks like [vector 2022](https://github.com/openzim/mwoffliner/blob/main/res/templates/pageVector2022.html) template is more "complete" than [vector-legacy](https://github.com/openzim/mwoffliner/blob/main/res/templates/pageVectorLegacy.html) one. Should we use this as a default template or does it miss something (besides what will be fixed in #2356)?

Markus-Rost:
I think vector-legacy would be better as default template, it already has only the basic elements we always need. Vector-2022 has quite a few skin specific container elements we don't actually need for the basic default template.

kelson42:
@Markus-Rost @benoit74 I let you decide what and how would be the best default skin. For the rest, I understand the issue as we should make all skins inheritable and instanciable. Is that correct?

Markus-Rost:
> I understand the issue as we should make all skins inheritable and instanciable. Is that correct?

Actually no, we would just make all unsupported skins use the [vector-legacy template](https://github.com/openzim/mwoffliner/blob/main/res/templates/pageVectorLegacy.html) (or a dedicated default one) with everything else staying basically the same as right now. (Just need to have #2356 for the skin body class)

https://github.com/openzim/mwoffliner/blob/600b9d34d419f09074d89ef0b650474885c344ab/src/renderers/action-parse.renderer.ts#L24-L28

Instead of throwing an error, we just place the page content of unknown skins inside our vector-legacy template. This should work fine for most skins as the get the required CSS from the wiki anyway.

[...]

PR #2402, opened 2025-07-01
Fetch article <html> and <body> CSS class from headhtml response 

### LINKED ISSUE #2356: Correctly set classes on article pages body
Currently, the in-ZIM article page has "random" constant classes set on the body. See e.g. for vector-2022 skin: https://github.com/openzim/mwoffliner/blob/da6eaf0c4b88583d69970e979cc5fb1ccfbb1582/res/templates/pageVector2022.html#L10

These classes should be adapted to the context, at least taking into account
- which NS this article comes from
- which diretion (ltr/rtl) the wiki is in

I'm pretty sure there are other stuff to tweak, but we will have to investigate

benoit74:
Fix #2356 

Changes:
- ~~automatically compute all body CSS classes in ActionParse renderer~~
  - ~~for this I had to pass `articleDetails` to the template rendering function~~
- fetch article body CSS class from `headhtml` response of the action=parse request
- side-effect: I realized that we were passing `isMainPage` info around while it is a known property of current `Mediawiki` singleton ; nothing to do inside the dump (all dumps have the same main page) and no reason to pass this metadata around while it can be easily (and cheaply) recomputed with `Mediawiki` singleton

### DIFF
--- a/res/templates/pageVector2022.html
+++ b/res/templates/pageVector2022.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html class="client-nojs" __ARTICLE_LANG_DIR__>
+<html class="__ARTICLE_HTML_CSS_CLASS__" __ARTICLE_LANG_DIR__>
   <head>
     <meta charset="UTF-8" />
     <title></title>
@@ -12,10 +12,7 @@
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css" />
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector-2022.css" />
   </head>
-  <body
-    class="skin--responsive skin-vector skin-vector-search-vue mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject mw-editable skin-vector-2022 action-view uls-dialog-sticky-hide"
-    cz-shortcut-listen="true"
-  >
+  <body class="__ARTICLE_BODY_CSS_CLASS__">
     <div class="mw-page-container">
       <div class="mw-page-container-inner">
         <div class="mw-content-container">
--- a/res/templates/pageVectorLegacy.html
+++ b/res/templates/pageVectorLegacy.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html class="client-nojs" __ARTICLE_LANG_DIR__>
+<html class="__ARTICLE_HTML_CSS_CLASS__" __ARTICLE_LANG_DIR__>
   <head>
     <meta charset="UTF-8" />
     <title></title>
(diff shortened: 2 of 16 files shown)


Markus-Rost:
I feel a bit bad to only have noticed this after you already put in the work for the body classes PR, but it might be easier to just use the `<body>` (and `<html>`) element returned in `headhtml` of the action=parse request.

Butchering `headhtml` for the parts we need is a bit ugly, but we will have to do it anyway for #2310 and the classes returned there are more complete compared to what we can compute ourselves. We can only automatically compute the classes always set by MW core, but skins might set additional classes we need. We already have this issue with the vector skins in this PR and #2396 will only add more. 

For the `vector` skin the `skin-vector-legacy` class is missing while the `vector-2022` skin is missing the `skin-vector` class. Both vector variants use the `skin-vector` class in addition to the unique `skin-vector-legacy` and `vector-2022` classes.

benoit74:
Thanks a lot ! Always better to realize this during PR review than even later. And I love the idea of relying on something computed at Mediawiki level rather than our own fragile and crappy logic. Even if it means butchering some lines of HTML.

src/renderers/action-parse.renderer.ts
```diff
@@ -105,12 +107,14 @@ export class ActionParseRenderer extends Renderer {
       return redirect
     })
 
-    return { data: data.parse.text, moduleDependencies, redirects: normalizedRedirects, displayTitle: data.parse.displaytitle }
+    const bodyCssClass = extractBodyCssClass(data.parse.headhtml)
```
Markus-Rost, line 112: While you are at it, also get the classes for `<html>`. Vector-2022 has a lot of clientpref classes there.

res/templates/pageVector2022.html
```diff
@@ -12,10 +12,7 @@
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__footer.css" />
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector-2022.css" />
   </head>
-  <body
-    class="skin--responsive skin-vector skin-vector-search-vue mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject mw-editable skin-vector-2022 action-view uls-dialog-sticky-hide"
-    cz-shortcut-listen="true"
-  >
+  <body class="__ARTICLE_BODY_CSS_CLASS__" cz-shortcut-listen="true">
```
Markus-Rost: Why do we add some random `cz-shortcut-listen` to the body? It's not something set by the wiki and google suggests it's from the colorzilla browser extension.

res/templates/pageVectorLegacy.html
```diff
@@ -13,10 +13,7 @@
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH__vector.css" />
   </head>
 
-  <body
-    class="mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject mw-editable skin-vector action-view minerva--history-page-action-enabled skin-vector-legacy"
-    cz-shortcut-listen="true"
-  >
+  <body class="__ARTICLE_BODY_CSS_CLASS__" cz-shortcut-listen="true">
```
Markus-Rost: Same regarding `cz-shortcut-listen`

test/unit/util/articles.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(extractBodyCssClass(fakeHeamHtml).split(' ').sort()).toEqual(
```
Markus-Rost: Typo, `fakeHeamHtml` instead of `fakeHeadHtml`

Markus-Rost (changes requested):

res/templates/pageVector2022.html
benoit74: Because I copy-pasted it from my browser which has Colorzilla extension 🥺 

Will remove it.

Markus-Rost (approved):

benoit74:
Just fixed nasty bug trapped by CI, if everything ends-up green I will merge

Issue #2399, opened 2025-06-30
Add support for (chinese) variants


benoit74:
Some wikis like zh.wikipedia.org do support variants.

In https://github.com/openzim/mwoffliner/issues/840, we will do the necessary to at least be consistent in the variant we use in a ZIM.

Once done, we will have to decide how we want to support variants in mwoffliner ZIMs (especially for zh.wikipedia.org).

benoit74:
There is basically two approachs (not necessarily exclusive).

## Create one ZIM supporting all variants

First solution could be to build one ZIM (per selection, flavor, ...) supporting all Chinese variants.

From user perspective, this simplifies the choice of which ZIM to download since there is only one (per selection, flavor, ...). 

For users wanting to have multiple variants available, it is also quite convenient to have only one ZIM to open and, in general, it will also reduce storage requirements (typically pictures if you use the `maxi` will be stored only once).

We can implement it either with "on-the-fly" JS-based conversion (https://github.com/openzim/mwoffliner/issues/2443) or "in-the-ZIM", with all variants retrieved from the server and stored inside the ZIM.

None of these approach are very good from my perspective:
- the "on-the-fly" conversion is going to cause discrepancies between in-ZIM content and online content
- the "all variants in ZIM" approach is going to cause significant bigger ZIM

Another significant problem is that in both cases, search is going to be a nightmare because we will typically have to add all variants and might get many duplicate items in search results, on per variant, when search word is used by multiple variants. Readers and ZIM format have never been meant to contain multiple language in one ZIM. 

This is a known limitation we are already facing with TED ZIMs. We might decide to overcome this limitation at some point, but we are not yet at this stage. I've opened https://github.com/openzim/overview/issues/54 to track the topic.

## Create one ZIM per variant

In the medium / short term, I seems way more approachable to first focus on producing 6 ZIMs, one per variant. Or even 7 if we also build the "no-conversion" variant, which will contain a mix of variants but be 100% identical to wikitext. 

This approach definitely means additional storage if someone wants two or more variants. But this looks like the more approachable solution with current software and ZIM-format.

Markus-Rost:
> Another significant problem is that in both cases, search is going to be a nightmare because we will typically have to add all variants and might get many duplicate items in search results, on per variant, when search word is used by multiple variants. Readers and ZIM format have never been meant to contain multiple language in one ZIM.

[...]

PR #2406, opened 2025-07-04
Always use `zh-cn` variant on `zh` wikis

### LINKED ISSUE #840: Default scraping of zh.wiki content should be mainland traditional
There are several versions of Chinese cohabiting on zh.wikipedia (mainland simplified, HongKong traditional, Macao traditional, Malaysian simplified, Singapore simplified, Taiwan traditional). The wiki switches easily from on to the other as only characters are impacted (not grammar).

I'm not entirely sure how this works on our end but from many reports I get we're not serving people with mainland traditional (大陆简体) content, which is both easier to use and has a lot more readers. It would be nice, therefore, to scrape it correctly. I also remember that this very issue was raised a while back, but I could not find it anywhere.

Markus-Rost:
Fix #840

Hardcoded a variant to use for Chinese wikis


src/util/dump.ts
```diff
@@ -90,7 +90,7 @@ export async function downloadModule(module: string, type: 'js' | 'css') {
     )
   }
 
-  const langMw = MediaWiki.metaData?.langMw || 'en'
+  const langMw = MediaWiki.metaData?.langVar || MediaWiki.metaData?.langMw || 'en'
```
benoit74: Shouldn't we rename the const `lang` or `moduleLang` since it might contain is the variant or the language or the MW?

src/MediaWiki.ts
```diff
@@ -525,6 +525,7 @@ class MediaWiki {
 
       textDir: textDir as TextDirection,
       langMw,
+      langVar: langMw === 'zh' ? 'zh-cn' : null,
```
benoit74, line 528: Is it really ok to do this for all mediawikis in Chinese? I suspect it will be fine for all Wikimedia ones (but I'm not even sure) but I'm concerned about other Mediawikis. Is there a way to check this variant is available? Or should this variant not be available, action=parse will just return "default" result?

benoit74 (commented):
See inline comments + shouldn't we also do something for the titles and display titles? It looks like by default they are "mostly" in `zh-cn`, but I would prefer we are sure we kinda force proper variant.

src/MediaWiki.ts
Markus-Rost: The variant should generally be available on ZH wikis, and if not MW will follow the fallback chain: `['zh-cn', 'zh-hans', 'zh', 'zh-hant', 'en']`

Forcing Mainland China variant might not be ideal for all wikis, especially Taiwanese wikis. But to solve this better we will need #2399 with a new CLI option.

Markus-Rost:
> shouldn't we also do something for the titles and display titles? It looks like by default they are "mostly" in `zh-cn`, but I would prefer we are sure we kinda force proper variant.

The display titles will already be converted correctly into the variant by adding the parameter to the action=parse request, as we are using the display title provided by that endpoint.

For the titles themselves I don't know how we could solve that. I'm only aware of an API option to convert variant names in `--articleList` into the actual page title (unless there is already a redirect). So the ZIM titles should stay the actual page title, whatever variant it is.

src/MediaWiki.ts
benoit74: Probably kinda rare anyway, looks like an acceptable compromise.

benoit74 (approved):

PR #2413, opened 2025-07-07
Support section redirects

### LINKED ISSUE #938: Handle properly articles list with redirect with local anchors

### LINKED ISSUE #2382: Support section redirects

benoit74:
Superseeds #2383 by using a branch from repo instead of fork + rebased + new commit to handle redirect with fragments

Fix #2382
Fixes #938 

This should probably have some tests, but I don't know how to add them. Also libzim itself has no tests for redirects with url fragments in the target path, so I don't know if this actually works as intended.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,4 +1,5 @@
 Unreleased:
+* FIX: Support section redirects (@Markus-Rost @benoit74 #2382)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
--- a/res/templates/html_redirect.html
+++ b/res/templates/html_redirect.html
@@ -1,0 +1,7 @@
+<html>
+  <head>
+    <title>__TITLE__</title>
+    <meta http-equiv="refresh" content="0;URL='__RELATIVE_FILE_PATH____TARGET__'" />
+  </head>
+  <body></body>
+</html>
--- a/res/templates/html_redirect.html
+++ b/res/templates/html_redirect.html
@@ -3,5 +3,7 @@
     <title>__TITLE__</title>
     <meta http-equiv="refresh" content="0;URL='__RELATIVE_FILE_PATH____TARGET__'" />
   </head>
-  <body></body>
+  <body>
+    <a href="__RELATIVE_FILE_PATH____TARGET__">__TITLE__</a>
+  </body>
 </html>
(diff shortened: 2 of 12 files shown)


kelson42:
@benoit74 We can simulate the situation in wiki.kiwix.org and then check what is written in ZIM file?

benoit74:
> @benoit74 We can simulate the situation in wiki.kiwix.org and then check what is written in ZIM file?

Why? Did you checked PR content? I've already added e2e tests based on en.wikipedia.org and bm.wikipedia.org, both checking what is written inside the ZIM. First (WPEN) checks behavior when an article list is passed, second (WPBM) checks behavior when we scrape the whole wiki. And in both cases I check a redirect with fragment and a redirect without fragment. Isn't this enough?

Markus-Rost:
Considering meta refresh is deprecated by W3C, the template should probably have a normal link to the target as fallback.

benoit74:
> Considering meta refresh is deprecated by W3C, the template should probably have a normal link to the target as fallback.

Really? Indeed a link in the body is a significant added value and small price to pay. But this also shows we should probably implement something in the libzim / readers.

benoit74:
For the record, I've opened https://github.com/openzim/libzim/issues/1001 for longer term solution

benoit74:
I just added the link in body

PR #2417, opened 2025-07-08
Keep all HTML tags from parser, since they might be used by CSS (especially for backlinks)

### LINKED ISSUE #881: Back links in references are missing
But they are existing online in mobile view! See for example here:
https://en.m.wikipedia.org/wiki/Kiwix#cite_ref-5

It allows to comme back to the text which has been referenced. This is important.

Why we don't have them?

benoit74:
Fix #881

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -42,6 +42,11 @@ const config = {
     /* Lead section to display */
     leadSectionId: '0',
 
+    /* List of style to be removed. 'onlyoffline' is a *hack* to
+    /* display WPEN medical articles, see for example
+    /* [[Carcinoid_syndrome]] */
+    cssClassCallsBlackList: ['plainlinks', 'onlyoffline'],
+
     /* All nodes with one of these ids will be removed */
     idBlackList: ['purgelink'],
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -42,11 +42,6 @@ const config = {
     /* Lead section to display */
     leadSectionId: '0',
 
-    /* List of style to be removed. 'onlyoffline' is a *hack* to
-    /* display WPEN medical articles, see for example
-    /* [[Carcinoid_syndrome]] */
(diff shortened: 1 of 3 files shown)


src/renderers/abstract.renderer.ts
```diff
@@ -682,39 +682,6 @@ export abstract class Renderer {
...
 
-  private clearNodes(parsoidDoc: DominoElement, filtersConfig: any) {
-    const allNodes: DominoElement[] = Array.from(parsoidDoc.getElementsByTagName('*'))
-    for (const node of allNodes) {
-      node.removeAttribute('data-parsoid')
-      if (this.constructor.name !== 'ActionParseRenderer') {
-        node.removeAttribute('typeof')
-      }
-      node.removeAttribute('about')
-      node.removeAttribute('data-mw')
```
Markus-Rost, line 690: At least `data-mw` should still be removed per https://github.com/openzim/mwoffliner/issues/881#issuecomment-3048963981

It contains huge amounts of bloat we don't need and keeping it would increase the size of each article significantly.

src/config.ts
```diff
@@ -42,11 +42,6 @@ const config = {
     /* Lead section to display */
     leadSectionId: '0',
 
-    /* List of style to be removed. 'onlyoffline' is a *hack* to
-    /* display WPEN medical articles, see for example
-    /* [[Carcinoid_syndrome]] */
-    cssClassCallsBlackList: ['plainlinks', 'onlyoffline'],
```
Markus-Rost: `onlyoffline` allows the addition of elements only visible in ZIMs. This might be useful to provide fallback for interactive elements when combined with `noprint` usage. We should keep removing this class to allow this functionality to keep working.

benoit74:
Fixed as suggested. I've kept logic removing `data-file-*` as well on `` since it seems safe to remove as well ... we'll see.

Markus-Rost (approved):

PR #2419, opened 2025-07-08
 Restore keepEmptyParagraphs correct behavior

### LINKED ISSUE #1866: Deprecate tests for details and sections
Current tests expect that **sections** and **details** will be removed if they have only **summary** tag inside without empty paragraph inside. The option `keepEmptyParagraphs` is responsible for removing these elements. The source code of it is:
```
/* Remove empty paragraphs */
  if (!dump.opts.keepEmptyParagraphs) {
    // Mobile view === details
    // Desktop view === section
    const sections: DominoElement[] = Array.from(parsoidDoc.querySelectorAll('details, section'))
    for (const section of sections) {
      if (
        section.children.length ===
        Array.from(section.children).filter((child: DominoElement) => {
          return child.matches('summary')
        }).length
      ) {
        DU.deleteNode(section)
      }
    }
  }
```
Consider this HTML is an example:
```
    <section>
        <summary>Section Summary</summary>
    </section>
    <details>
        <summary>Details Summary</summary>
    </details>
    <section>
        <summary>Section Summary</summary>
        <p>Some content</p>
    </section>
    <details>
        <summary>Details Summary</summary>
        <p>Some content</p>
    </details>
```

After transformation, the first **section** and **details** elements will be removed, as they only contain **summary** children. So the final output will look like this:
```
    <section>
        <summary>Section Summary</summary>
        <p>Some content</p>
    </section>
    <details>
        <summary>Details Summary</summary>
        <p>Some content</p>
    </details>
```
Current Parsoid version doesn't have the support of **details** tags (see https://phabricator.wikimedia.org/T31118) so `keepEmptyParagraphs` should be refactored with related unit tests since we expect to use **page/html** and **page/mobile-html** for desktop and mobile view respectively.

### LINKED ISSUE #1915: `<summary>` and `<details>` DOM nodes are no longer supported
Before `mobile-section` deprecation mwoffliner used related JSON response to generate collapsible sections using Swig templates (see [res/templates/section_wrapper.html](https://github.com/openzim/mwoffliner/blob/main/res/templates/section_wrapper.html) file). Since `page/html` and `page/mobile-html` are responsible for outputting HTML content, no collapse sections are available. The symptom of this issue is mentioned in https://github.com/openzim/mwoffliner/issues/1860 and https://github.com/openzim/mwoffliner/issues/1866

There are multiple ways to bring these sections to work:
1. Apply `page/html` output and related transformations from [mobileapps](https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/routes/page/mobile-sections.js) in mwoffliner.
2. Create our own collapsible sections but on top of `page/mobile-html`

benoit74:
Fix #1866
Fix #1915 

Nota: all renderers are fixed, it was pretty easy/obvious to fix them all, see new tests

### DIFF
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -768,17 +768,17 @@ export abstract class Renderer {
     })
 
     /* Remove empty paragraphs */
-    // TODO: Refactor this option to work with page/html and page/mobile-html output. See issues/1866
     if (!dump.opts.keepEmptyParagraphs) {
-      // Mobile view === details
-      // Desktop view === section
-      const sections: DominoElement[] = Array.from(parsoidDoc.querySelectorAll('details, section'))
+      const sections: DominoElement[] = Array.from(parsoidDoc.querySelectorAll('section'))
       for (const section of sections) {
         if (
-          section.children.length ===
           Array.from(section.children).filter((child: DominoElement) => {
-            return child.matches('summary')
-          }).length
+            // Check if there is more than the expected children we have in all sections
+            // RestApi : h2 tag
+            // WikimediaDesktop, VisualEditor, Action Parse: mw-heading class
+            // WikimediaMobile: pcs-edit-section-header class
+            return !child.matches('h2') && !child.classList.contains('mw-heading') && !child.classList.contains('pcs-edit-section-header')
+          }).length === 0
         ) {
           DU.deleteNode(section)
         }
--- a/Changelog
+++ b/Changelog
@@ -7,6 +7,7 @@ Unreleased:
 * NEW: Add progress logs while writing redirects (@benoit74 #2381)
 * FIX: Keep some HTML tags from parser, since they might be used by CSS (especially for backlinks) (@benoit74 #881)
 * NEW: Automatically detect and use mediawiki default skin (@benoit74 @Markus-Rost #2213, #2358, #2396)
+* FIX: Restore --keepEmptyParagraphs correct behavior on all renderers, except ActionParse when wiki is not yet using Parsoid parser (@benoit74 #1866 #1915)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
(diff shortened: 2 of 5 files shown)


src/renderers/abstract.renderer.ts
```diff
@@ -772,17 +772,17 @@ export abstract class Renderer {
     })
 
     /* Remove empty paragraphs */
-    // TODO: Refactor this option to work with page/html and page/mobile-html output. See issues/1866
     if (!dump.opts.keepEmptyParagraphs) {
-      // Mobile view === details
-      // Desktop view === section
-      const sections: DominoElement[] = Array.from(parsoidDoc.querySelectorAll('details, section'))
+      const sections: DominoElement[] = Array.from(parsoidDoc.querySelectorAll('section'))
```
Markus-Rost, line 772: This won't do anything with ActionParse for wikis that don't use Parsoid.

benoit74: Yes, I let @kelson42 confirm but from my understanding this is "accepted" in the sense that this feature is anyway mostly only for wikimedia wikis which are using Parsoid. And we have no real alternative for other wikis (besides very complex and fragile logic) since all HTML is "at the same level" without proper sections.

kelson42: I confirm, but would probably good to somehow inform the user. Can we generate an error if this happens and invite the user to use `--keepEmptyParagraphs`?

benoit74: @kelson42 your point makes no sense ; for wikis which don't use Parsoid, the code will not trigger so we will always keep empty paragraphs. Passing the `--keepEmptyParagraphs` will not change anything. And it is not really possible to detect that we've kept empty paragraphs since we don't know these are empty paragraphs otherwise we would have deleted them.

kelson42: @benoit74 your understand me properly, but you miss IMHO the point to do so. This is not about the feature itself, it is about forcing the awareness of the user that MWoffliner can not behave in a noraml manner and can only work with this option. That way things are coherent: mwoffliner will deliver like specified.

benoit74: I would not bother users about having to set a flag to get the same behavior than if the flag is not set. Of course, users won't know. But they don't know about this feature anyway in the first place.

kelson42: We have to  bother users if the software does not deliver like expected. We don't do things behind the scene without clearly inform users... and just a warning is IMO not enough. Anyway, not a deal breaker for me. At least put a clear warning.

benoit74: > At least put a clear warning

I have no clue how to do this, especially reliably. I've moved this topic to https://github.com/openzim/mwoffliner/issues/2431 so that we can at least close this PR. 

@kelson42 please review

Issue #2422, opened 2025-07-10
Automatically guess / retrieve mediawiki paths


benoit74:
Currently, one needs to configure `mwUrl` and adjust `mwWikiPath`, `mwIndexPhpPath`, `mwActionApiPath`, `mwRestApiPath`, and `mwModulePath`.

We should find a way to guess or retrieve these paths automatically, there is just too much room for error here, especially `mwWikiPath` and `mwIndexPhpPath` which could be wrong without the scrapper really noticing until many links get broken.

To be investigated, looks like these are not exposed in an API. Some can be found on `Special:Version` page but HTML needs to be parsed, and some are missing on this page, and not all values are easy to adjust correctly. For instance on https://minecraft.wiki/w/Special:Version, we have `articlePath` set to `/w/$1` which means that `mwWikiPath` is `/w/`, and `index.php` set to `/` which means that `mwIndexPhpPath` is `/index.php` ... not really straightforward conversions. But maybe we don't have anything better. 

Note this is a enhancement related to https://github.com/openzim/mwoffliner/issues/2421 (which should be implemented first since this is a kind of bug)

Markus-Rost:
General siteinfo in the API includes this information:

* `mwWikiPath` is `articlepath` in the API.
  * Note that MW default for this value is `/w/index.php/$1` or `/w/index.php?title=$1` depending on config of the web server.
* `mwIndexPhpPath` is `script` in the API
  * MCW has set a custom path for index.php to the root
* `mwActionApiPath`, `mwRestApiPath`, `mwModulePath` (and `mwIndexPhpPath`) are build using `scriptpath` in the API. (Scriptpath + `/api.php`, `/rest.php`, `/load.php`, `/index.php`)
  * Each script can be set to specific paths as well (like MCW with index.php), though the one build with scriptpath usually redirects or works as well (not always, so the individual settings still need to exist)

https://zh.wikipedia.org/w/api.php?action=query&format=json&meta=siteinfo&siprop=general&formatversion=2
```json
{
    "articlepath": "/wiki/$1",
    "scriptpath": "/w",
    "script": "/w/index.php",
    "variantarticlepath": "/$2/$1",
    "server": "//zh.wikipedia.org",
    "servername": "zh.wikipedia.org",
}
```

https://minecraft.wiki/api.php?action=query&format=json&meta=siteinfo&siprop=general&formatversion=2
```json
{
    "articlepath": "/w/$1",
    "scriptpath": "",
[...]

PR #2428, opened 2025-07-10
Check validity of mwIndexPhpPath and mwWikiPath at startup

### LINKED ISSUE #2421: Automatically check `mwWikiPath` and `mwIndexPhpPath`
Currently, the scraper checks that `mwActionApiPath`, `mwRestApiPath`, and `mwModulePath` are correct. But it does not check `mwWikiPath` and `mwIndexPhpPath`. It should check them as well and fail scrape if they are not correct, since it means quite a significant share of URL rewriting will be broken.

benoit74:
Fix #2421

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -276,6 +276,20 @@ class MediaWiki {
     return this.#hasModuleApi
   }
 
+  public async checkWikiPath(): Promise<boolean> {
+    const checkUrl = this.urlDirector.buildURL(`${this.#wikiPath}${this.apiCheckArticleId}`).href
+    const result = await checkApiAvailability(checkUrl)
+    logger.log('Checked WikiPath at', checkUrl, '-- result is: ', result)
+    return result
+  }
+
+  public async checkIndexPhpPath(): Promise<boolean> {
+    const checkUrl = this.urlDirector.buildURL(this.#indexPhpPath, { title: this.apiCheckArticleId }).href
+    const result = await checkApiAvailability(checkUrl)
+    logger.log('Checked IndexPhpPath at', checkUrl, '-- result is: ', result)
+    return result
+  }
+
   private setWikimediaDesktopApiUrl() {
     this.wikimediaDesktopApiUrl = this.urlDirector.buildWikimediaDesktopApiUrl()
   }
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -13,7 +13,7 @@ import WikimediaMobileURLDirector from './util/builders/url/mobile.director.js'
 import VisualEditorURLDirector from './util/builders/url/visual-editor.director.js'
 import RestApiURLDirector from './util/builders/url/rest-api.director.js'
 import ActionParseURLDirector from './util/builders/url/action-parse.director.js'
-import { checkApiAvailability } from './util/mw-api.js'
+import { checkApiAvailability, checkArticleUrl } from './util/mw-api.js'
 import { BLACKLISTED_NS } from './util/const.js'
 
 export interface QueryOpts {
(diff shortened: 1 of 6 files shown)


src/MediaWiki.ts
```diff
@@ -276,6 +276,20 @@ class MediaWiki {
...
   }
 
+  public async checkWikiPath(): Promise<boolean> {
+    const checkUrl = this.urlDirector.buildURL(`${this.#wikiPath}${this.apiCheckArticleId}`).href
+    const result = await checkApiAvailability(checkUrl)
+    logger.log('Checked WikiPath at', checkUrl, '-- result is: ', result)
+    return result
+  }
+
+  public async checkIndexPhpPath(): Promise<boolean> {
```
Markus-Rost: Many wikis redirect such index.php requests to their wiki path, that would fail this check. (Any tests on fandom.com wikis should include this behavior)

src/MediaWiki.ts
```diff
@@ -276,6 +276,20 @@ class MediaWiki {
     return this.#hasModuleApi
   }
 
+  public async checkWikiPath(): Promise<boolean> {
```
Markus-Rost: The main page name used for `apiCheckArticleId` might be a redirect on some wikis, failing this check. See Special:MyLanguage/MediaWiki as main page on mediawiki.org

Markus-Rost (changes requested):
Both checks might fail on a multitude of wikis because they require HTTP 200 and no redirect.

src/MediaWiki.ts
benoit74: Good point. 

However, redirects are not handled at HTTP level, see below with `Geography_of_Nairobi` which is a redirect:

```
> curl -L -I https://en.wikipedia.org/wiki/Geography_of_Nairobi
HTTP/2 200 
...
```

I will however add a test case to confirm this does not change in the future

src/MediaWiki.ts
benoit74: Indeed, I will add `redirect: no` to fix this.

src/MediaWiki.ts
benoit74: Edit: "it depends"

```
> curl -L -I https://pokemon.fandom.com/wiki/National_Dex
HTTP/2 301
...
location: https://pokemon.fandom.com/wiki/List_of_Pok%C3%A9mon
...

HTTP/2 200
...
```

benoit74:
Plus we do not check sufficiently the content of the response, many wikis might reply 200 but still have only garbage inside the page, defeating the check. I will enhance the checks.

test/unit/mwApiCapabilities.test.ts
```diff
@@ -41,30 +58,48 @@ describe('Checking Mediawiki capabilities', () => {
...
+    expect(await MediaWiki.checkWikiPath()).toBe(true)
+    expect(await MediaWiki.checkIndexPhpPath('MCE')).toBe(true) // check with a redirected article
+    expect(await MediaWiki.checkWikiPath('MCE')).toBe(true) // check with a redirected article
+  })
+
+  test('test capabilities of minecraft.wiki with incorrect path config', async () => {
+    MediaWiki.base = 'https://minecraft.wiki'
+    Downloader.init = { uaString: `${config.userAgent} (contact@kiwix.org)`, speed: 1, reqTimeout: 1000 * 60, webp: true, optimisationCacheUrl: '' }
+    expect(await MediaWiki.checkIndexPhpPath()).toBe(false)
+    expect(await MediaWiki.checkWikiPath()).toBe(true) // automatically redirect by minecraft.wiki
```
Markus-Rost, line 86: This means we actually failed to confirm the validity of mwWikiPath. Our check says `/wiki/` is the correct path for minecraft.wiki but all wiki links use `/w/` so we still end up with a ZIM where all links got removed.

benoit74: Indeed. Looks like an acceptable limitation of the check. Isn't this an acceptable limitation, or do you have suggestion to make it any better? I don't know how to make a distinction between a "normal" redirect and an "abnormal" redirect.

Markus-Rost: The only redirects we want to allow are page redirects, so we should confirm that our mwWikiPath is still part of the url after the redirect.

Though as noted in https://github.com/openzim/mwoffliner/issues/2421#issuecomment-3058687417 I'm not sure these checks are actually needed as #2422 would remove both of these CLI arguments.

benoit74: I'm pretty sure we will have wiki in the wild which will redirect from say `/wiki/` to `/index.php?title=...`, so we must accept redirects on wiki path as well.

I don't know how to have a bullet-proof solution, and a good solution is anyway a mix of https://github.com/openzim/mwoffliner/issues/2421 (this PR) and https://github.com/openzim/mwoffliner/issues/2422

benoit74: > Though as noted in https://github.com/openzim/mwoffliner/issues/2421#issuecomment-3058687417 I'm not sure these checks are actually needed as https://github.com/openzim/mwoffliner/issues/2422 would remove both of these CLI arguments.

This is indeed a good point, I started this PR before you mentioned in #2422 that we have a reliable way to retrieve these values and drop the CLI arguments. Maybe we should just focus on #2422 and drop this PR. Tbh I'm a bit concerned about the fact that the logic of #2422 would fail under some condition and we would miss the CLI arguments.

kelson42:
To me, the status of this PR is a bit unclear, what should come next?

benoit74:
A decision on what is the status of this PR 🤣 

Do we really need this PR if we implement #2422? Since #2422 seems easier / more reliable than expected, plus this PR is less reliable than expected?

kelson42:
#2422 is for 2.0.0 and I doubt this is that easy to fully the added value of such a move.

In addition, I have remarks about the two weaknesses of this PR listed above:
* I don't think this is a good idea to support redirects in paths cmd line arguments in this PR. It has not been the case so far and would make things more complicated. If we want to do that I recommend a dedicated issue.
* Your point of checking response body as well, is a good point. But if we want to do that "in general" I would recommend to open a dedicated issue as well.

All in one, many reasons to mwerge this right now.

benoit74:
Let's speak about it live, I don't get your points and I believe your probably didn't got mine (e.g. this PR is already handling redirects). Pretty sure this PR is not ready to be merged.

benoit74:
Issue abandoned.

PR #2433, opened 2025-07-15
Automatically use mediawiki default skin with fallback template

### LINKED ISSUE #2213: Allow scraper user to choose skin to use

### LINKED ISSUE #2358: Automatically detect / use default skin
When we request main page of a Mediawiki, we get sufficient info about which skin is used by default with CSS classes of the `body` tag.

For https://minecraft.wiki, we have `skin-vector-legacy` and `skin-vector`:

```
<body class="skin-vector-legacy mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject page-Minecraft_Wiki rootpage-Minecraft_Wiki skin-vector action-view wgl-lightmode wgl-theme-light" cz-shortcut-listen="true">
```

https://minecraft.wiki/api.php?action=parse&format=json&page=MediaWiki:Sidebar&parsoid=1&prop=templates|modules|jsconfigvars|headhtml|text&useskin=vector-legacy is not working but https://minecraft.wiki/api.php?action=parse&format=json&page=MediaWiki:Sidebar&parsoid=1&prop=templates|modules|jsconfigvars|headhtml|text&useskin=vector is OK => we know that default skin is `vector`

For https://wiki.selfhtml.org/, we have `skin-selfhtml`:

```
<body class="mediawiki ltr sitedir-ltr capitalize-all-nouns ns-0 ns-subject page-SELFHTML rootpage-SELFHTML skin-selfhtml action-view" cz-shortcut-listen="true">
```

=> we know that skin is selfhtml (but it is hard to confirm because this old mediawiki version does not support the `useskin` parameter in `action=parse`)

I'm sure we can leverage this information to guess which skin the wiki uses by default.

### LINKED ISSUE #2396: Create a general skin template instead of supporting all skins

Markus-Rost:
Fix #2213
Fix #2358
Fix #2396

Makes most https://github.com/openzim/mwoffliner/labels/skin issues obsolete

Adds a fallback template based on vector legacy if the specific skin has no template of its own.

Adds a `forceSkin` CLI argument to use a skin supported by the wiki other than the default skin.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -6,6 +6,7 @@ Unreleased:
 * CHANGED: Use subtitle from ActionParse (@Markus-Rost #179)
 * NEW: Add progress logs while writing redirects (@benoit74 #2381)
 * FIX: Keep some HTML tags from parser, since they might be used by CSS (especially for backlinks) (@benoit74 #881)
+* NEW: Automatically detect and use mediawiki default skin (@benoit74 #2358)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
--- a/Changelog
+++ b/Changelog
@@ -6,7 +6,7 @@ Unreleased:
 * CHANGED: Use subtitle from ActionParse (@Markus-Rost #179)
 * NEW: Add progress logs while writing redirects (@benoit74 #2381)
 * FIX: Keep some HTML tags from parser, since they might be used by CSS (especially for backlinks) (@benoit74 #881)
-* NEW: Automatically detect and use mediawiki default skin (@benoit74 #2358)
+* NEW: Automatically detect and use mediawiki default skin (@benoit74 @Markus-Rost #2213, #2358, #2396)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
--- a/src/renderers/action-parse.renderer.ts
+++ b/src/renderers/action-parse.renderer.ts
@@ -1,10 +1,11 @@
 import domino from 'domino'
 import { DownloadOpts, DownloadRes, Renderer, RenderOptsModules } from './abstract.renderer.js'
 import { RenderOpts, RenderOutput } from './abstract.renderer.js'
+import * as logger from '../Logger.js'
 import { config } from '../config.js'
 import { genCanonicalLink, genHeaderScript, genHeaderCSSLink, getStaticFiles, getRelativeFilePath } from '../util/misc.js'
 import MediaWiki from '../MediaWiki.js'
-import { htmlVectorLegacyTemplateCode, htmlVector2022TemplateCode } from '../Templates.js'
+import { htmlVectorLegacyTemplateCode, htmlVector2022TemplateCode, htmlFallbackTemplateCode } from '../Templates.js'
 import Downloader, { DownloadError } from '../Downloader.js'
 import Gadgets from '../Gadgets.js'
 import { extractBodyCssClass, extractHtmlCssClass } from '../util/articles.js'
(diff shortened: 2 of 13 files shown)


kelson42:
We have problems here with the CI. Should be fixed IMHO before review.

Markus-Rost:
They are unrelated to the changes here. I've re-run the jobs though.

benoit74:
Note this fixes #2213 as well, I've updated first comment

Changelog
```diff
@@ -6,6 +6,7 @@ Unreleased:
 * CHANGED: Use subtitle from ActionParse (@Markus-Rost #179)
 * NEW: Add progress logs while writing redirects (@benoit74 #2381)
 * FIX: Keep some HTML tags from parser, since they might be used by CSS (especially for backlinks) (@benoit74 #881)
+* NEW: Automatically detect and use mediawiki default skin (@benoit74 #2358)
```
benoit74: Please add entries for the two other issues as well

benoit74 (changes requested):
Thank you.

We should:
- add an e2e test case scraping one wiki we know we do not have the skin for (and mention it in the relevant issue so that once we implement this skin we move the e2e test to another wiki ; no need to scrape the whole wiki, only the main page is probably enough
- this test will show that the computation of `this.staticFilesList = getStaticFiles(...` is wrong when using the fallback skin: we do not have a `${skin}.css` resource needed
- add a warning message when we use the fallback skin saying that it is expected the rendering will be a bit "off" ; for instance I did a test on pokemon.fandom.org and it is far from getting a good looking result ; every text is here, readable, but the look is "barebone"

Markus-Rost:
> Thank you.
> 
> We should:
> 
> * add an e2e test case scraping one wiki we know we do not have the skin for (and mention it in the relevant issue so that once we implement this skin we move the e2e test to another wiki ; no need to scrape the whole wiki, only the main page is probably enough

I've added a combined test for the `forceSkin` option and the fallback skin by scraping a Wikipedia page with an unsupported skin (`monobook` in this case)

> * this test will show that the computation of `this.staticFilesList = getStaticFiles(...` is wrong when using the fallback skin: we do not have a `${skin}.css` resource needed
> * add a warning message when we use the fallback skin saying that it is expected the rendering will be a bit "off" ; for instance I did a test on pokemon.fandom.org and it is far from getting a good looking result ; every text is here, readable, but the look is "barebone"

fixed that and added the warning as well. Let me know if the warning text is ok or should be more detailed.

src/renderers/action-parse.renderer.ts
```diff
@@ -15,18 +16,20 @@ export class ActionParseRenderer extends Renderer {
   constructor() {
     super()
     if (this.staticFilesList.length === 0) {
-      this.staticFilesList = getStaticFiles(config.output.jsResourcesCommon, config.output.cssResourcesCommon.concat(MediaWiki.skin)).concat('external-link.svg')
+      let cssResourcesCommon = config.output.cssResourcesCommon
+      if (['vector', 'vector-2022'].includes(MediaWiki.skin)) {
+        cssResourcesCommon = cssResourcesCommon.concat(MediaWiki.skin)
+      } else {
+        logger.warn(`Unsupported skin ${MediaWiki.skin}, using fallback template to display pages.`)
```
benoit74: While convenient, I'm worried we might get into a situation where we are not aligned anymore between which skins have a custom CSS and which ones use the fallback template. 

I prefer we move this warning in `templateDeskopArticle(...)`, where we can just check there that `htmlTemplateCode` is `htmlFallbackTemplateCode`. It would be more accurate and definitely more future-proof.

Markus-Rost: I was considering that, but having the warning inside `templateDeskopArticle()` means it gets logged for every single page inside the ZIM. That seems a bit excessive.

Markus-Rost: But I could probably move the definition of `htmlTemplateCode` into the constructor. It's the same for all pages anyway.

benoit74 (approved):
LGTM, thank you!

evrial:
I think the issue remains
--forceSkin=vector has no effect
```
docker run --volume=$(pwd)/out:/out -ti ghcr.io/openzim/mwoffliner mwoffliner --verbose --mwUrl=https://www.sigidwiki.com --adminEmail=foo@bar.net --mwActionApiPath=api.php
[log] [2025-08-04T19:16:14.280Z] Starting mwoffliner v1.16.1-dev0...
[info] [2025-08-04T19:16:14.284Z] Using custom flavour: no
[log] [2025-08-04T19:16:14.870Z] closing sanitize redis DB
[log] [2025-08-04T19:16:14.871Z] Getting site info...
[log] [2025-08-04T19:16:14.871Z] Getting sub-title...
[info] [2025-08-04T19:16:14.872Z] Getting JSON from [https://www.sigidwiki.com/api.php?action=query&meta=siteinfo&siprop=general%7Cskins%7Crightsinfo&format=json&formatversion=2]
[info] [2025-08-04T19:16:14.874Z] Downloading [https://www.sigidwiki.com/wiki/]
[log] [2025-08-04T19:16:15.042Z] Text direction is [ltr]
[error] [2025-08-04T19:16:15.043Z] FATAL - Failed to get MediaWiki Metadata
[error] [2025-08-04T19:16:15.043Z] Failed to run mwoffliner after [1s]:
 Error: This wiki has no default skin:
[{"code":"metrolook","*":"Metrolook"},{"code":"vector","*":"Vector","unusable":""},{"code":"monobook_old","*":"MonoBook_OLD"},{"code":"monobook","*":"MonoBook","default":""},{"code":"cologneblue","*":"CologneBlue","unusable":""},{"code":"modern","*":"Modern","unusable":""}]
    at MediaWiki.getDefaultSkin (/tmp/mwoffliner/src/MediaWiki.ts:440:13)
    at MediaWiki.getSiteInfo (/tmp/mwoffliner/src/MediaWiki.ts:486:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Promise.all (index 0)
    at async MediaWiki.getMwMetaData (/tmp/mwoffliner/src/MediaWiki.ts:517:116)
    at async Module.execute (/tmp/mwoffliner/src/mwoffliner.lib.ts:194:18)
```

Markus-Rost:
The wiki at sigidwiki.com uses MediaWiki 1.22.3, which is below the minimal required version of 1.27

However I do not know how this failed to get caught by the mwMinimalVersion check.

evrial:
> The wiki at sigidwiki.com uses MediaWiki 1.22.3, which is below the minimal required version of 1.27
> 
> However I do not know how this failed to get caught by the mwMinimalVersion check.

I did some black wizardry to see what happens if ... :)

Markus-Rost:
> > The wiki at sigidwiki.com uses MediaWiki 1.22.3, which is below the minimal required version of 1.27
> > However I do not know how this failed to get caught by the mwMinimalVersion check.
> 
> I did some black wizardry to see what happens if ... :)

That would explain it :D 

The wiki doesn't have JSON formatversion 2 yet, which would return boolean values as `true` or `false` instead of the XML compatible version of having keys only on true and set to `""`. Therefore mwoffliner can't properly understand a lot of the API responses, like which skin is the default one.

evrial:
> The wiki doesn't have JSON formatversion 2 yet, which would return boolean values as `true` or `false` instead of the XML compatible version of having keys only on true and set to `""`. Therefore mwoffliner can't properly understand a lot of the API responses, like which skin is the default one.

Makes sense, so zero chance to make it work?

Markus-Rost:
> Makes sense, so zero chance to make it work?

Yeah it would break on pretty much every API request, including on the one getting the page content.

PR #2441, opened 2025-07-21
Use mainpage-title for Main Page display title

### LINKED ISSUE #2436: Hide/use correct main page display title
[`MediaWiki:Mainpage-title`](https://www.mediawiki.org/wiki/MediaWiki:Mainpage-title) is used to set the display title of the main page, but this is not returned as the parse api display title. If the message is set to an empty string (like on Wikipedia), the first heading gets hidden completely.

Markus-Rost:
Fix #2436

Use the mainpage-title system message for the wiki Main Page display title.

Also did some smaller fixes:
- Removed `openzim-page-title` element from the heading, it seems to be unused anyway
- Don't load user specific CSS and JS modules
- Remove filtering on html element classes
  - `client-nojs` will always be set anyway, because the client itself would change that class and we don't provide the inline JS doing that.
  - `vector-feature-night-mode-disabled` is only relevant for added the JS toggle We include neither the relevant JS nor the required HTML elements
  - `skin-theme-clientpref-day` should be provided in case any styles expect one of the theme classes to always exist


Markus-Rost:
> Please fix the failing e2e tests and we will be good to merge

I'm actually a bit stuck there, because I don't know where the unexpected reserved word is. I'm a bit blocked there by not being able to run the tests locally, so it would be great if you could point me in the right direction.

benoit74:
Don't you achieve to run tests with something like `npm run test:pattern 'test/e2e/en10.e2e.test.ts'`? Or is it linked to any other sort of issue with your local setup?

Anyway, I ran e2e tests locally and they are fine... relaunched it to check if it is not a transient issue

Markus-Rost:
I'm on windows, so I'm blocked by openzim/node-libzim#113. When I attempted to install node-libzim, it failed due to being on windows and aborted the installation of all dependencies. Now any attempted npm install of other packages will always try to install libzim as well for some reason and get aborted. So my local setup for mwoffliner is more plain text editing than IDE at the moment. 

I did try re-running the tests as well and they failed again, but maybe you have more luck.

benoit74:
No more luck ... what a nightmare ... I will try to isolate the issue on a distinct PR ...

Markus-Rost:
Has to be in the changes I made, so my guess is on either `cssClass` being used as a const as well as in the filter, or the const being named `message`, both in articles.ts

benoit74:
I just tested wrong branch ... I've just pushed the fix.

Problem was that you used `await` in the `describe` function, without marking the function as `async`. And marking the function as `async` does not help because `describe` does not want promises ...

And there was a small bug in the scraper code. I let you review my commit.

Markus-Rost:
Ah, thanks for catching that. I did not expect the error to be in the test because en10.e2e.test.ts was failing with the same error and I didn't touch that one.

PR #2450, opened 2025-07-25
Add langVariant argument

### LINKED ISSUE #2399: Add support for (chinese) variants

### LINKED ISSUE #2449: Remove Table of Contents from vector-legacy
The vector-2022 skin does not include the TOC inside the article body (unlike vector-legacy), this means we are currently missing a TOC on all ZIMs using that skin. We should build our own TOC and add it to the skin template.

Markus-Rost:
Fix #2399
Fix #2449

Medium term fix for #2399 by adding a `langVariant` argument. This replaces the temporary fix for #840 

Disable the TOC for all skins as this feature should be provided by the reader.


Markus-Rost:
Superseded by #2456

Issue #2210, opened 2025-04-01
Do we still need multiple renderers ?


benoit74:
https://github.com/openzim/mwoffliner/pull/2203 has introduced the ActionParse renderer which seems quite promising.

It is supposed to be well working on all Mediawiki instances, so we might wonder if we really want to continue to maintain all renderers, especially old/clunky ones like `VisualEditor`.

To be analyzed once we've got feedback from using ActionParse in the field.

kelson42:
I'm open on this. Each renderer should be the topic of a pro/con analysis.

Here are the points to take in consideration to me:
* What is the impact regarding "old" Mediawiki support
* Wikimedia dedicated API use to be faster have dedicated cache strategy
* We should anyway keep a perfectly modularity on this topic and (re)introduction of an end-point should not lead to architectural challenges
* We should be sure that actionparse API does the job properly (so its support pretty completly implemented) before considering removing the others

At this stage, to move this issue forward, an anlysis for each API endpoint is necessary

Considering this is strongly API breaking, I would prefer to have it in 2.0.0 to respect the sementic versioning.

benoit74:
We should probably drop all renderers except ActionParse. But keep the code structure to easily add a new renderer should we need this.

PR #2456, opened 2025-07-27
Automatically guess / retrieve mediawiki paths

### LINKED ISSUE #2399: Add support for (chinese) variants

### LINKED ISSUE #2422: Automatically guess / retrieve mediawiki paths

Markus-Rost:
Fix #2422
Fix #2399

- Remove `mwWikiPath` and `mwIndexPhpPath` CLI parameter, use values from the API instead
- Build `mwRestApiPath` and `mwModulePath` CLI parameter default values from the API
- Add `langVariant` CLI parameter for language conversion
  - Medium term fix for #2399, this replaces the temporary fix for #840

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -451,7 +451,7 @@ class MediaWiki {
     return defaultSkins[0]
   }
 
-  public async getSiteInfo() {
+  public async getSiteInfo({ mwWikiPath, mwIndexPhpPath, mwRestApiPath, mwModulePath, forceSkin, langVariant }: SiteInfoArgv = {}) {
     logger.log('Getting site info...')
     const body = await Downloader.querySiteInfo()
 
@@ -497,7 +497,36 @@ class MediaWiki {
       return possibleLangPairs[0] || ['en', 'eng']
     })
 
-    this.skin = this.getDefaultSkin(body.query.skins)
+    // Use CLI parameter and set MediaWiki config
+    if (mwWikiPath) {
+      mwWikiPath = mwWikiPath + '$1'
+      if (mwWikiPath !== generalEntries.articlepath) {
+        logger.warn(`mwWikiPath [${mwWikiPath}] does not match the path [${generalEntries.articlepath}] returned by the wiki.`)
+      }
+    }
+    const articlepath = mwWikiPath || generalEntries.articlepath
+    if (articlepath.includes('?') || !articlepath.endsWith('$1')) {
+      throw new Error(`Article path [${articlepath}] is not supported`)
+    }
+    this.wikiPath = articlepath.replace('$1', '')
+    if (mwIndexPhpPath && mwIndexPhpPath !== generalEntries.script) {
+      logger.warn(`mwIndexPhpPath [${mwIndexPhpPath}] does not match the path [${generalEntries.script}] returned by the wiki.`)
+    }
+    this.indexPhpPath = mwIndexPhpPath || generalEntries.script
+    this.restApiPath = mwRestApiPath || generalEntries.scriptpath + '/rest.php'
+    this.modulePathOpt = mwModulePath || generalEntries.scriptpath + '/load.php'
+
+    const skins: SiteInfoSkin[] = body.query.skins.filter((skin) => !skin.unusable)
+    if (forceSkin && !skins.map((skin) => skin.code).includes(forceSkin)) {
+      throw new Error(`Skin [${forceSkin}] is not usable on the wiki`)
+    }
+    this.skin = forceSkin || this.getDefaultSkin(skins)
+
+    const validLangVars = (generalEntries.variants || []).map((e: any) => e.code)
+    if (langVariant && !validLangVars.includes(langVariant)) {
+      throw new Error(`Language variant [${langVariant}] is not available on the wiki`)
+    }
+    const langVar = langVariant || null
 
     return {
       mainPage,
@@ -505,6 +534,7 @@ class MediaWiki {
       siteName,
       textDir,
       langMw,
+      langVar,
       langIso2,
       langIso3,
       logo,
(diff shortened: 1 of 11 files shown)


kelson42:
I won't make the full review, but I would be carreful about proposing PR fixing many different things at the same time.

I would recommend to multiple PR, with each one based on the branch of the other.

Markus-Rost:
> I won't make the full review, but I would be carreful about proposing PR fixing many different things at the same time.
> 
> I would recommend to multiple PR, with each one based on the branch of the other.

I'll see if I can split them, however at least `langVariant` and the path options are in the same area and therefore can't be split

Markus-Rost:
Test currently fails due to #2468 so we need to merge #2469 first

benoit74:
@Markus-Rost regarding `langVariant`, do I get it correctly that with this PR merged: 
- if we do not pass `--langVariant`, we will have "original" content, probably mixing multiple variants on WPZH, just like before `1.16.0`
- if we pass `--langVariant zh-cn` to WPZH, we will get the same result as `1.16.0`

Markus-Rost:
> @Markus-Rost regarding `langVariant`, do I get it correctly that with this PR merged:
> 
> * if we do not pass `--langVariant`, we will have "original" content, probably mixing multiple variants on WPZH, just like before `1.16.0`
> * if we pass `--langVariant zh-cn` to WPZH, we will get the same result as `1.16.0`

That is correct.

benoit74:
@Markus-Rost please rebase + force push

Markus-Rost:
> @Markus-Rost please rebase + force push

Done, tests now failing due to nodejs/node#59480

benoit74 (changes requested):
Thank you very much, I really like the middle ground you achieved between guessing as much as possible but not guessing the impossible. Having to specify at least wiki URL + API location seems very reasonable.

I'm a bit concerned that removing the two CLI parameter are a breaking change from semantic versioning perspective, so we should rename 1.17.0 to 2.0.0 ; do you think we could postpone CLI parameter removal to later and only compute default values for the time being? I feel like it is a kind of additional burden. But breaking semantic versioning is not really good practice either. And I don't want to publish a 2.0.0 now since we will need a new major once we implement #2210. Or maybe it is time to implement #2210. @kelson42 WDYT, are you ready for #2210 now?

Markus-Rost:
> do you think we could postpone CLI parameter removal to later and only compute default values for the time being?

How about marking them both as deprecated options which won't do anything anymore? That should prevent the change from being breaking

kelson42:
Please keep the working options, update the usage(), and create issue to remove them in 2.0. All of this should be trivial. To me, we just have now a far better/smarter way to decide what to use if the user is not specific.

kelson42:
> @Markus-Rost regarding `langVariant`, do I get it correctly that with this PR merged:
> 
> * if we do not pass `--langVariant`, we will have "original" content, probably mixing multiple variants on WPZH, just like before `1.16.0`
> * if we pass `--langVariant zh-cn` to WPZH, we will get the same result as `1.16.0`

We should probably corresponding issues for Zimfarm and Zim-requests, so recipes will be updated for mwoffliner 1.17.

Markus-Rost:
> Please keep the working options, update the usage(), and create issue to remove them in 2.0. All of this should be trivial. To me, we just have now a far better/smarter way to decide what to use if the user is not specific.

I've marked the options as deprecated but kept them working. Should they be set to a value different from the path return by the api, a warning will be logged. I have create #2489 to remove them in 2.0

src/MediaWiki.ts
```diff
@@ -497,14 +497,44 @@ class MediaWiki {
       return possibleLangPairs[0] || ['en', 'eng']
     })
 
-    this.skin = this.getDefaultSkin(body.query.skins)
+    // Use CLI parameter and set MediaWiki config
+    if (mwWikiPath) {
+      mwWikiPath = mwWikiPath + '$1'
+      if (mwWikiPath !== generalEntries.articlepath) {
+        logger.warn('mwWikiPath does not match the value returned by the wiki.')
```
benoit74: Please log both values as well, so that it is easier / faster to diagnose

src/MediaWiki.ts
```diff
@@ -497,14 +497,44 @@ class MediaWiki {
...
+      }
+    }
+    const articlepath = mwWikiPath || generalEntries.articlepath
+    if (articlepath.includes('?') || !articlepath.endsWith('$1')) {
+      throw new Error(`Article path [${articlepath}] is not supported`)
+    }
+    this.wikiPath = articlepath.replace('$1', '')
+    if (mwIndexPhpPath && mwIndexPhpPath !== generalEntries.script) {
+      logger.warn('mwIndexPhpPath does not match the value returned by the wiki.')
```
benoit74: Please log both values as well, so that it is easier / faster to diagnose

benoit74:
> We should probably corresponding issues for Zimfarm and Zim-requests, so recipes will be updated for mwoffliner 1.17

This is part of the release process. No need to open a specific issue for now. At release, I will just like usually create a Zimfarm PR to adjust flags, update recipes to adjust existing settings, update mwoffliner wiki, ...

PR #2458, opened 2025-07-27
Merge initial API calls fetching site info

### LINKED ISSUE #2385: Merge calls to `buildGadgetsUrl`, `buildSiteInfoURL` and `buildNamespacesURL`
We make three distinct calls at the beginning of the scraper where one could be sufficient.

I'm not sure about the exact benefit / downside, so to be investigated.

Benefit seems to be that we make less request and hence less load on the upstream.

Downside is that results are harder to process since it is a mix of multiple data, especially considering the fact that this data might be paginated. I'm a bit concerned it might make code a bit too hard to read.

Markus-Rost:
Fix #2385

Depends on #2456

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -26,6 +26,7 @@ Unreleased:
 * CHANGED: Deprecate `mwWikiPath` and `mwIndexPhpPath` CLI parameter, use values from the API instead (@Markus-Rost #2422)
 * CHANGED: Build `mwRestApiPath` and `mwModulePath` CLI parameter default values from the API (@Markus-Rost #2422)
 * NEW: Add `langVariant` CLI parameter for language conversion (@Markus-Rost #2399)
+* FIX: Merge initial API calls fetching site info (@Markus-Rost #2385)
 
 1.16.0:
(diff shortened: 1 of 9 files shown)


benoit74 (commented):
I do not see anything odd in code, but it looks like this broke something given CI results. Can you have a look?

Markus-Rost:
> I do not see anything odd in code, but it looks like this broke something given CI results. Can you have a look?

@benoit74 I noticed the failures as well but so far haven't been able to find the cause for them. I would appreciate it if you could take a look as well.

benoit74:
Looks like it was (as usual) a pretty silly bug: in `Mediawiki` class, metadata are cached but not namespaces which are reset when calling `reset()` function. And namespaces are updated only when we have to update metadata and hence get site info again.

So when calling the scraper multiple times, all calls but the first one had no namespaces to retrieve ... I feel like on reset we should reset metadata as well, so this is what I changed.

Let's see if this is the only bug

benoit74 (approved):
LGTM

benoit74:
Please rebase, squash commits and merge.

PR #2462, opened 2025-07-28
Support main page at domain root

### LINKED ISSUE #2461: Links to main page are broken if main page is at domain root
Wikis using `$wgMainPageIsDomainRoot` have the main page located at the domain root. This breaks `extractPageTitleFromHref` which does not identify a link to the domain root to actually be a link to the main page title.

https://dev.library.kiwix.org/viewer#minecraftwiki_en_all_maxi_2025-07/Movie%3AWiki

Image

Links to https://dev.library.kiwix.org/ instead of https://dev.library.kiwix.org/content/minecraftwiki_en_all_maxi_2025-07/Minecraft_Wiki

Markus-Rost:
Fix #2461

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -13,6 +13,7 @@ Unreleased:
 * FIX: Avoid path conflicts between articles, assets and resources (@benoit74 #2333 #2408)
 * NEW: Use ZIM illustration as HTML favicon (@benoit74 #2437)
 * FIX: Restore webp polyfills operation (@benoit74 #2438)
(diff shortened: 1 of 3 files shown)


benoit74 (changes requested):
Please add an e2e test case so that we check this continue to work as expected in the future.

Markus-Rost:
> Please add an e2e test case so that we check this continue to work as expected in the future.

I've added `mainPageIsDomainRoot.e2e.test.ts`, using the article subtitle of a subpage because that seems like the most reliable place to have a link to the main page long term.

PR #2463, opened 2025-07-28
Fix absolute file links

### LINKED ISSUE #2459: portal namespace renders as external links on he.wikipedia
Haven't tried on other wikis, but on the Hebrew Wikipedia the portal links (`פורטל:` namespace) that are front and center to the mobile front page are rendered as external links in the ZIM file:

Image

Markus-Rost:
Fix #2459

Treat absolute URLs to the wiki article path as local links. The check ensures we only do this for absolute URLs where `extractPageTitleFromHref` will always return a title.

Also made sure to rewrite any local links we couldn't find a title for and stop removing all anchors from redirect links.

### DIFF
--- a/src/util/rewriteUrls.ts
+++ b/src/util/rewriteUrls.ts
@@ -130,7 +130,7 @@ function rewriteUrlNoArticleCheck(articleId: string, dump: Dump, linkNode: Domin
     }
   }
 
-  if (rel) {
+  if (rel && !href.startsWith(MediaWiki.webUrl.href)) {
     // This is Parsoid HTML
 
     /* Add 'external' class to interwiki links */
@@ -159,6 +159,9 @@ function rewriteUrlNoArticleCheck(articleId: string, dump: Dump, linkNode: Domin
     return title
   }
 
+  // Rewrite any urls still remaining
+  DU.appendToAttr(linkNode, 'class', 'external')
+  linkNode.setAttribute('href', getFullUrl(href, MediaWiki.baseUrl))
   return null
 }
 
@@ -207,12 +210,7 @@ async function rewriteUrls(articleId: string, dump: Dump, linkNodes: DominoEleme
     const articlesRedirected = await RedisStore.redirectsXId.existsMany(unmirroredTitles)
     for (const articleTitle of unmirroredTitles) {
       const redirect = articlesRedirected[articleTitle]
-      if (redirect) {
-        const href = encodeArticleIdForZimHtmlUrl(articleTitle)
-        wikilinkMappings[articleTitle].forEach((linkNode: DominoElement) => {
-          linkNode.setAttribute('href', href)
-        })
-      } else {
+      if (!redirect) {
         wikilinkMappings[articleTitle].forEach((linkNode: DominoElement) => {
           migrateChildren(linkNode, linkNode.parentNode, linkNode)
           linkNode.parentNode.removeChild(linkNode)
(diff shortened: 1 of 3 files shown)


src/util/rewriteUrls.ts
```diff
@@ -196,12 +202,7 @@ async function rewriteUrls(articleId: string, dump: Dump, linkNodes: DominoEleme
     const articlesRedirected = await RedisStore.redirectsXId.existsMany(unmirroredTitles)
     for (const articleTitle of unmirroredTitles) {
       const redirect = articlesRedirected[articleTitle]
-      if (redirect) {
-        const href = encodeArticleIdForZimHtmlUrl(articleTitle)
-        wikilinkMappings[articleTitle].forEach((linkNode: DominoElement) => {
-          linkNode.setAttribute('href', href)
-        })
-      } else {
+      if (!redirect) {
```
benoit74, line 213: I don't get why we don't process redirects anymore.

src/util/rewriteUrls.ts
```diff
@@ -148,6 +148,12 @@ function rewriteUrlNoArticleCheck(articleId: string, dump: Dump, linkNode: Domin
     return title
   }
 
+  // Rewrite any relative urls still remaining
```
benoit74: Is this about relative or absolute URL, I don't get it. And I don't get why we need a special treatment here only for relative/absolute URLs.

benoit74 (commented):
I'm a bit troubled by this PR. Could you please add tests (unit tests are probably sufficient) so that it gets a bit clearer about what we had before and what we will get now?

src/util/rewriteUrls.ts
Markus-Rost: We didn't actually process anything here. `articleTitle` is still the same as we already set in Line 147, except now we are removing any anchor from the link we previously were keeping.

src/util/rewriteUrls.ts
Markus-Rost: Uhm, I think this kinda ended up in the wrong PR. 😅 

Anyhow, this is a fallback solution for issues like #2461 where `extractPageTitleFromHref` returned no title. This turns any relative urls into absolute urls to prevent them from pointing to random places inside the ZIM reader.

Markus-Rost:
> I'm a bit troubled by this PR. Could you please add tests (unit tests are probably sufficient) so that it gets a bit clearer about what we had before and what we will get now?

This fix for the linked issue is just the change in line 122, but I had two other improvements for the rewriteUrls file and included them here, making this PR a bit messy.

I've added tests for turning absolute wiki links into internal links as well as for the relative url fallback check.

src/util/rewriteUrls.ts
Markus-Rost: Ideally this code should never be reached, but if a link does slip through it's better to have the fallback than broken links inside the ZIM.

benoit74: Got it, I got also confused because I confused absolute URL with absolute path ...

Shouldn't we also consider relative path like './xxxx'? Do they really exists? I suspect we always have absolute paths, but better to ask than being sorry in few days ^^

src/util/rewriteUrls.ts
benoit74: Got it, thank you.

src/util/rewriteUrls.ts
Markus-Rost: I don't think I have ever seen relative paths in MediaWiki, though considering this is already a fallback, I see no reason not to account for them as well

benoit74: OK, please ask for review again once you've adapted code to consider relative paths as well. Thank you!

Markus-Rost: I've just removed the `if` around the rewriting as it won't change any absolute urls reaching that place and anchor-only links escape early. So everything that is able to reach this part should be fine with this rewriting.

benoit74: Indeed

Issue #2479, opened 2025-08-18
Links of mdwiki.org `Graves'_disease` article are badly rewritten


benoit74:
See https://github.com/openzim/zim-requests/issues/1260#issuecomment-3194708559

Two issues:
- the "muscle weakness" link points to online article while they should be removed (we remove links to articles which were not scraped)
- the "muscle weakness" link point to the wrong article (Myopathy instead of Muscle_weakness)

Jaifroid:
Please note I identified the issue when testing the test ZIM in https://github.com/kiwix/kiwix-js-pwa/issues/651#issuecomment-3191652182, but `Graves'_disease` is just an example. Exactly the same is happening on many other pages. For example, in, `Hyperthyroidism` the first four links are all absolute links to mdwiki.

Markus-Rost:
This is happening for all links to mirrored pages, as they include `rel="nofollow"` and as such get treated as external links:

`<a href="/wiki/Muscle_weakness" class="mirror-link" title="Muscle weakness" rel="nofollow">muscle weakness</a>`

This wasn't an issue before #2391, because Parsoid treats mirror links like normal internal links:

`<a rel="mw:WikiLink" href="https://mdwiki.org/wiki/Muscle_weakness" title="Muscle weakness" id="mwIw">muscle weakness</a>`

Jaifroid:
> the "muscle weakness" link point to the wrong article (Myopathy instead of Muscle_weakness)

@benoit74 That's just a redirect. There are tons of such cases, where a redirect points to a differently named underlying article.

Markus-Rost:
> the "muscle weakness" link point to the wrong article (Myopathy instead of Muscle_weakness)

It points to the "correct" article at the time of the scrape, the link on the wiki got updated only afterwards:
https://mdwiki.org/w/index.php?title=Graves%27_disease&diff=prev&oldid=1475497

PR #2482, opened 2025-08-18
Get the whole article when the specified section is not available

### LINKED ISSUE #2480: Some mdwiki.org do not have section 0 despite being wikitext contentmodel
See https://github.com/openzim/zim-requests/issues/1260#issuecomment-3188100097

This is a follow-up to https://github.com/openzim/mwoffliner/issues/2444

Markus-Rost:
Fix #2480

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -131,7 +131,7 @@ async function getAllArticlesToKeep(articleDetailXId: RKVS<ArticleDetail>, dump:
       try {
         const mainPage = isMainPage(articleId)
         const renderer = mainPage ? mainPageRenderer : articlesRenderer
-        const leadSectionId = dump.nodet && !articleDetail.contentmodel ? config.filters.leadSectionId : ''
+        const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
         const articleUrl = mainPage ? Downloader.getMainPageUrl(articleId) : Downloader.getArticleUrl(articleId, { sectionId: leadSectionId })
 
         rets = await Downloader.getArticle(articleId, articleDetailXId, renderer, articleUrl, dump, articleDetail)
@@ -306,7 +306,7 @@ export async function saveArticles(zimCreator: Creator, dump: Dump) {
         try {
           const mainPage = isMainPage(articleId)
           const renderer = mainPage ? RenderingContext.mainPageRenderer : RenderingContext.articlesRenderer
-          const leadSectionId = dump.nodet ? config.filters.leadSectionId : ''
+          const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
           const articleUrl = mainPage ? Downloader.getMainPageUrl(articleId) : Downloader.getArticleUrl(articleId, { sectionId: leadSectionId })
 
           rets = await Downloader.getArticle(articleId, articleDetailXId, renderer, articleUrl, dump, articleDetail)
--- a/Changelog
+++ b/Changelog
@@ -29,6 +29,7 @@ Unreleased:
 * FIX: Merge initial API calls fetching site info (@Markus-Rost #2385)
 * FIX: Rewrite non-Parsoid links properly (@Markus-Rost #2479)
 * FIX: Don't reserve space for elements that don't exist (@Markus-Rost #2495)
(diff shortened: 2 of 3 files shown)


benoit74:
Given the doubt I/we have in #2480, I suggest that we split this PR in two and focus only on #2479 for now. Especially since #2479 is the only real significant problem impacting mdwiki.org which need to be solved asap. The `nodet` version of mdwiki.org has never been published by Kiwix, so let's take time to reflect a bit.

Please add a test to showcase what we expect to receive for mirror articles with native parser (or an e2e test on mdwiki.org, I don't mind much).

Markus-Rost:
> Given the doubt I/we have in #2480, I suggest that we split this PR in two and focus only on #2479 for now. Especially since #2479 is the only real significant problem impacting mdwiki.org which need to be solved asap. The `nodet` version of mdwiki.org has never been published by Kiwix, so let's take time to reflect a bit.
> 
> Please add a test to showcase what we expect to receive for mirror articles with native parser (or an e2e test on mdwiki.org, I don't mind much).

I have created #2494 for #2479, rewriting the native parser links properly.

Keeping the change to get the whole article when the specified section is not available in this PR. I don't think hitting the `nosuchsection` error on a page should completely prevent a `nodet` run. And just not including the page at all doesn't seem like a great solution either.

src/util/saveArticles.ts
```diff
@@ -131,7 +131,7 @@ async function getAllArticlesToKeep(articleDetailXId: RKVS<ArticleDetail>, dump:
       try {
         const mainPage = isMainPage(articleId)
         const renderer = mainPage ? mainPageRenderer : articlesRenderer
-        const leadSectionId = dump.nodet && !articleDetail.contentmodel ? config.filters.leadSectionId : ''
+        const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
```
benoit74, line 134: Shouldn't we apply the same condition in `saveArticles`? I just realize it already misses the condition on `!articleDetail.contentmodel`.

benoit74:
Let me rephrase the situation and correct me if I'm wrong.

When an article is mirrored on mdwiki.org, it seems that we always have:
- `missing` property set to `true`
- content comes from Parsoid and hence has `<section>`
- we cannot ask for `section=0` to `action=parse` because these are not real sections (see https://mdwiki.org/w/api.php?action=parse&format=json&prop=modules|jsconfigvars|headhtml|text|displaytitle|subtitle&usearticle=1&disabletoc=1&disableeditsection=1&disablelimitreport=1&page=List_of_biobanks&useskin=vector&redirects=1&formatversion=2&section=0)

This causes problem when building the `nodet` flavor of mdwiki.org.

What you propose is two folds:
- first, whenever `missing` article property is `true`, we do not request `section=0` because we known it is deemed to fail
- second, whenever we get a `nosuchsection` error, we fallback to not asking for `section=0`

In both cases, this will request the whole article content, but:
- it is generally OK on mdwiki.org because article content seems to always come from Parsoid, have `<section>` elements and the scraper will take care of keeping only the first section (see * below)
- it is probably pretty not doing much harm on other conditions: if the article fails to provide section=0 + does not have `<section>` elements (we never saw this in the wild) then we will just have too much HTML

Is it correct? Please correct where I'm wrong / misunderstanding

*code deleting all sections but the first one:

https://github.com/openzim/mwoffliner/blob/3a3f4f9826522c853b81ca341f1e24a98dc40242/src/renderers/abstract.renderer.ts#L641-L651

src/util/saveArticles.ts
Markus-Rost: Yes we should, added that.

Markus-Rost:
@benoit74 Yes, that is all correct.

benoit74:
@kelson42 can you please have a look at https://github.com/openzim/mwoffliner/pull/2482#issuecomment-3213850757 and advise if this seems OK to you or ask for a live discussion?

benoit74 (approved):
Let's merge, LGTM and rollback is pretty easy.

@Markus-Rost please rebase and merge.

kelson42:
> @kelson42 can you please have a look at https://github.com/openzim/mwoffliner/pull/2482#issuecomment-3213850757 and advise if this seems OK to you or ask for a live discussion?

Sorry for late feedback.... but unsure if it will do the job  I will check when mdwiki will be scrapable again.

Issue #1491, opened 2021-07-05
Ancestry table not expanded


Inbefortus:
Kiwix Version: 3.4.4
ZIM: 2021-03 English Wikipedia

The table carrying ancestry is not unfolded.

Kiwix:

![Screenshot_20210705-092217_Kiwix](https://user-images.githubusercontent.com/71934042/124435889-bc009900-dd75-11eb-895d-a4a5d448ec40.jpg)

![Screenshot_20210705-092901_Kiwix](https://user-images.githubusercontent.com/71934042/124435893-bd31c600-dd75-11eb-924f-9622a8c92812.jpg)

Original English Wikipedia:

![Screenshot_20210705-092250_Samsung Internet](https://user-images.githubusercontent.com/71934042/124435944-cc187880-dd75-11eb-9002-97600efc3d3d.jpg)

![Screenshot_20210705-093032_Samsung Internet](https://user-images.githubusercontent.com/71934042/124435947-cd49a580-dd75-11eb-8e82-4df75530eee2.jpg)

kelson42:
Which article is that?

Inbefortus:
> Which article is that?

https://en.m.wikipedia.org/wiki/Louis_Antoine,_Duke_of_Enghien
and
https://en.m.wikipedia.org/w/index.php?title=Henry_III_of_England&oldid=1011571233&mobileaction=toggle_view_mobile

Inbefortus:
@kelson42 I think this should rather be transferred to Kiwix Android as the ancestor table is observable in Kiwix JS PWA and not in  Android and mwoffliner did everything just right.

Kiwix JS PWA:

![Screenshot_20210727-092747_Samsung Internet](https://user-images.githubusercontent.com/71934042/127126379-b7d0e4ec-6c19-43b4-94d2-e51f3afe01d6.jpg)
![Screenshot_20210727-092847_Samsung Internet](https://user-images.githubusercontent.com/71934042/127126385-f76ad95d-d9f8-4245-b21f-8d3364626583.jpg)

kelson42:
@Inbefortus Thx for this new info. But I will run the investigation first here and then triage the ticket accordingly. In general, the reader don't and shouldn't manipulate in any manner the content... and the way how things are displayed. Therefore, for the moment, I believe the ticket is at the right place.

kelson42:
I guess it's fixed now. I will check.

kelson42:
[...]

PR #2483, opened 2025-08-18
Add JavaScript option

### LINKED ISSUE #2310: Wikipedia articles scraped with ActionParse API contain scripts that attempt to load external content

Markus-Rost:
Fix #2310

- Add `--javaScript` option with values "none", "trusted" or "all" (default being "trusted").
- Add `--addModules` option for additional ResourceLoader modules

### DIFF
--- a/res/templates/javaScript.html
+++ b/res/templates/javaScript.html
@@ -1,0 +1,8 @@
+<script id="mwoffliner-jsConfigVars">
+      document.documentElement.classList.replace('client-nojs', 'client-js');
+      RLCONF = __ARTICLE_CONFIGVARS__;
+      RLSTATE = __ARTICLE_CSS_STATE__;
+      RLPAGEMODULES = __ARTICLE_JS_MODULES__;
+      
+      (RLQ=window.RLQ||[]).push(function(){mw.loader.impl(function(){return["user.options@12s5i",function($,jQuery,require,module){mw.user.tokens.set({"patrolToken":"+\\","watchToken":"+\\","csrfToken":"+\\"});}];});});
+    </script>
--- a/src/util/misc.ts
+++ b/src/util/misc.ts
@@ -379,6 +379,25 @@ export function sanitizeString(str: string) {
   return str.replace(/[&<>"'*=//]/g, ' ')
 }
 
+const unsafeJsCharMap = {
+  '<': '\\u003C',
+  '>': '\\u003E',
+  '&': '\\u0026',
+  '\\': '\\\\',
+  '\b': '\\b',
+  '\f': '\\f',
+  '\n': '\\n',
+  '\r': '\\r',
+  '\t': '\\t',
+  '\0': '\\0',
+  '\u2028': '\\u2028',
+  '\u2029': '\\u2029',
+}
+/** Escaped JSON.stringify for usage in inline scripts */
+export function jsonStringify(data: any) {
+  return JSON.stringify(data).replace(/[<>&\\\b\f\n\r\t\0\u2028\u2029]/g, (char) => unsafeJsCharMap[char] || char)
+}
+
 // We will need the encoded URL on article load so that we can set the hrefs of anchor tag correctly,
 // but we must not encode the '/' character or else relative links may fail
 export function encodeArticleIdForZimHtmlUrl(articleId: string) {
--- a/Changelog
+++ b/Changelog
@@ -1,7 +1,9 @@
 Unreleased:
+* NEW: Add support for ResourceLoader based JavaScript (@Markus-Rost #2310)
 * CHANGED: Keep only ActionParse renderer (@benoit74 #2210)
 
 1.17.5:
+* FIX: Ignore articles returning permissiondenied error code (@kunalsiyag #2604)
 * NEW: Add validation for internationalization (i18n) files (@ShreyashSri #2563)
 * CHANGED: Update README to recommend Docker image installation (@kunalsiyag #2601)
 * CHANGED: Upgrade @openzim/libzim to v4.0.0 (@dhyaniritesh89 #2590)
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -270,6 +270,23 @@
       "title": "Language Variant",
       "description": "Use a specific language variant, only for wikis supporting language conversion",
       "required": false
+    },
+    "javaScript": {
+      "type": "string-enum",
+      "title": "JavaScript",
+      "description": "Amount of JavaScript being allowed in pages (default: trusted)",
+      "required": false,
+      "choices": [
+        { "title": "NONE", "value": "none" },
+        { "title": "TRUSTED", "value": "trusted" },
+        { "title": "ALL", "value": "all" }
+      ]
+    },
+    "addModules": {
+      "type": "string",
+      "title": "Additional modules",
+      "description": "Add additional ResourceLoader modules for dynamic loading (comma separated list)",
+      "required": false
     }
   }
 }
--- a/res/templates/pageFallback.html
+++ b/res/templates/pageFallback.html
@@ -3,13 +3,15 @@
   <head>
     <meta charset="UTF-8" />
     <title></title>
+    __ARTICLE_JAVASCRIPT__
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/png" href="__RELATIVE_FILE_PATH____RES_DIR__/favicon.png" />
     __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_BEFORE_META__
+    __ARTICLE_JS_STARTUP__
     <meta name="ResourceLoaderDynamicStyles" content="" />
     __ARTICLE_CSS_AFTER_META__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/site.styles.css" />
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/noscript.css" />
+    __ARTICLE_CSS_NOSCRIPT__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____RES_DIR__/footer.css" />
   </head>
 
@@ -24,6 +26,6 @@
         <div id="mw-content-text" class="mw-body-content"></div>
       </div>
     </div>
-    __ARTICLE_CONFIGVARS_LIST__ __ARTICLE_JS_LIST__
+    __ARTICLE_JS_LIST__
   </body>
 </html>
--- a/res/templates/pageVector2022.html
+++ b/res/templates/pageVector2022.html
@@ -3,13 +3,15 @@
   <head>
     <meta charset="UTF-8" />
     <title></title>
+    __ARTICLE_JAVASCRIPT__
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/png" href="__RELATIVE_FILE_PATH____RES_DIR__/favicon.png" />
     __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_BEFORE_META__
+    __ARTICLE_JS_STARTUP__
     <meta name="ResourceLoaderDynamicStyles" content="" />
     __ARTICLE_CSS_AFTER_META__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/site.styles.css" />
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/noscript.css" />
+    __ARTICLE_CSS_NOSCRIPT__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____RES_DIR__/footer.css" />
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____RES_DIR__/vector-2022.css" />
   </head>
@@ -32,6 +34,6 @@
         </div>
       </div>
     </div>
-    __ARTICLE_CONFIGVARS_LIST__ __ARTICLE_JS_LIST__
+    __ARTICLE_JS_LIST__
   </body>
 </html>
--- a/res/templates/pageVectorLegacy.html
+++ b/res/templates/pageVectorLegacy.html
@@ -3,13 +3,15 @@
   <head>
     <meta charset="UTF-8" />
     <title></title>
+    __ARTICLE_JAVASCRIPT__
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" type="image/png" href="__RELATIVE_FILE_PATH____RES_DIR__/favicon.png" />
     __ARTICLE_CANONICAL_LINK__ __ARTICLE_CSS_BEFORE_META__
+    __ARTICLE_JS_STARTUP__
     <meta name="ResourceLoaderDynamicStyles" content="" />
     __ARTICLE_CSS_AFTER_META__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/site.styles.css" />
-    <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____MW_DIR__/noscript.css" />
+    __ARTICLE_CSS_NOSCRIPT__
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____RES_DIR__/footer.css" />
     <link rel="stylesheet" type="text/css" href="__RELATIVE_FILE_PATH____RES_DIR__/vector.css" />
   </head>
@@ -25,6 +27,6 @@
         <div id="mw-content-text" class="mw-body-content"></div>
       </div>
     </div>
-    __ARTICLE_CONFIGVARS_LIST__ __ARTICLE_JS_LIST__
+    __ARTICLE_JS_LIST__
   </body>
 </html>
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -3,7 +3,6 @@ import { config } from './config.js'
 import { normalizeMwResponse, DB_ERROR, WEAK_ETAG_REGEX, stripHttpFromUrl, isBitmapImageMimeType, isWebpCandidateImageMimeType } from './util/index.js'
 import { Readable } from 'stream'
 import deepmerge from 'deepmerge'
-import * as domino from 'domino'
 import { default as imagemin } from 'imagemin'
 import imageminAdvPng from 'imagemin-advpng'
 import type { BackoffStrategy } from 'backoff'
@@ -56,6 +55,7 @@ interface DownloaderOpts {
   optimisationCacheUrl: string
   s3?: S3
   webp: boolean
+  trustedJs?: string[]
   backoffOptions?: BackoffOptions
   insecure?: boolean
 }
@@ -120,6 +120,7 @@ class Downloader {
   private _arrayBufferRequestOptions: AxiosRequestConfig
   private _jsonRequestOptions: AxiosRequestConfig
   private _streamRequestOptions: AxiosRequestConfig
+  public trustedJs: string[] = []
 
   private uaString: string
   private backoffOptions: BackoffOptions
@@ -157,13 +158,14 @@ class Downloader {
     return this._apiUrlDirector
   }
 
-  set init({ uaString, speed, reqTimeout, optimisationCacheUrl, s3, webp, backoffOptions, insecure }: DownloaderOpts) {
+  set init({ uaString, speed, reqTimeout, optimisationCacheUrl, s3, webp, trustedJs = config.output.mw.js_trusted.slice(), backoffOptions, insecure }: DownloaderOpts) {
     this.reset()
     this.uaString = uaString
     this._speed = speed
     this._requestTimeout = reqTimeout
     this.optimisationCacheUrl = optimisationCacheUrl
     this._webp = webp
+    this.trustedJs = trustedJs
     this.s3 = s3
     this._apiUrlDirector = new ApiURLDirector(MediaWiki.actionApiUrl.href)
     this.insecure = insecure
@@ -266,6 +268,7 @@ class Downloader {
     this._requestTimeout = undefined
     this.optimisationCacheUrl = undefined
     this._webp = false
+    this.trustedJs = []
     this.s3 = undefined
     this._apiUrlDirector = undefined
     this.insecure = false
@@ -842,27 +845,6 @@ class Downloader {
       stream.on('error', reject)
       stream.on('end', () => resolve(Buffer.concat(chunks)))
     })
-  }
-
-  public static extractJsConfigVars(headhtml: string) {
-    let jsConfigVars = ''
-
-    // Saving, as a js module, the jsconfigvars that are set in the header of a wikipedia page
-    // the script below extracts the config with a regex executed on the page header returned from the api
-    const scriptTags = domino.createDocument(`${headhtml}</body></html>`).getElementsByTagName('script')
-    const regex = /mw\.config\.set\(\{.*?\}\);/gm
-    for (let i = 0; i < scriptTags.length; i += 1) {
-      if (scriptTags[i].text.includes('mw.config.set')) {
-        jsConfigVars = regex.exec(scriptTags[i].text)[0] || ''
-        jsConfigVars = `(window.RLQ=window.RLQ||[]).push(function() {${jsConfigVars}});`
-      } else if (scriptTags[i].text.includes('RLCONF') || scriptTags[i].text.includes('RLSTATE') || scriptTags[i].text.includes('RLPAGEMODULES')) {
-        jsConfigVars = scriptTags[i].text
-      }
-    }
-
-    jsConfigVars = jsConfigVars.replace('nosuchaction', 'view') // to replace the wgAction config that is set to 'nosuchaction' from api but should be 'view'
-
-    return jsConfigVars
   }
 }
 
--- a/src/Gadgets.ts
+++ b/src/Gadgets.ts
@@ -37,7 +37,7 @@ class Gadgets {
       const module = gadget.metadata.module
       if (module.peers && module.peers.length) {
         // Only JS Gadgets can have peers
-        cssGadgets.concat(module.peers)
+        cssGadgets.push(...module.peers)
         return jsGadgets.push(gadget.id)
       }
       if (module.scripts && module.scripts.length) return jsGadgets.push(gadget.id)
--- a/src/Templates.ts
+++ b/src/Templates.ts
@@ -38,6 +38,10 @@ const htmlRedirectTemplateCode = () => {
   return readTemplate(config.output.templates.htmlRedirect)
 }
 
+const javaScriptTemplateCode = () => {
+  return readTemplate(config.output.templates.javaScript)
+}
+
 const articleListHomeTemplate = readTemplate(config.output.templates.articleListHomeTemplate)
 
 export {
@@ -49,6 +53,7 @@ export {
   htmlVector2022TemplateCode,
   htmlFallbackTemplateCode,
   htmlRedirectTemplateCode,
+  javaScriptTemplateCode,
   articleListHomeTemplate,
   categoriesTemplate,
   subCategoriesTemplate,
--- a/src/config.ts
+++ b/src/config.ts
@@ -69,12 +69,33 @@ const config = {
     mw: {
       css: [
         'site.styles', // always needed
-        'noscript', // recommended until we solve https://github.com/openzim/mwoffliner/issues/2310
+        'noscript', // in case client has JS disabled
       ],
       js: [
         // base JS scripts always needed / never returned on API calls
-        'startup',
+        'jquery',
+        'mediawiki.base',
       ],
+      js_trusted: [
+        // JS modules and their dependencies trusted to not request any external resources
+        'jquery',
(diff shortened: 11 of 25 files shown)


Markus-Rost:
This draft is currently made as a proof of concept with a disregard for any other renderer it might be breaking.

Markus-Rost:
I created some example ZIM files including no JavaScript, all JavaScript or only JavaScript trusted not to request external content. Please take a look at them and let me know how you feel about them.

- All JavaScript disabled
  - [wikipedia_en_10_JS_none_2025-08.zip](https://github.com/user-attachments/files/21842690/wikipedia_en_10_JS_none_2025-08.zip)
  - `mwoffliner --mwUrl=https://en.wikipedia.org/ --articleList=http://download.openzim.org/wp1/enwiki/tops/10.tsv --javaScript=none`
- Only trusted JavaScript enabled
  - [wikipedia_en_10_JS_trusted_2025-08.zip](https://github.com/user-attachments/files/21842670/wikipedia_en_10_JS_trusted_2025-08.zip)
  - `mwoffliner --mwUrl=https://en.wikipedia.org/ --articleList=http://download.openzim.org/wp1/enwiki/tops/10.tsv --javaScript=trusted`
- Trusted JavaScript enabled, as well as some additional user created scripts not requesting external content
  - [wikipedia_en_10_JS_trusted_addModules_2025-08.zip](https://github.com/user-attachments/files/21842669/wikipedia_en_10_JS_trusted_addModules_2025-08.zip)
  - `mwoffliner --mwUrl=https://en.wikipedia.org/ --articleList=http://download.openzim.org/wp1/enwiki/tops/10.tsv --javaScript=trusted --addModules=site,ext.gadget.ReferenceTooltips,oojs,oojs-ui,ext.gadget.switcher,ext.gadget.Calculator,@wikimedia/codex`
- All JavaScript enabled, including scripts requesting external content
  - [wikipedia_en_10_JS_all_2025-08.zip](https://github.com/user-attachments/files/21842671/wikipedia_en_10_JS_all_2025-08.zip)
  - `mwoffliner --mwUrl=https://en.wikipedia.org/ --articleList=http://download.openzim.org/wp1/enwiki/tops/10.tsv --javaScript=all`

benoit74:
I tried PoC ZIMs.

For some reasons, the `wikipedia_en_10_JS_trusted_addModules_2025-08.zim` is causing a significant problem to my kiwix-serve / kiwix-apple readers. Both simply fails to load the ZIM main article (or any other article in kiwix-serve). Surprisingly I rebuild a ZIM with same command and few other articles and it works fine. Don't know what is broken, but something is.

All javascripts version is obviously causing problems in the web console ... but nothing really visible in the articles.

The "trusted javascript" works very well. 

I'm not sufficiently export to assess the list of trusted modules, or the one of "addModules" version, but this is more a detail we can refine over the years anyway.

The only things I can say is that I confirm that the "trusted javascript with additional modules" is required/enough to solve #1491 and #1911 articles. Only "trusted javascript" is not enough. 

I've tested with JS disabled in Chrome and I confirm the ZIM degrades nicely (i.e. it looks like we get the same content than currently without this PR merged).

All points above makes me consider that we might want to change default scraper behavior (compared to current PR behavior), so that scraper creates "near-perfect" ZIM for "newbies", without needing additional flags. Especially important for people willing to ZIM wikipedia which is somehow our primary audience:
- we should probably have a `--nojavascript` option instead of a `--javascript` option, for those willing to not have any JS (even if ZIM with JS seems to degrade nicely in browsers with JS disabled, pretty sure some folk would prefer to not have any JS code in their ZIMs)
- we should maybe add more scripts to the trusted list

@Markus-Rost Thanks a lot anyway, very impressive change. Glad you made it work! I did not had any look at the code since you specified it is only a PoC. Let me know when it is worth it.

Markus-Rost:
> For some reasons, the `wikipedia_en_10_JS_trusted_addModules_2025-08.zim` is causing a significant problem to my kiwix-serve / kiwix-apple readers. Both simply fails to load the ZIM main article (or any other article in kiwix-serve). Surprisingly I rebuild a ZIM with same command and few other articles and it works fine. Don't know what is broken, but something is.

Oh, I'm getting that as well. Not sure what happened there. I tested it locally using 100.tsv which worked fine and created the 10.tsv version to fit the GitHub upload limit, but didn't actually test that one again.

> The only things I can say is that I confirm that the "trusted javascript with additional modules" is required/enough to solve #1491 and #1911 articles. Only "trusted javascript" is not enough.

Yes, those tables need the `site` module.

> All points above makes me consider that we might want to change default scraper behavior (compared to current PR behavior), so that scraper creates "near-perfect" ZIM for "newbies", without needing additional flags. Especially important for people willing to ZIM wikipedia which is somehow our primary audience:

Do you mean having all JS included by default? That doesn't seem like a good idea as some JS not expecting a request to fail can easily break stuff on the page.

> * we should probably have a `--nojavascript` option instead of a `--javascript` option, for those willing to not have any JS (even if ZIM with JS seems to degrade nicely in browsers with JS disabled, pretty sure some folk would prefer to not have any JS code in their ZIMs)

I've already added `--javaScript=none`, does no js really need it's own option?

> * we should maybe add more scripts to the trusted list

I've looked through basically all the modules I could find loaded on Wikipedia articles as well as a few other wikis and those are all the modules which won't try to load any non-scraped content. Sadly all three MathJax extensions either load external JS or from an extension path we can't easily rewrite or guess ahead of time. @benoit74 let me know about any specific scripts to add you have in mind and I'll take a look at them.

benoit74:
OK, then what I meant is having `--javascript=trusted` as default. If we have `--javascript=none`, this is OK.

And what I meant by having more modules was having everything needed for wikipedia (having the ancestry table in mind). But if this comes from the `site` modules, and since (correct me if I'm wrong) we could have anything in the `site` module, then we can't do that. Or maybe we could add the `site` modules automatically only for wikimedia projects... not sure it is such a good idea.

Markus-Rost:
> OK, then what I meant is having `--javascript=trusted` as default. If we have `--javascript=none`, this is OK.

`--javaScript=trusted` is already the default in this PR 

> And what I meant by having more modules was having everything needed for wikipedia (having the ancestry table in mind). But if this comes from the `site` modules, and since (correct me if I'm wrong) we could have anything in the `site` module, then we can't do that. Or maybe we could add the `site` modules automatically only for wikimedia projects... not sure it is such a good idea.

Yeah, the `site` module can contain anything and change at any time. We also can't just include it for all Wikimedia projects, as on some wikis it might load external content ([like on es.wikipedia.org](https://es.wikipedia.org/wiki/MediaWiki:Common.js#L-98--L-119)).

Markus-Rost:
> @Markus-Rost Thanks a lot anyway, very impressive change. Glad you made it work! I did not had any look at the code since you specified it is only a PoC. Let me know when it is worth it.

@benoit74 Oh please take a look at the code. I assume that I've broken all renderers besides ActionParse, though I'm also not familiar at all with those renderers and have no idea how badly broken they are. I'm certainly gonna need some help with getting those issues fixed.

Jaifroid:
Apologies for the delay. I've now tested these archives in Kiwix JS Browser Extension, with the same results as @benoit74 reported. 

* The JS_all shows errors in console relating to failures to fetch some scripts (but I'm not actually seeing blocked calls to external resources).
* JS_trusted shows a bunch of failures on the landing page, but not on other pages I tested.
* JS_none also shows failures on the landing page (clearly references to modules not included in the archive have not been removed), and also on article pages a failure to find `C/_mw_/startup.js`.
* JS_trusted_addModules fails to find the landing page, but article pages are fine.

I don't see any differences in formatting, suggesting to me that at least for the tested pages JS_none is as good as other formats, and would be faster to load on older devices due to fewer resources needing to be extracted. However, it would be necessary to remove references to scripts not in the archive to avoid the multiple console errors from attempts to load these. However, JS_trusted seems like a good compromise to me, so long as calls to untrusted resources are removed from the landing page.

Regarding the core issue of calling external scripts or resources #2310, it seems none of the test pages I noticed had that problem in any of these test archives, but possibly I didn't test deeply enough.

benoit74:
> Regarding the core issue of calling external scripts or resources https://github.com/openzim/mwoffliner/issues/2310, it seems none of the test pages I noticed had that problem in any of these test archives, but possibly I didn't test deeply enough.

Currently the scraper has indeed more or less solved #2310 by removing all JS, so that we de-facto do not load anymore external scripts. But this is not really a long term solution. This is why we kept #2310 open saying this was more an interim fix and the medium term solution is this PR, adding back "just enough" scripts.

Markus-Rost:
> * The JS_all shows errors in console relating to failures to fetch some scripts (but I'm not actually seeing blocked calls to external resources).

> Regarding the core issue of calling external scripts or resources #2310, it seems none of the test pages I noticed had that problem in any of these test archives, but possibly I didn't test deeply enough.

The calling of external scripts seen in #2310 is mostly just the attempted loading of JS modules. By rewriting startup.js to load those modules from `C/_mw_/MODULE.js`, they are now the failures to fetch some scripts you noticed in JS_all (though a lot of the called scripts are included in the ZIM). 

Of course JS_all can still have scripts attempting to load external resources, just that en.wikipedia is a bad example for that because they have very little JS doing that in the first place. Getting JS_all on es.wikipedia for example will load external scripts through the `site` module as mentioned in an earlier comment.

> * JS_trusted shows a bunch of failures on the landing page, but not on other pages I tested.
> * JS_none also shows failures on the landing page (clearly references to modules not included in the archive have not been removed), and also on article pages a failure to find `C/_mw_/startup.js`.
> * JS_trusted_addModules fails to find the landing page, but article pages are fine.

I have the suspicion that my filtering of JS modules stopped the already existing landing page JS from being included 😅 Ah, I'll fix JS_none including startup.

> I don't see any differences in formatting, suggesting to me that at least for the tested pages JS_none is as good as other formats, and would be faster to load on older devices due to fewer resources needing to be extracted. However, it would be necessary to remove references to scripts not in the archive to avoid the multiple console errors from attempts to load these. However, JS_trusted seems like a good compromise to me, so long as calls to untrusted resources are removed from the landing page.

You should notice small things like the navboxes at the bottom of articles being collapsed by default now or the video/audio player popup on "Michael Jackson", but yes there aren't a lot of differences in formatting as en.wikipedia does not rely as heavily on JS as other wikis might.

res/templates/javaScript.html
```diff
@@ -0,0 +1,7 @@
+<script>
```
benoit74: Inline scripts is taboo in ZIMs. Experienced showed it could cause problem on some readers around security constraints. It seems to be more or less resolved, but given the price to pay is not so high, we should transform this into JS scripts of their own.

Markus-Rost: The content of the script is different for every article, so that will be one script file for each article in the zim.

benoit74: Didn't realized that ... even if you already explained it in the past ... sorry. 

This is indeed going to a be a bit "painful" (more a technical constraint anyway than a real usability problem).

@kelson42 @Jaifroid I would appreciate your experience on that point. 

Which alternative is best:
- have few lines of inline JS in all articles (and only 1 HTML file per article)
- or have 1 HTML and 1 JS file per article but no inline JS at all (more or less doubling the number of entries in the ZIM)

benoit74: @kelson42 ping

benoit74: We've discussed about this topic with @kelson42 and he confirmed it feels way better to avoid inline JS at all prices, especially for very popular ZIMs like the ones made by mwoffliner.

He however came up with a new idea which could avoid to have 1 HTML + 1 JS file per article. It lookt like it could be possible to include variables we currently add in the JS (RLCONF, RLSTATE, RLPAGEMODULES) inside the HTML instead (with some "hidden" DOM properties, e.g `<meta name="RLCONF" content="..." />`, but I'm open to other ideas) and load these variables inside the appropriate JS object at page load. With this "workaround" we will need the same JS code on all articles, hence only one JS file per ZIM and no more inline JS.

@Markus-Rost WDYT, do you think this would be feasible?

Markus-Rost: Having the values in the DOM would technically be possible, but it seems like a very ugly work-around considering that embedding JSON data (like those variables are) is correctly done using `<script type="whatever">` tags as well. The `startup` module depends on those variables, so I would have to prepend the loading of the variables to the startup module.

However I'm not sure this is really worth it considering everything using the CSP to disallow inline scripts will most likely also disallow eval. That means the ES6 feature check (`new Function( '(a = 0) => a' );`) in the `startup` module will always fail and no JavaScript modules will be loaded. So everything which can't have inline scripts would fall back to running without JavaScript anyway. This feature check was added in MediaWiki 1.41 (released end of 2023) so this would be an issue for most MediaWiki sites.

Jaifroid: Not that I recommend it, but if it's essential, it is possible to allow `eval` with the `unsafe-eval` option in the CSP meta tag. while still keeping the block on external content.

Markus-Rost: I don't think we should compromise the CSP of the reader, that sounds like something that will just create more unintended side effects.

Jaifroid: I agree! I was more referring to any CSP that you might want to add to Wikipedia articles in the ZIM (for readers that don't enforce one). Indeed, a CSP can't in fact be loosened by a second CSP, it can only be made stricter.

benoit74: I think that @Markus-Rost point is correct: we probably shouldn't mind about using `inline` JS in mwoffliner ZIMs because we have a nice no-JS fallback present in the ZIM anyway. When the browser context does not allows to load inline JS, then JS will anyway not load at all and we are back to the no-JS fallback (which is not a dirty fallback at all since this is what is currently used on all production ZIMs).

I just tested test ZIMs from https://github.com/openzim/mwoffliner/issues/2310#issuecomment-3198240947 on Kiwix JS browser extension in all modes (restricted, with/without service worker, ...) and it looked fine. When in restricted mode, JS did not loaded but display was still fine. Of course there was some errors in the console, but I don't think we should really care. @Jaifroid do you confirm?

kelson42: If inline JS is now allowed, please open a dedicated issue in `openzim/overview` to manage this, because this is everything but clear why now it should be OK, whereas it has been forbidden during years. We can not change the rules, just because it is convenient to get a PR merged - as it.

Jaifroid: @benoit74 On my side, I had to develop a few years ago workarounds for lack of inline JS support in some contexts, notably Chromium extension (even in ServiceWorker mode). In such contexts the user is effectively redirected to a PWA from the extension (where there is no such restriction). Restricted mode is now effectively a no-JS mode that is kept for compatibility with old systems and as a security/screening measure when opening a new untrusted ZIM for the first time.

This should be an Org decision (@kelson42 ). There is absolutely no need to hold things up due to Kiwix JS or the PWA any more. There are independent arguments for avoiding inline JS in ZIMs where we know that JS is not needed, and for trying to keep our baseline, bread-and-butter Wikipedia ZIMs as simple as possible and compatible with even the oldest frameworks. With a bit of work, any inline script snippet can be converted into an attached script loaded at runtime (allowed by all frameworks of course), especially when we know the contents of inline JS script tags in a given context. Nevertheless, it seems reasonable to allow ZIM creators to take that decision themselves, if they need inline JS for their Mediawiki ZIMs, but for us to act conservatively vis-à-vis official Kiwix Wikipedia ZIMs.

As you say, Wikipedia ZIMs provide a good NOSCRIPT default in any case for now.. It should be possible to provide the long-awaited ZIM JS API #87 independently of whether or not inline JS is allowed.

Markus-Rost: > If inline JS is now allowed, please open a dedicated issue in `openzim/overview` to manage this, because this is everything but clear why now it should be OK, whereas it has been forbidden during years. We can not change the rules, just because it is convenient to get a PR merged - as it.

@kelson42 As far as I can tell, the reason for inline scripts to be disallowed is that some readers are/were blocking inline scripts from executing, resulting in (partially) unusable ZIMs. At least that was the case in all issues regarding inline scripts I could find (#324, #1578, #2096, kiwix/kiwix-js#355, kiwix/kiwix-js#789, kiwix/kiwix-js#865, openzim/overview#52), please let me know if there are additional reasons against inline scripts that I missed.

So avoiding inline scripts was to ensure anything needing those inline scripts would work properly in all ZIM readers. Now in this specific case here, avoiding inline scripts won't actually make the feature work in those readers as they are also blocking the eval based feature check in the `startup` JS and thus will always fall back to a noscript behavior. This means we have nothing to gain here from avoiding inline scripts.

See also openzim/overview#52 for a similar situation, through with a lot less graceful noscript behavior.

benoit74:
@kelson42 we need your decision on this one as well, it sat idle for way too long. 

I again see no reason to be worried about the inline scripts, because as stated by @Markus-Rost the ZIM will anyway fallback to no-JS even if we remove these inline scripts (because startup.js also requires `eval` which is blocked in same context than inline scripts). 

I tried the [wikipedia_en_10_JS_trusted_2025-08.zip](https://github.com/user-attachments/files/21842670/wikipedia_en_10_JS_trusted_2025-08.zip) ZIM in kiwix-desktop (it was your main concern) and it works like a charm, including all JS-based functions (like show/hide at the bottom of "Michael Jackson" page).

Taking measures to remove these inline scripts without providing any added value to the end users looks like a waste of our precious resources and a risk to introduce bugs.

The no-JS fallback is working exactly as-of today (without this PR merged), i.e. the ZIM will still be high-quality enough.

I understand you would prefer to provide same ZIM behavior on all readers, but at some point I feel like it is important to recognize that the effort is not worth it.

I've also opened https://github.com/openzim/overview/issues/62 to discuss the general policy for openZIM

benoit74:
I also tried the [wikipedia_en_10_JS_trusted_2025-08.zip](https://github.com/user-attachments/files/21842670/wikipedia_en_10_JS_trusted_2025-08.zip) ZIM in kiwix-js and kiwix-pwa in Restricted mode and I confirm it fallbacks very nice to no-JS mode (i.e. the ZIM looks just like a currently prod ZIM not supporting JS at all).

Jaifroid:
> I also tried the [wikipedia_en_10_JS_trusted_2025-08.zip](https://github.com/user-attachments/files/21842670/wikipedia_en_10_JS_trusted_2025-08.zip) ZIM in kiwix-js and kiwix-pwa in Restricted mode and I confirm it fallbacks very nice to no-JS mode (i.e. the ZIM looks just like a currently prod ZIM not supporting JS at all).

That one has a bunch of violations in ServiceWorkerLocal mode in Chrome, with the CSP violations looking pretty rough:

image

But more worryingly, in ServiceWorker mode, we get a bunch of attempts to access an external PHP server (wikipedia.org/w/load.php). At the very least those calls should be removed ISTM! -

image

Markus-Rost:
> > I also tried the [wikipedia_en_10_JS_trusted_2025-08.zip](https://github.com/user-attachments/files/21842670/wikipedia_en_10_JS_trusted_2025-08.zip) ZIM in kiwix-js and kiwix-pwa in Restricted mode and I confirm it fallbacks very nice to no-JS mode (i.e. the ZIM looks just like a currently prod ZIM not supporting JS at all).
> 
> That one has a bunch of violations in ServiceWorkerLocal mode in Chrome, with the CSP violations looking pretty rough:
> 
> image

I'm only seeing the two CSP violations from the two inline scripts existing. These are expected (and can be reduced to just one).

> But more worryingly, in ServiceWorker mode, we get a bunch of attempts to access an external PHP server (wikipedia.org/w/load.php). At the very least those calls should be removed ISTM! -
> 
> image

This is worrying, as only the startup module does those requests and has been modified in this PR to make JS work properly in the first place. The only way I can think of for you to get these errors in the PR ZIM again is if your service worker is somehow using a startup.js file cached from a different Wikipedia ZIM, which seems like a major security issue. @Jaifroid

Markus-Rost:
> But more worryingly, in ServiceWorker mode, we get a bunch of attempts to access an external PHP server (wikipedia.org/w/load.php). At the very least those calls should be removed ISTM! -
> 
> image

Oh, I found where this is coming from. It's the `sourceMappingURL` for modules loaded from local storage. Should be fairly easy to just remove that annotation, though the source map is only loaded by the browser dev tools anyway.

Jaifroid:
> This is worrying, as only the startup module does those requests and has been modified in this PR to make JS work properly in the first place. The only way I can think of for you to get these errors in the PR ZIM again is if your service worker is somehow using a startup.js file cached from a different Wikipedia ZIM, which seems like a major security issue. @Jaifroid

@Markus-Rost Just to confirm that I only do these tests in the Browser Extension, which takes a purist approach and only uses the resources supplied in the ZIM except (currently) for an added dark CSS if the user selects that (to be changed soon to native dark). (Even in the PWA, I never supply JS, only CSS for some other transforms.)

Markus-Rost:
I'm having less time to work on this now, but I would like to avoid this PR becoming stale. @benoit74 can you review the PR further (besides the inline script existing), so that I have some pointers to work on whenever I'm available?

src/util/misc.ts
```diff
@@ -210,7 +210,7 @@ export function cssPath(css: string, subDirectory = '') {
 }
 export function jsPath(js: string, subDirectory = '') {
   const path = isNodeModule(js) ? normalizeModule(js) : js
-  return `${subDirectory ? `${subDirectory}/` : ''}${path.replace(/(\.js)?$/, '')}.js`
+  return `${subDirectory ? `${subDirectory}/` : ''}${path}.js`
```
benoit74: This looks like the culprit breaking all e2e tests. I don't get the rationale behind this change, I feel like previous code was pretty harmless even for this PR changes. Could you explain / advise?

Markus-Rost: Some module names end in `.js`
Anything using those will break, unless we duplicate the replace on multiple different occasions.
Additionally, this creates a possible name conflict if a wiki has modules named both `name` and `name.js`.

Accepting the occurrence of some modules possibly being stored as `name.js.js` avoids all these issues for essentially no cost at all.

benoit74: The problem with this change is that we need to have the `.js` extension in static files list because this list contains a bunch of different files (CSS, JS, ...).

I see two ways forward: 
- modify "legacy" code to remove `.js` extension before calling `jsPath`
- modify "this PR" code to always add `.js` extension before calling `jsPath` so that we can remove it safely (once, should the module contain `.js` as well)

WDYT?

I tend to prefer second option because it ensures we have better chances of not breaking legacy things since they are kept untouched and because it avoids creating a discrepancy between `cssPath`, `anyPath` (which do remove the extension) and `jsPath` (which would not remove it).

Or maybe it is a good opportunity to rewrite all this code which indeed does look a bit fragile, but this is beyond this PR scope.

Markus-Rost:
> What about `DO_PROPAGATION`, `ALL_READY_FUNCTION` and `LOAD_PHP`? Looks like these were "useful" things at some point, but indeed never called since years since the module name was not correct so `startup.js` was never fixed.

I'm unsure what `DO_PROPAGATION` was supposed to solve. It removed a short timeout before each modules got loaded. I suspect the reason was to avoid using `mw.requestIdleCallback` because the `mw` class wasn't properly initialized before. Regarding the reason of the timeout, the MW script has a comment about it:
```
 		// Yield for two reasons:
		// * Allow successive calls to mw.loader.impl() from the same
		//   load.php response, or from the same asyncEval() to be in the
		//   propagation batch.
		// * Allow the browser to breathe between the reception of
		//   module source code and the execution of it.
		//
		// Use a high priority because the user may be waiting for interactions
		// to start being possible. But, first provide a moment (up to 'timeout')
		// for native input event handling (e.g. scrolling/typing/clicking).
		mw.requestIdleCallback( doPropagation, { timeout: 1 } );
```

`ALL_READY_FUNCTION` just disabled the check if all dependency modules are loaded. Again likely existed because the resource loader didn't work fully, but is no longer needed as the resource loader now works as intended inside the ZIM.

No idea what `LOAD_PHP` was intended to do, it does not match anything on any recent startup module (besides possibly nuking half the minified module code due to an overly agressive regex)

src/util/misc.ts
Markus-Rost: > The problem with this change is that we need to have the `.js` extension in static files list because this list contains a bunch of different files (CSS, JS, ...).

I think you are reading the diff wrong. The files will still always end in `.js`, just some files will end in `.js.js` because the module name itself ends in `.js`.

I'm honestly not sure how this change could break anything because it only removes an unnecessary modification of some js path names.

Markus-Rost: This change won't even effect 90% of modules, including those throwing errors in the e2e test. So the test errors must be coming from somewhere else.

benoit74: What I do know is that when I zimdump a ZIM generated by this PR, I have a file at `_res_/article_list_home.js.js` in the ZIM instead of `_res_/article_list_home.js`. Same for `_res_/images_loaded.min.js.js`, `_res_/masonry.min.js.js` and `_res_/script.js.js`. And I'm sure this is what causes many CI e2e errors (zimcheck complains that there is no file at `_res_/article_list_home.js` ... which is true)

benoit74:
OK then this PR should completely remove `DO_PROPAGATION`, `ALL_READY_FUNCTION` and `LOAD_PHP` for the time being, no need to keep in codebase stuff we do not really understand and does not seem to provide much added value.

src/util/misc.ts
Markus-Rost: Ah, that explains the broken main pages as well. Indeed the easiest solution would be to just always add the `.js` to module names then. Although adding `.js` to then remove it again feels like an ugly solution. Looks like `cssPath` does not need something like that.

benoit74: `cssPath` uses `anyPath` which in turn automatically removes the `.css` extension if present. So it has the same flaw / problem. I agree it is a bit ugly ... but I'm not sure to have the courage to fix all this ^^

Markus-Rost: Reverted

Markus-Rost:
Stuff left to do for this PR:

1. Write tests for the new parameters.
2. Figure out what to do with `jsConfigVars`, it's a custom JS module shared across all articles for some renderers, but for this PR I need it as a key-value object specific to each article. (Also did this ever work properly for the renderers having it shared across all articles? It's being pulled from the html of individual articles so some config variables are most likely page specific.)
4. Make `webpHandler` work again, it's the only custom module in ActionParse renderer so I just commented it out for now: https://github.com/openzim/mwoffliner/blob/51c5d466b08565e4168aa6fad455d3465288ba3b/src/renderers/action-parse.renderer.ts#L60-L62

benoit74:
I'm not sure at all that `jsConfigVars` ever worked correctly in other renderers. Note that we plan to also implement https://github.com/openzim/mwoffliner/issues/2210 for this same release, so looks like you should probably not be too ashamed of making things incompatible with other renderers.

benoit74:
@Markus-Rost I'm currently in holidays till the end of the month ; I will review it asap early February

Markus-Rost:
> I feel like we are now good to merge this.
> 
> We just need to open two new issues to not forget that we might want to:
> 
> * add back `webphandler` (to be confirmed in 2.0.0, I'm not that certain it is required anymore in 2026, maybe better to remove now)

I managed to keep it working as is in this PR. Can't comment on if it is generally needed anymore by now or not.

> * add proper support of `jsConfigVars` ; I agree it might not have really worked, and there is probably something like 99% chance it is currently broken in released versions ; I don't feel like this is something for 2.0.0, more something to keep track of

This PR adds proper support for `jsConfigVars` in ActionParse renderer (having them per-page based) as that's needed for the JS to work properly. 

The probably broken behavior is only in the other renderers, which as you previously mentioned might not be an issue anymore with 2.0.0 if #2210 gets implemented.

benoit74:
Perfect, then we are "simply" good to merge from my perspective

@kelson42 we need your feedback on this one ASAP now. Without constructive feedback on what still needs to be enhanced by the end of the week, I will merge.

kelson42:
@benoit74 @Markus-Rost Even if not participating I'm glad this moves forward. Will come back on this tomorrow morning.

kelson42:
@Markus-Rost @benoit74 Thank you very much for all the hard work and your patience. I think this PR does, in general, the right things but is stays pretty "experimental" and will probably need quite a few refinements. We should therefore merge, release and see, how it behaves with the hundreds of different wikis we scrape.

Glad we merge, but CI should pass first. Fixing broken/unstable CI should be a priority for this repository.

benoit74:
Rebased on top of new `main` commit + added new flags to `offliner-definition.json` so that we have access to them in the Zimfarm.

benoit74:
(sorry, I fucked-up first rebase, my local branch was totally out-of-date)

benoit74:
@Markus-Rost can you please quickly double-check I did not fucked-up the rebase twice and agree that I can proceed with the merge?

Markus-Rost:
> @Markus-Rost can you please quickly double-check I did not fucked-up the rebase twice and agree that I can proceed with the merge?

I cleaned up some now obsolete code, specially in regards to jsConfigVars. Should be good to merge now from my side.

benoit74:
Thank you very much again @Markus-Rost, so glad I can finally press the "merge" button on this long awaited PR. And very sorry it took so long.

PR #2484, opened 2025-08-19
Pin Node version to 24.5 until Jest issue is solved

benoit74:
Node.JS 24.6 introduced a regression in Jest

See https://github.com/nodejs/node/issues/59480

Until this upstream issue is solved, let's use Node.JS 24.5


benoit74:
@Markus-Rost FYI, you can rebase all your PRs experiencing errors like to Node 24.6, this should fix them

PR #2496, opened 2025-08-23
Don't reserve space for elements that don't exist

### LINKED ISSUE #2495: Don't reserve space for elements that don't exist
The vector-2022 skin reserves space for a TOC and appearance sidebar on the right and left of the content area, this means the content is not centered in our ZIMs. We should stop reserving the space for those elements which aren't part of our skin anyway.

Markus-Rost:
Fix #2495


benoit74 (approved):
LGTM

PR #2501, opened 2025-08-26
Fix logging + documentation around `RedisKvs.iterateItems`

### LINKED ISSUE #2500: `RedisKvs.iterateItems` return number of running workers, not worker ID
`RedisKvs.iterateItems` function is iterating over of Redis Hash and calls a given function for every list of items it grabs from the Hash with the `SCAN` operation.

Code is at `src/util/RedisKvs.ts`:

https://github.com/openzim/mwoffliner/blob/585a134914e5bd392ec19f5dee252d23e17cb596/src/util/RedisKvs.ts#L82

The function which is called by `RedisKvs.iterateItems` is passed two parameters:
- the list of items it grabbed in `SCAN` operation: `items`
- the amount of currently running workers: `runningWorkers`

The type signature in `src/types.d.ts` is however a bit blurry by using `run` name instead of `runningWorkers`:

https://github.com/openzim/mwoffliner/blob/585a134914e5bd392ec19f5dee252d23e17cb596/src/types.d.ts#L70

The two places where scraper uses this `runningWorkers` parameter is misleadingly treating it as `workerId`:

https://github.com/openzim/mwoffliner/blob/585a134914e5bd392ec19f5dee252d23e17cb596/src/util/categories.ts#L12

https://github.com/openzim/mwoffliner/blob/585a134914e5bd392ec19f5dee252d23e17cb596/src/util/saveArticles.ts#L30

https://github.com/openzim/mwoffliner/blob/585a134914e5bd392ec19f5dee252d23e17cb596/src/util/saveArticles.ts#L281

This does not impact scraper behavior but is clearly misleading the person reading logs / codebase.

It would help as well to have more documentation in `src/util/RedisKvs.ts` about what is function is doing (it is at least the third time I scratches my head trying to understand what `iterateItems` is doing and what are those batches of items and how it distributes items over workers).

benoit74:
Fix #2500 

Changes:
- fix documentation
- fix logging
- remove unneeded `async` in promise executor

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -30,6 +30,7 @@ Unreleased:
 * FIX: Rewrite non-Parsoid links properly (@Markus-Rost #2479)
 * FIX: Don't reserve space for elements that don't exist (@Markus-Rost #2495)
 * FIX: Get the whole article when the specified section is not available (@Markus-Rost #2480)
+* FIX: Fix logging + documentation around `RedisKvs.iterateItems` (@benoit74 #2500)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -67,7 +67,7 @@ interface RKVS<T> {
   deleteMany: (prop: string[]) => Promise<number>
   keys: () => Promise<string[]>
   len: () => Promise<number>
-  iterateItems: (numWorkers: number, func: (items: KVS<T>, run: number) => Promise<void>) => Promise<void>
+  iterateItems: (numWorkers: number, func: (items: KVS<T>, runningWorkers: number) => Promise<void>) => Promise<void>
   scan: (cursor: number) => Promise<{ cursor: number; items: KVS<T> }>
   flush: () => Promise<number>
 }
--- a/src/util/RedisKvs.ts
+++ b/src/util/RedisKvs.ts
@@ -79,9 +79,21 @@ export default class RedisKvs<T> implements RKVS<T> {
     return this.redisClient.hLen(this.dbName)
   }
 
+  /**
+   * Iteratively call function passed with batches of items from the underlying KVS hash, with a given number of workers
+   * working in parallel.
+   *
+   * This function scans the underlying KVS hash for items to batch. Batch size is "around" 10 items (Redis does
+   * not garantee exact batch size for SCAN operation). For every batch retrieved, passed "func" is called with the
+   * "items" retrieved.
+   *
+   * "numWorkers" are started in parallel, each processing a batch of items. Number of "runningWorkers" is passed to
+   * "func" called, mostly useful for logging / debugging purposes.
+   *
+   * This function returns when all items have been batched and passed to processing function.
+   */
   public iterateItems(numWorkers: number, func: (items: KVS<T>, runningWorkers: number) => Promise<void>): Promise<void> {
-    // eslint-disable-next-line no-async-promise-executor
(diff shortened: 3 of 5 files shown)


benoit74:
The correct name is `runningWorkers` because this is what is computed / passed in function: https://github.com/openzim/mwoffliner/blob/6568693185f79e3e0863f4d0f9770ffbb372edfd/src/util/RedisKvs.ts#L82

Previous logs produced something like this when numWorkers is 3:

> Worker [1] processing batch of [50] files
Worker [2] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [2] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [3] processing batch of [50] files
Worker [2] processing batch of [50] files
Worker [2] processing batch of [48] files

Anyone can see this problem with official Docker image with something like:

```
docker run --rm --name mwoffliner_test ghcr.io/openzim/mwoffliner:latest mwoffliner --adminEmail=contact@kiwix.org --articleList=https://download.openzim.org/wp1/enwiki_2025-03/tops/100.tsv --customZimDescription="Test" --customZimTitle=Test --filenamePrefix=tests_en_mwoffliner --format=maxi --mwUrl=https://en.wikipedia.org --outputDirectory=/output --publisher=openZIM --verbose=info --webp
```

If you focus on the step downloading files, you will see that you never get numbers 1, 2, 3, 4 again after the initial batch of 10 files.

And while it might look compelling, this is not the "reality". In fact the `1`, `2`, `3` are the number of running workers. There is not such things as a workerId in the util function managing it, it just starts jobs until there is a many jobs running than expected number of workers. 

The only exception is https://github.com/openzim/mwoffliner/blob/6568693185f79e3e0863f4d0f9770ffbb372edfd/src/util/mw-api.ts#L11 where the way we creates jobs to run are allowing to have something like a workerId. 


From my understanding, correct logs (representing what the real current situation is) should be:

> Worker processing batch of [50] files - 1 worker(s) running
 Worker processing batch of [50] files - 2 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 2 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 3 worker(s) running
 Worker processing batch of [50] files - 2 worker(s) running
 Worker processing batch of [50] files - 2 worker(s) running

Does this help understand the situation?

benoit74:
Do you feel like we should simply get rid of this value? I agree it does not seems to be a very useful value, especially since it might be outdated at any moment like you said. But I do not achieved to find the background of why it was introduced, so I wanted to at least fix documentation + naming + log so that at least we have a clearer and less wrong situation.

Markus-Rost:
Maybe turn it into a proper incremental worker id and add logs both at the start and end of the function. This way you can still see how many workers are currently active.

benoit74:
There is no such thing as a worker ID, there is just "up to `numWorkers` parallel tasks running. So it is mostly impossible to say that worker x started a given task. Plus on a personal basis I'm really not convinced it adds any value to the scraper, not even for debugging. Feel free to open an issue or a PR if you feel like you have something worth it.

I will merge this PR as-is, goal was anyway first and foremost to stop having confusing and plain wrong logs. Goal is achieved from my PoV.

PR #2503, opened 2025-08-28
Add redirects from all acceptable namespaces

### LINKED ISSUE #2429: Add more redirects, especially all redirects from article namespace
Currently, when scraper searched for article redirects, it searches for redirects only in namespace 0, no matter in which namespace the article itself is.

In general, this is fine because it avoids to bloat the ZIM with many unneeded redirects.

But we do have occurence where this causes some redirect to go missing, including very visible ones like in https://github.com/openzim/zim-requests/issues/777#issuecomment-3061562928

I think we should add redirects from namespace 0 and from article own namespace.

Looks like [mediawiki API](https://www.mediawiki.org/wiki/API:Redirects) does not support this filtering, so we should probably query API for all redirects, and then filter on our own. This would even allow to do even more intelligent filtering: including redirects from namespace 0 and from article own namespace when we are making a selection (`--articleList` is used) and from selected namespaces when we retrieve all articles from given namespaces.

benoit74:
Fix #2429

### DIFF
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -249,14 +249,17 @@ export function normalizeMwResponse(response: MwApiQueryResponse): QueryMwRet {
     }
     const articleId = String(id).replace(/ /g, '_')
     if (page.redirects) {
-      page.redirects = page.redirects.map((redirect) => {
-        // The API returns the redirect title (!?), we fake the
-        // redirectId by putting the underscore. That way we
-        // secure the URL rewritting works fine.
-        redirect.title = String(redirect.title).replace(/ /g, '_')
+      page.redirects = page.redirects
+        // drop redirects from talk (not subject) namespaces and from User namespace, except if namespace has been expressly requested
+        .filter((redirect) => (redirect.ns % 2 === 0 && redirect.ns !== 2) || MediaWiki.namespacesToMirror.some((ns) => MediaWiki.namespaces[ns].num === redirect.ns))
+        .map((redirect) => {
+          // The API returns the redirect title (!?), we fake the
+          // redirectId by putting the underscore. That way we
+          // secure the URL rewritting works fine.
+          redirect.title = String(redirect.title).replace(/ /g, '_')
 
-        return redirect
-      })
+          return redirect
+        })
     }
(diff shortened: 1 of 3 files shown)


src/util/mw-api.ts
```diff
@@ -249,14 +249,19 @@ export function normalizeMwResponse(response: MwApiQueryResponse): QueryMwRet {
...
-        // secure the URL rewritting works fine.
-        redirect.title = String(redirect.title).replace(/ /g, '_')
-
-        return redirect
-      })
+      page.redirects = page.redirects
+        // remove redirects in User namespace
+        .filter((redirect) => redirect.ns !== 2)
+        // keep only redirects from special namespaces, from subject (not talk) namespaces and from expressly requested namespaces
+        .filter((redirect) => redirect.ns < 0 || redirect.ns % 2 === 0 || MediaWiki.namespacesToMirror.some((ns) => MediaWiki.namespaces[ns].num === redirect.ns))
```
Markus-Rost: Why are we keeping redirects from special namespaces?

benoit74: I thought we might have some wikis where people are linking to a redirect in the special namespace for some reason. And I don't feel like it could introduce "bad" entries in the search result. 

My approach here was to keep all redirects besides the one we know for sure have a high chance of not been interesting: talk redirects (not sure it is even possible) + user namespace redirects.

But I might miss something, how do you see this being a problem?

Markus-Rost: Special pages don't show up in prop=redirects

Also the only special pages that show up in the api as redirects are user based. 
- Special:MyLanguage is used a lot for links on wikis with Extension:Translate, but as mentioned won't show up here.
- Special:MyPage / Special:MyTalk won't be used a lot, but can reveal the scraping user/ip. However they also won't show up here, only when directly querying or parsing as title.

benoit74: Let's remove special namespaces then, thank you!

benoit74: Or ignore them more exactly

Markus-Rost (approved):

Issue #2478, opened 2025-08-18
MWOffliner looking for x86 libjpeg.9.dylib on Apple Silicon


maxwell-martin:
maxwellmartin@Maxwells-MacBook-Air ~ % npm i -g mwoffliner
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated q@1.5.1: You or someone you depend on is using Q, the JavaScript Promise library that gave JavaScript developers strong feelings about promises. They can almost certainly migrate to the native JavaScript promise now. Thank you literally everyone for joining me in this bet against the odds. Be excellent to each other.
npm warn deprecated
npm warn deprecated (For a CapTP with native promises, see @endo/eventual-send and @endo/captp)
npm warn deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm warn deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm warn deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm warn deprecated swig-templates@2.0.3: unmaintained
npm error code 1
npm error path /opt/homebrew/lib/node_modules/mwoffliner/node_modules/jpegoptim-bin
npm error command failed
npm error command sh -c node lib/install.js
npm error compiling from source
npm error Command failed: /opt/homebrew/lib/node_modules/mwoffliner/node_modules/jpegoptim-bin/vendor/jpegoptim --version
npm error dyld[95676]: Library not loaded: /usr/local/opt/jpeg/lib/libjpeg.9.dylib
npm error   Referenced from: <D098E391-1E09-3994-B7C3-D9F78412A72E> /opt/homebrew/lib/node_modules/mwoffliner/node_modules/jpegoptim-bin/vendor/jpegoptim
[...]

PR #2504, opened 2025-08-29
Update README with clearer npm installation steps for Apple Silicon

maxwell-martin:
Added clearer installation instructions when using npm on Apple Silicon devices to prevent the annoying "jpegoptim pre-build test failed" error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,16 @@ npm i -g mwoffliner
 how your `npm` / OS is configured. `npm` permission checking can be a bit annoying for a
 newcomer. Please read the documentation carefully if you hit problems: https://docs.npmjs.com/cli/v7/using-npm/scripts#user
 
+> [!WARNING]
+> Installing with `npm` on Apple Silicon devices specifically requires the following steps:
+```
+brew install jpeg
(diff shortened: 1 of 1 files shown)


benoit74 (changes requested):
- The fact that `libjpeg` must be installed is generic to all platforms ; adding instructions only for Mac is kinda embarrassing ; and it is not something needing a warning, it is just a prerequisite
- The problem is is also not specific to Apple Silicon, I did not tested it (yet) but I'm almost 99% sure it applies to all Apple machines
- Both points above make me sure we must not formulate it like this is currently done in this PR, try to find something clearer / less misleading
- See https://github.com/openzim/mwoffliner/issues/2478#issuecomment-3237879077 ; I don't want to add something on the README which is not necessary **and** could harm the user by modifying his `~/.zshrc` and causing conflicts with `jpeg-turbo`

maxwell-martin:
See my comment on #2478 to your most recent comment. It doesn't make sense that I have to export specific variables to various paths to install an npm package. I just thought that my changes would help users on Mac, specifically. You obviously know more about this than I do. But, clearly there was an issue with the instructions to the point that we got into this discussion in the first place?

PR #2505, opened 2025-08-31
Also retry `internal_api_error_MWException` errors

benoit74:
https://farm.openzim.org/pipeline/78de7699-9395-40ab-9845-7272f6a60555/ and previous runs of mdwiki.org (see https://github.com/openzim/zim-requests/issues/1260#issuecomment-3204128919 and https://github.com/openzim/zim-requests/issues/1260#issuecomment-3209373923) proved that we can have what looks like transient `internal_api_error_MWException` errors.

Retrying this error seems to be able to make it work, at least in all cases doing the query manually afterwards in a success.


benoit74:
Not waiting for review, this is a minor change. Feedback still welcomed if you feel like this is something silly.

Markus-Rost:
MWException is the generic error code.

PR #2509, opened 2025-09-01
Also retry `internal_api_error_RuntimeException` errors

benoit74:


Markus-Rost (commented):
Considering this list of errors to retry already includes the generic error code, it seems easier to just retry all errors by default at this point and switch to a list of error codes *not* to retry instead.

benoit74:
Original idea was that error codes referenced in `error.manager.ts` should not be retried. But I'm not convinced by that anymore.

I start to consider that we should probably retry all `internal_api_error_*` errors. 

This would come with the price of risking overwhelming a bit the upstream. For instance, `internal_api_error_Wikimedia\\RequestTimeout\\RequestTimeoutException` should maybe not be retried because they are probably overwhelming the upstream ... but at the same time they should be retried because second call might succeed because data will be in cache.

And it would slow down significantly the scraper when retrying due to backoff logic. Retry rate is kinda "slow" (exponential backoff starting at 1 sec with factor 2 and maximum 60 secs, retrying up to 10 times before giving up), which is good to not overwhelm the upstream but means that scraper will get slow when we have permanent failures handled by `error.manager.ts` ; but this is not expected to be a "normal" situation.

Note that we are already retrying 500, 502, 503 errors despite some of these being handled by `error.manager.ts`, because experienced showed these errors can be transient.

benoit74:
Closing this PR in favor of #2515

PR #2512, opened 2025-09-02
Check API availability at scraper startup instead of raw mwUrl which might not be reachable

### LINKED ISSUE #2447: MWoffliner request MediWiki unrelated URL although proper configuration
MWoffliner request wrong `https://zdoom.org/` but this URL is no served by Mediawiki and fails, see https://farm.openzim.org/pipeline/5300ed54-6287-42a3-b4ad-0502278e9a45/debug.

recipe is configure with a `mwWikiPath` like `/wiki/` so it should request `https://zdoom.org/wiki/`

Related to #2422

benoit74:
Fix #2447

Changes:
- so far, at scraper startup we've checked the raw `mwUrl` ; this is not really correct because we never use this raw URL
- with this PR, we check the `api.php` API reachability since this is supposed to be correct in all situations and has become the only real required API / setting (with `mwUrl`) since #2456
- changed a bit wording to not speak about validity but reachability ; the URL might be valid but the scraper fails to reach it for whatever reason and this "valid" word was confusing

### DIFF
--- a/src/sanitize-argument.ts
+++ b/src/sanitize-argument.ts
@@ -1,6 +1,6 @@
 import S3 from './S3.js'
 import RedisStore from './RedisStore.js'
-import { fileURLToPath } from 'url'
+import { fileURLToPath, URL } from 'url'
 import pathParser from 'path'
 import * as logger from './Logger.js'
 import fs from 'fs'
@@ -9,6 +9,7 @@ import * as path from 'path'
 import { parameterDescriptions } from './parameterList.js'
 import { RENDERERS_LIST } from './util/const.js'
 import Downloader from './Downloader.js'
+import MediaWiki from './MediaWiki.js'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
@@ -100,10 +101,10 @@ export async function sanitize_all(argv: any) {
 // perform live checks of arguments needing a connection "somewhere"
 export async function check_all(argv: any) {
   // extracting required arguments
-  const { mwUrl, customZimFavicon, optimisationCacheUrl, insecure } = argv
+  const { mwUrl, mwActionApiPath, customZimFavicon, optimisationCacheUrl, insecure } = argv
 
-  // check mwUrl availability
-  await check_mwUrl(mwUrl)
+  // check mediawiki API availability
+  await check_mwApiReachability(mwUrl, mwActionApiPath)
 
   // check s3 availability
   await check_s3(optimisationCacheUrl, insecure)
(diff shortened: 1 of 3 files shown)


src/sanitize-argument.ts
```diff
@@ -180,9 +181,10 @@ export function sanitize_speed(_speed: any) {
   }
 }
 
-export async function check_mwUrl(mwUrl: string) {
-  await Downloader.get(mwUrl).catch((err) => {
-    throw new Error(`mwUrl [${mwUrl}] is not valid:\n${err}`)
+export async function check_mwApiReachability(mwUrl: string, mwActionApiPath: string) {
+  const apiFullUrl = new URL(mwActionApiPath || MediaWiki.actionApiPath, mwUrl).toString()
+  await Downloader.get(apiFullUrl).catch((err) => {
+    throw new Error(`Mediawiki API URL [${apiFullUrl}] is not reachable:\n${err}`)
```
Markus-Rost: Not sure if this would be too much at this point, but we could hit `?action=query&format=json&formatversion=2` to ensure there is actually a MediaWiki Action API endpoint here.

Markus-Rost: If we do hit that query we will want to check for a valid JSON response

src/sanitize-argument.ts
```diff
@@ -180,9 +181,11 @@ export function sanitize_speed(_speed: any) {
   }
 }
 
-export async function check_mwUrl(mwUrl: string) {
-  await Downloader.get(mwUrl).catch((err) => {
-    throw new Error(`mwUrl [${mwUrl}] is not valid:\n${err}`)
+export async function check_mwApiReachability(mwUrl: string, mwActionApiPath: string) {
+  // We are building an API query "by-hand" because we've not yet initialized the whole URL director / URL builder stack
+  const apiQueryUrl = `${new URL(mwActionApiPath || MediaWiki.actionApiPath, mwUrl).toString()}?action=query&format=json&formatversion=2`
+  await Downloader.get(apiQueryUrl).catch((err) => {
```
Markus-Rost: We should confirm that the api query returns valid json.

benoit74: How would you do that? just check response header or do you want to check something else?

Markus-Rost:
```suggestion
  await Downloader.get(apiQueryUrl, Downloader.jsonRequestOptions).catch((err) => {
```
Just adding the JSON request options should be enough to have it throw an error if the body can't be parsed as JSON.

benoit74: Nope, when HTML is returned (for instance) this tries to interpret the HTML as JSON, and achieves to do so by assuming it is just a string.

I've instead added a check on returned content-type, should be enough and OK.

Issue #2513, opened 2025-09-02
Rename --keepEmptyParagraphs to --keepEmptySections


kelson42:
The current naming is really confusing for the users.

Don't forget to rename variables and any underlying functions/code accordingly.

kelson42:
@ziaddevv has prepared a PR for this issue.
@ziaddevv If you put a comment, I can assign this issue to you

ziaddevv:
> [@ziaddevv](https://github.com/ziaddevv) has prepared a PR for this issue.
> [@ziaddevv](https://github.com/ziaddevv) If you put a comment, I can assign this issue to you

Yes, You can assign it to me

Issue #2489, opened 2025-08-20
Remove mwWikiPath and mwIndexPhpPath options


Markus-Rost:
Remove `mwWikiPath` and `mwIndexPhpPath` options as they are pulled directly from the wiki since #2456

benoit74:
We should remove all deprecated CLI parameters in next major: at least `mwWikiPath`, `mwIndexPhpPath`, and `keepEmptyParagraphs` as of today

PR #2514, opened 2025-09-03
Rename --keepEmptyParagraphs to --keepEmptySections

ziaddevv:
- Updated cli parameter from --keepEmptyParagraphs to --keepEmptySections
- Updated all internal references in TypeScript source files

Fixes openzim/mwoffliner#2513


kelson42:
We will have to wait the release of version 1.17.0 before merging this PR.

src/renderers/abstract.renderer.ts
```diff
@@ -768,7 +768,7 @@ export abstract class Renderer {
     })
 
     /* Remove empty paragraphs */
```
benoit74: Please fix comment as well

benoit74 (changes requested):

benoit74:
> We will have to wait the release of version 1.17.0 before merging this PR.

I imagine you mean that we will have to wait for next major before merging this PR? Currently there is no plan for a `1.18.0`, but I can pretty easily imagine we might need to make a patch `1.17.1` to fix some urgent bugs, so I would prefer we do not merge changes requiring a major like this too fast

benoit74:
@ziaddevv please also add an entry in `CHANGELOG` file

ziaddevv:
> @ziaddevv please also add an entry in `CHANGELOG` file

Thank you for the feedback! I've updated the comment and added an entry in the ChangeLog file

benoit74 (approved):
LGTM. Moving to draft to not merge it by accident, to be merged only at next major version

test/unit/saveArticles.test.ts
```diff
@@ -199,7 +199,7 @@ describe('saveArticles', () => {
 
     test(`Keep empty paragraphs for ${renderer} renderer`, async () => {
```
Markus-Rost: Please update the description of the test as well.

Markus-Rost (changes requested):

test/unit/saveArticles.test.ts
ziaddevv: done , so the test just right above it have the same issue but it doesn't directly reference the new CLI behavior (keepEmptySections)  do i have to change it ?  test(`Remove empty paragraphs for ${renderer} renderer`, async () => {

ziaddevv: also what about the old changelogs ?

Markus-Rost: The test above is testing the default behavior, the description should be updated as well.

Old changelogs don't need to be changed.

kelson42: I we want to merge straight, we can also keep backward compat. to be removed later

Markus-Rost:
> I we want to merge straight, we can also keep backward compat. to be removed later

True, just deprecating the old argument name while adding the new one should be fairly simple:
1. Have both names in the parameter list, description of keepEmptyParagraphs suggesting the new name
2. Add `.deprecateOption('keepEmptyParagraphs', 'use --keepEmptySections')` to `cli.ts`
3. Have mwoffliner.lib.ts#L344 take `keepEmptySections: keepEmptySections ?? keepEmptyParagraphs`

kelson42:
> > I we want to merge straight, we can also keep backward compat. to be removed later
> 
> True, just deprecating the old argument name while adding the new one should be fairly simple:
> 
> 1. Have both names in the parameter list, description of keepEmptyParagraphs suggesting the new name
> 2. Add `.deprecateOption('keepEmptyParagraphs', 'use --keepEmptySections')` to `cli.ts`
> 3. Have mwoffliner.lib.ts#L344 take `keepEmptySections: keepEmptySections ?? keepEmptyParagraphs`

@ziaddevv Possible to do that?

ziaddevv:
yes i'll do it , i though the issue was just drafted

ziaddevv:
I kept the parameter descriptions the same  so that users running --help will still see the deprecated options, making the backward compatibility clear , Is that ok ?

kelson42:
> I kept the parameter descriptions the same so that users running --help will still see the deprecated options, making the backward compatibility clear , Is that ok ?

No, please move everything to the new naming. Just secure that if someone use the old option naming, then it continues to work fine.

ziaddevv:
no i just meant to say that in this step  `Have both names in the parameter list, description of keepEmptyParagraphs suggesting the new name ` I kept the description of keepEmptyParagraphs the same as KeepEmptySections and user can see the suggestion say ` [deprecated: use --keepEmptySections instead]`

benoit74:
Thank you! 

I cleaned this up (rebase was needed) and mentioned final deletion of CLI parameter in https://github.com/openzim/mwoffliner/issues/2489

I wanna merge this soon to release it in 1.17.1

benoit74:
unit-test is failing due to PR from fork missing secret to S3. Merging anyway

Issue #1136, opened 2020-05-24
Supports dark mode with color-scheme and prefers-color-scheme


automactic:
Zim files are essentially bundles of html content, to support dark mode, one approach is to use [CSS Color Schemes](https://www.w3.org/TR/css-color-adjust-1/). While still in draft, it's already [widely adopted](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). And I think this more standardized way of handling color scheme is better than rolling our own solution.

**Step 1**: let the browser know the page supports dark and light scheme:
```
:root {
  color-scheme: light dark;
}
```
or 
```
<meta name="color-scheme" content="light dark">
```

**Step 2**: adjust existing css for specific color scheme
```
@media (prefers-color-scheme: dark) {
    xxxx
}
```
I know this step will take up a lot of time and energy, but in most of the situations, the user agent style sheet handles things pretty well.

kelson42:
I really wonder why this is not handle properly like this upstream on Wikipedia.org.

Jaifroid:
Isn't there a proposal for an API in ZIMs that might handle this kind of transform? Currently the different readers handle dark mode by injecting CSS stylesheets. By using an inversion stylesheet and re-inverting certain image type, it's fairly easy to come up with a "universal" (though not perfect) dark mode, and that could be the API. (It's what most of the current readers do.)

Otherwise handling dark mode as a set of themes is something that each Wiki would have to provide, as they would need to be tailored to the different styles used.

Lucent:
Related #2086

kelson42:
After our discussion at https://github.com/openzim/overview/issues/53, we should really implement this.

@benoit74 Do you think it would be easy using approach suggested by @Lucent in #2086?

benoit74:
If I get it correctly, what @Lucent showed in #2086 is that the foundation decided to not resort on standard `prefers-color-scheme` attributes but rather implement their on additional CSS class on `<html>` tag. This means that our current strategy in https://github.com/openzim/overview/issues/53 is not going to easily operate correctly on mwoffliner ZIMs since it means we would have to (automatically) reverse engineer the whole CSS to transform from `.skin-theme-clientpref-night` to `@media (prefers-color-scheme: dark)` ; nothing infeasible, but not easy.

Not speaking about the fact that support for dark mode seems to be well behind the fact that we should be able to ZIM more Mediawikis (support more skins), fetch categories, ... from my PoV.

[...]

PR #2518, opened 2025-09-08
If vector2022 night mode is enabled, use the preferred color scheme

### LINKED ISSUE #2086: Support Vector skin and its dark/night mode
Current exports of Wikipedia using the Vector 2022 default theme support light and dark mode via toggling a single class on `<html>`, `skin-theme-clientpref-night`. You can see this class inserted even in current mwoffliner dumps, probably by header JS, but it has no effect, perhaps due to pulling CSS for the Minerva skin.

Will the new Vector skin eventually be supported or is there some way of forcing mwoffliner to pull Vector skin themed pages? This would make light/dark theming trivial and solve all related bugs.

I will sponsor a bounty of your choosing to get this done.

Markus-Rost:
Fix #2086
Fix partially #1136 (for wikis using vector-2022 like Wikipedia)

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -35,6 +35,7 @@ Unreleased:
 * FIX: Fix logging + documentation around `RedisKvs.iterateItems` (@benoit74 #2500)
 * FIX: Add redirects from all acceptable namespaces (@benoit74 #2429)
 * FIX: Check API availability at scraper startup instead of raw mwUrl which might not be reachable (@benoit74 #2447)
+* CHANGED: If vector2022 night mode is enabled, use the preferred color scheme (@Markus-Rost #2086)
 
 1.16.0:
 * CHANGED: ActionParse renderer is now the preferred one when available (@benoit74 #2183)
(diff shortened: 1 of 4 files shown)


Markus-Rost:
@benoit74 The missing styles are from the `ext.wikimediamessages.styles` module which is specific to only Wikimedia wikis. The upstream issue has been fixed with MediaWiki 1.45.0-wmf.18, which is being rolled out on Wikimedia wikis since Monday and will be live on all Wikimedia wikis (including Wikipedia) later today ([Roadmap](https://www.mediawiki.org/wiki/MediaWiki_1.45/Roadmap#18)). This PR is a draft until the patch is live on en.wikipedia.org later today.

With the styles issue being specific to Wikimedia wikis and the Wikimedia wikis using the latest MediaWiki version, there is no need for a legacy fix in this case.

> Second because it is mostly impossible here. I don't think we can know all modules we are missing due to upstream bug ; ext.wikimediamessages.styles is the most prominent one, but can't there be other modules?

There could be more modules missing, however those would be missing in the current light mode ZIM as well. `ext.wikimediamessages.styles` just went unnoticed before because it contains mainly darkmode styles for Wikipedia. As such other modules missing would be distinct issues and not directly related to the change here.

Markus-Rost:
@benoit74 the patch is live on all Wikimedia wikis, you should be seeing proper dark mode on enWP now.

kelson42:
Assuming I get it right: from now we have light/dark CSS themes on Wikimedia wikis and in ZIM files created with dev version of MWoffliner, bug kudos to @Markus-Rost and @benoit74 !!! I will test this and report bugs if necessary.

PR #2525, opened 2025-09-15
Retry all Mediawiki errors except few known ones deemed to fail

### LINKED ISSUE #2515: Change posture around retry logic when calling the API
Currently, when calling any Mediawiki API, the scraper:
- systematically retries all 429, 500, 502, 503, 504 errors 
- only retries response with `error.code` inside JSON when this code is among few "authorized" values encountered in the wild

Retrying currently means applying an exponential backoff of at least 1 sec, up to 60 secs, with at most 10 retries before giving up.

We are gradually adding more and more code to the list of `error.code` worth a retry. And it looks like the list might never end.

We recently added #2505, #2506 and are about to consider #2509 for instance. All those errors proved to be transient on mdwiki.org

It is also very frustrating to fail a scrape and have to wait for a scraper release, for something which might have been handled automatically.

The original idea of retrying only few codes was to avoid retrying permanent errors. But it looks like it is maybe better to retry an error which will finally but permanent and stop the scraper, rather than failing immediately for something which would have work on retry. At least it is way less deceptive for the end user.

My proposition is hence to retry all JSON response with an `error.code` value except if it contains what really looks like permanent or harmful errors:
- `missingtitle`
- `internal_api_error_MediaWiki\\Revision\\BadRevisionException`
- `internal_api_error_Wikimedia\\Assert\\UnreachableException`
- `internal_api_error_Wikimedia\\Assert\\InvariantException`
- `internal_api_error_Wikimedia\\Parsoid\\Core\\ResourceLimitExceededException`

This means we will retry these errors which will be replaced by a placeholder if permanent:
- `internal_api_error_Wikimedia\\RequestTimeout\\RequestTimeoutException` (might work on second try due to caching at various level)
- `internal_api_error_InvalidArgumentException`
- `internal_api_error_Error`
- `internal_api_error_TypeError`

@kelson42 I need your feedback on this issue before considering merging #2509, since both are moving in opposite direction ; implementing this issue looks like a requirement for proper mdwiki.org operation

benoit74:
Fix #2515

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -32,7 +32,7 @@ Unreleased:
 * FIX: Don't reserve space for elements that don't exist (@Markus-Rost #2495)
 * FIX: Get the whole article when the specified section is not available (@Markus-Rost #2480)
 * FIX: Retry ERR_CANCELED errors since this might be a transient problem (@benoit74 #2414)
-* FIX: Also retry `internal_api_error_MWException` and `internal_api_error_ThrottledError` errors (@benoit74 #2505 #2506)
+* FIX: Retry all Mediawiki errors except few known ones deemed to fail (@benoit74 #2505 #2506 #2515)
 * FIX: Fix logging + documentation around `RedisKvs.iterateItems` (@benoit74 #2500)
 * FIX: Add redirects from all acceptable namespaces (@benoit74 #2429)
 * FIX: Check API availability at scraper startup instead of raw mwUrl which might not be reachable (@benoit74 #2447)
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -203,6 +203,7 @@ class Downloader {
           err.responseData?.error?.code &&
           ![
             'missingtitle',
+            'readapidenied',
             'internal_api_error_MediaWiki\\Revision\\BadRevisionException',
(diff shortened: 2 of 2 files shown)


benoit74:
@kelson42 please take a look, I would like to move this forward

Shall we merge this "as-is" and add error codes deemed to fail (not worth retrying) a bit later ? Retrying these fatal error code is not going to cause much harm, scraper will retry 10 times and then give up (fail) anyway. So missing few codes is not a very big deal.

kelson42:
> In the mean time I already noticed the `readapidenied` error code missing.

@benoit74 What about the mgmt of this specific kind of error?

kelson42 (approved):
@benoit74 Good to merge now IMHO. Obviously I woukd prefer to have @Markus-Rost proper review, but we also need to move forward. Agree with you that the hard error list will be updated later, but please put inside already what is already clearly known by us.

benoit74:
> In the mean time I already noticed the readapidenied error code missing.

> @benoit74 What about the mgmt of this specific kind of error?

It has already been added to the list of error causing scraper to immediately fail (no retry)

PR #2530, opened 2025-09-16
Better handling of audio resources in articles for full ZIMs

### LINKED ISSUE #2420: Unable to find an appropriate video/audio source for an media element in article
By [scraping https://inciclopedia.org, we get a lot of warnings of this type for many articles](https://farm.openzim.org/pipeline/dfc8f888-fce9-4d4d-9529-ea5585e79e51/debug):
```
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.522Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.523Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.523Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.523Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
[warn] [2025-07-09T10:21:06.523Z] Unable to find an appropriate video/audio source for an media element in article 'We_Will_Rock_You'
```

The problem is that the video/audio sources seem to be available, see https://inciclopedia.org/wiki/We_Will_Rock_You.

It looks like a bug to me.

benoit74:
Fix #2420

## Changes

- Differentiate handling of audio from video resources
- For both, take into account the (probably rare) situations where:
  - we have only an `src` attribute and no children `<source>`
  - we have an `src` attribute and children `<source>`
- For audio, implement a basic selection logic:
  - prefer the first ogg source if available
  - otherwise use the first source

### DIFF
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -106,35 +106,47 @@ export abstract class Renderer {
     }
   }
 
-  protected async treatVideo(
+  protected async treatAudioVideo(
     dump: Dump,
     srcCache: KVS<boolean>,
     articleId: string,
-    videoEl: DominoElement,
+    audioVideoEl: DominoElement,
   ): Promise<{ imageDependencies: string[]; videoDependencies: string[]; subtitles: string[] }> {
     const imageDependencies: string[] = []
     const videoDependencies: string[] = []
     const subtitles: string[] = []
 
     if (dump.nopic || dump.novid || dump.nodet) {
-      DOMUtils.deleteNode(videoEl)
+      DOMUtils.deleteNode(audioVideoEl)
       return { imageDependencies, videoDependencies, subtitles }
     }
 
-    this.adjustVideoElementAttributes(videoEl)
+    const isVideo = audioVideoEl.tagName === 'VIDEO'
+    const isAudio = audioVideoEl.tagName === 'AUDIO'
 
-    const chosenVideoSourceEl = this.chooseBestVideoSource(videoEl)
+    if (!isAudio && !isVideo) {
+      throw new Error(`Cannot treat non audio/video HTML element: \n${audioVideoEl.outerHTML}`)
+    }
 
-    if (!chosenVideoSourceEl) {
-      logger.warn(`Unable to find an appropriate video/audio source for an media element in article '${articleId}'`)
-      DOMUtils.deleteNode(videoEl)
+    if (isVideo) {
+      this.adjustVideoElementAttributes(audioVideoEl)
+    }
+
+    const chosenAudioVideoSourceEl = isVideo ? this.chooseBestVideoSource(audioVideoEl) : this.chooseBestAudioSource(audioVideoEl)
+
+    if (!chosenAudioVideoSourceEl) {
+      logger.warn(`Deleting audio/video HTML node missing an appropriate source in article '${articleId}':\n${audioVideoEl.outerHTML}`)
+      DOMUtils.deleteNode(audioVideoEl)
       return { imageDependencies, videoDependencies, subtitles }
     }
 
-    this.handleVideoPoster(videoEl, articleId, imageDependencies, srcCache)
-    this.updateVideoSrc(chosenVideoSourceEl, articleId, srcCache, videoDependencies)
+    if (isVideo) {
+      this.handleVideoPoster(audioVideoEl, articleId, imageDependencies, srcCache)
+    }
 
-    const trackElements = Array.from(videoEl.querySelectorAll('track'))
+    this.updateAudioVideoSrc(chosenAudioVideoSourceEl, articleId, srcCache, videoDependencies)
+
+    const trackElements = Array.from(audioVideoEl.querySelectorAll('track'))
     for (const track of trackElements) {
       subtitles.push(await this.treatSubtitle(track, articleId))
     }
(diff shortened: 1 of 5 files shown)


src/renderers/abstract.renderer.ts
```diff
@@ -222,6 +244,54 @@ export abstract class Renderer {
...
+    const audioSourceEls: any[] = Array.from(audioEl.children).filter((child: any) => child.tagName === 'SOURCE')
+
+    // Looks like in some cases we get the `src` in `resource` attribute
+    const originalSrc = audioEl.getAttribute('src') || audioEl.getAttribute('resource')
+
+    // Take into account the standard case where <audio> has no <source> child
+    if (audioSourceEls.length == 0) {
+      if (originalSrc) {
+        // If audio has a single `src` attribute, this is an acceptable src
+        return audioEl
```
Markus-Rost, line 267: We accept the `resource` attribute for audio elements as a fallback, however `updateAudioVideoSrc` only accepts the `src` attribute.

Markus-Rost (changes requested):

src/renderers/abstract.renderer.ts
benoit74: This was intentional, because I've never seen the `resource` attribute used in videos (while I do for audios). But I clearly miss experience on this. Do you confirm it could appear and we should fallback for videos as well? Or maybe we should fallback since this is anyway mostly harmless.

Markus-Rost: I mean that we are returning `audioEl` (with no `src`, only `resource`) as value for `updateAudioVideoSrc`, but that function won't expect the `resource` attribute.

I think the best solution is to set the `src` to the value of `originalSrc` before returing the audio element.

benoit74: `src` attribute will be set in `updateAudioVideoSrc` ; but we should indeed cleanup the `resource` attribute should it exists.

Markus-Rost: `updateAudioVideoSrc` only rewrites the `src` attribute of the audio/video element. However our audio element might not have a `src` attribute, having only `resource` instead. Now we are removing the `resource` attribute but never setting the `src` attribute. This means we are providing `updateAudioVideoSrc` with an audio element which has neither a `src` nor a `resource` attribute.

Markus-Rost: Also, can you link an example of audio elements using `resource`? Those elements wouldn't work so they are likely rewritten using JS onclick to play the audio. It seems likely to me that they are also rewritten using JS on load, with the actual html returned by the API probably having no-JS compatible audio elements. I would like to confirm that suspicion, but I can't find an instance of audio using `resource`.

benoit74: audio element using `resource` are not returned can be found here for instance: https://fr.wiktionary.org/w/api.php?action=parse&format=json&prop=templates|modules|jsconfigvars|headhtml|text&usearticle=1&page=maison&useskin=vector-2022&formatversion=2

These however have a set of `source` children. I did not encountered in the wild cases where we have only an <audio> element without sources. But I feel like it might happens.

> updateAudioVideoSrc only rewrites the src attribute of the audio/video element. However our audio element might not have a src attribute, having only resource instead. Now we are removing the resource attribute but never setting the src attribute. This means we are providing updateAudioVideoSrc with an audio element which has neither a src nor a resource attribute.

🤦🏻‍♂️ OMG

 I will add unit tests to cover this code, these discussions and "painful" code analyses would not have happened with proper unit tests. Sorry for that

benoit74: All things have been fixed, especially new unit tests to check everything is working as expected.

benoit74:
@kelson42 please take a look, I would like to move this forward

benoit74:
@kelson42 please take a look, I would like to move this forward 

Looks like I've solved (very valid) concerns of @Markus-Rost

kelson42:
@benoit74 Please rebase an fix conflcit so we can merge

PR #2531, opened 2025-09-16
Add `maxlag` API parameter to reduce pressure on Mediawiki servers during high load

### LINKED ISSUE #2365: Consider using maxlag in API requests
Scaping a wiki is usually not interactive, i.e. a user is not waiting for the result, so we should try to use the [`maxlag` parameter](https://www.mediawiki.org/wiki/Manual:Maxlag_parameter) on API calls in accordance with the [etiquette](https://www.mediawiki.org/wiki/API:Etiquette).

The parameter will cause requests to get rejected during high server load, so using it should probably be based on an option.

benoit74:
Fix #2365

Note: as indicated in the code, we currently do not respect the `Retry-After` header value since this is not possible with current code architecture and it is probably acceptable (exponential backoff is maybe even better since if the lag situation continues a bit, then it will give even more time for server to recover compared to the constant `Retry-After` value usually returned by Mediawiki servers).


kelson42:
Merging as we need to move forward

Issue #2572, opened 2025-11-06
Very long retry intervals


benoit74:
```
[log] [2025-11-06T13:38:48.820Z] Received a [status=429], slowing down upload.wikimedia.org to 37328747.82353801ms interval
```

Looks like upload.wikimiedia.org is "lying" and "always" returning 429 on some URLs like https://upload.wikimedia.org/wikipedia/commons/thumb/3/34/PAN_Logo.svg/40px-PAN_Logo.svg.png

This is very detrimental to our logic which assumes 429 are "real" query to slow down. I'm not sure what should be done.

benoit74:
Seen on https://farm.openzim.org/pipeline/8b75b35d-65e6-41db-a23d-df89e841d255/debug, but also https://farm.openzim.org/pipeline/f464bfb0-d82d-4485-9222-786f369d62e8/debug, https://farm.openzim.org/pipeline/cfa81b8a-520d-4dd2-84ee-03811685922f/debug

benoit74:
I've opened https://phabricator.wikimedia.org/T409450 on Wikimedia side

benoit74:
Proper upstream Wikimedia task is in fact https://phabricator.wikimedia.org/T175512

EDWARD-012:
Hi @benoit74 👋

I'd like to work on this issue. I've looked at the root cause — when a server returns 429, the scraper computes an unreasonably large retry interval (e.g. ~37 million ms) and effectively hangs.

My proposed approach:
- Introduce a `MAX_RETRY_INTERVAL_MS` cap inside the retry logic to prevent any single delay from exceeding a sane threshold (e.g. 5 minutes)
- Log a `warn` when the cap is triggered so operators can detect abnormal server behavior

One thing I'd like your input on should we also consider skipping the retry entirely when the computed interval is suspiciously large, rather than capping and retrying? Happy to discuss the best approach before implementing.

The change would go in `Downloader.ts` inside the `backoffOptions` strategy. Could you please assign this to me? Thanks!

benoit74:
There is already a maximum retry interval implemented:
- in backoff options (60s)
- in logic handling pause between host files downloads on a given host (20s)

The only reason this issue is still open is that for now we ignore all 429, 503 and 524 errors whose url match `/^https?:\/\/upload\.wikimedia\.org\/.*\/thumb\//` in the logic handling pause between host files downloads on a given host because we can't trust this value.

We are waiting on input from Wikimedia foundation or we will have to live with this.

PR #2574, opened 2025-11-13
Fix backoff and slow down issues

benoit74:
This is a containment PR for issue #2572 

Changes:
- do not slow down file downloads when the 429, 503 or 524 comes from wikimedia thumbnails (a bit hackish tbh, but not real other alternative ATM)
- reset backoff interval between queries: this is mostly unrelated, but discovered this issue while looking into the code, without the reset backoff always restarts at previous value


benoit74:
@kelson42 up

PR #2575, opened 2025-11-13
Ask for pageid in rdprop for proper mdwiki API operation

benoit74:
When retrieving article details including their redirects with `action=query`, we need `title` and `fragment` of each redirect. `pageid` is not required by mwoffliner because we do not manipulate IDs but titles.

However, mdwiki.org API needed a special `action=query` API implementation to retrieve redirects of mirrored articles.

And this special API implementation does not work ATM when it comes to have continuation of redirects (`rdcontinue`) if `pageid` is not asked for (and hence present in the `rdcontinue` value).

Since this is expected to have negligible impact in most cases, I propose we include the `pageid` in `rdprop` by default in mwoffliner. We could implement something special for mdwiki.org, but I do not feel like this is worth the additional complexity.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,4 +1,5 @@
 Unreleased:
+* CHANGED: pass `pageid` in `rdprop` for proper mdwiki API operation (@benoit74 #2557)
 
 1.17.2:
 * FIX: Fix parsing of potentially missing retry-after header (@benoit74 #2556)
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
LGTM if this is a no impact PR for non-mdiwiki instance (which seems to be the case).

@benoit74 Please link the PR to the issue (we had a discussion somewhere about this fix).

benoit74:
Nope, there is no issue, we discussed about it live, hence the fact this PR is in the milestone directly (instead of the issue)

PR #2587, opened 2025-12-15
change customZimFavicon flag from url to image blob

elfkuzco:
## Rationale
See openzim/zimfarm#1576

## Changes
- change customZimFavicon flag from url to image blob


benoit74 (approved):
LGTM

Issue #2588, opened 2025-12-15
Upgrade to node-libzim 4.0.0


benoit74:
https://www.npmjs.com/package/@openzim/libzim/v/4.0.0 has just been released, we need to upgrade.

No big breaking changes besides unused code around cluster cache management.

kelson42:
@kelvinhammond Is that something you might have a PR already?

kelvinhammond:
> @kelvinhammond Is that something you might have a PR already?
I have not created a PR for this.

benoit74:
Work has already been done and merged in https://github.com/openzim/mwoffliner/pull/2590 (only incorrectly linked to the issue)

PR #2590, opened 2025-12-16
Upgrade @openzim/libzim to v4.0.0

dhyaniritesh89:
This PR upgrades @openzim/libzim from v3.5.0 to v4.0.0 as suggested in #2588.

Notes on testing:
- Unit and non-S3 tests pass locally.
- S3 optimisation tests require real AWS credentials and an existing S3 bucket, so they were not runnable locally.
- CI should validate S3-related tests in the appropriate environment.


dhyaniritesh89:
Changelog updated in the Unreleased section as requested. Thanks!

PR #2591, opened 2025-12-16
Do not retry HTTP 429 without Retry-After header

dhyaniritesh89:


benoit74:
@dhyaniritesh89 can you explain why you propose this change? 

In practice I see that servers often reply with 429 with `Retry-After` header, so I see this change with a pretty negative apriori since it means more failure where we could have success, but I might miss something.

dhyaniritesh89:
@benoit74 
thanks for asking this change mainly comes from issue #2572 and what i saw while looking into it.
in some cases certain URLs keep returning HTTP 429 without a retry-after header and this seems to happen consistently for those URLs.

right now we treat all 429 responses as transient and retry with exponential backoff. when retry-after is missing, this can lead to very long backoff times from minutes to hours for requests that don’t seem to ever succeed, which ends up stalling the pipeline.

the idea behind this change was to be a bit more conservative:
if retry-after is present -  we retry as before
if it’s missing -  we stop retrying to avoid getting stuck backing off indefinitely

totally open to other approaches as well.

benoit74:
We've already implemented a containment action for https://github.com/openzim/mwoffliner/issues/2572 in https://github.com/openzim/mwoffliner/pull/2574 ; the problem is not that we retry, the problem is that on some hosts (upload.wikimedia.org) we should not increase request interval when receiving 429 because we can't be sure it is a real 429 for us or something else. I don't think this PR enhance anyhow the situation

dhyaniritesh89:
thanks for the explanation that makes sense. i hadn’t fully considered that 429 responses from upload.wikimedia.org can be ambiguous and not necessarily tied to our own request rate, so increasing the interval could indeed be counter productive in those cases. agreed that this doesn’t meaningfully improve the current behavior. i am fine closing the PR if you think that’s the right call.

Issue #2608, opened 2026-02-02
Multiple big WP wikis froze while downloading files


benoit74:
See https://farm.openzim.org/pipeline/10b12dcc-93d5-46d4-88b5-cebd87411cc3/debug (https://farm.openzim.org/recipes/wikipedia_uk_all)

Note this seems very close to https://github.com/openzim/mwoffliner/issues/2596 while not exactly identical, hence a distinct issue

Last log lines (many days ago):

```
[log] [2026-01-28T21:22:49.982Z] Progress downloading files [1596400/2265915] [70.5%]
[log] [2026-01-28T21:28:55.942Z] Progress downloading files [1598640/2265915] [70.6%]
```

benoit74:
Same problem on https://farm.openzim.org/pipeline/3986796c-e037-4563-9691-e59887d2dbaf/debug (https://farm.openzim.org/recipes/wikipedia_ru_all)

kelson42:
@benoit74 This is a priority to fix

benoit74:
I've created https://farm.openzim.org/recipes/wikipedia_ru_all_maxi, launched https://farm.openzim.org/pipeline/a315536c-3854-4831-b66c-3c397e17b158, just to try to create maxi of WPRU with dev version. Recipe to be deleted obviously once completed. Wait and see

benoit74:
Worked well this time ... Recipe deleted and file move to prod manually. I will request WPEN maxi now.

benoit74:
Looks like https://github.com/openzim/mwoffliner/pull/2612 solved the issue

PR #2612, opened 2026-02-09
Try to fix freeze issues

benoit74:
Trying to find a solution for https://github.com/openzim/mwoffliner/issues/2608, we have identified two things:
- there is no maximum retry interval, so it might happens the interval is way too big ; not sure it is our issue since we do not have corresponding messages in the few logs which have been kept on the Zimfarm, but we only have last log lines
- there is no safety net to exit infinite while loop when waiting for a file to be ready to download

I'm not convinced this will be enough to exit the freeze condition, but this is my best bet since I can't reliably reproduce the issue without waiting for hours.

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -805,7 +805,7 @@ class Downloader {
   }
 
   private errHandler(err: any, url: string, handler: any): void {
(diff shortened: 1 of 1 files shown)


benoit74:
CI errors are unrelated, let's merge

Issue #829, opened 2019-06-20
Various problems with category pages


TheNetStriker:
I've just tested the new --getCategories function and found a few issues:

1) Links to pages that contain slashes and points do not work. (e.g. "Page/Subpage" or "This.That") They do not seem to contain a link at all.
2) Links to pages that contain slashes do also not work on regular pages. I guess the problem is that the /A/ part is missing in the link. (e.g. zim://Page/Subpage instead of zim://A/Page/Subpage)
3) Page links on the category page are not in alphabetical order.
4) The sort keyword of the categories is not taken into account. (e.g. a page with the name Some:Page with the category tag [[Category:Test|Page]] is listed under the letter S instead of P in the Test category)

If those problems could be fixed the category funtion would be perfect for me.

kelson42:
@TheNetStriker Thank you for your feedback about categories, this is one of the first one and like you see this still a work-in-progress. I'll have a look in detail to what you have reported and might split-up the whole in many smaller tickets.

rgaudin:
Here a some additions based on [this bm zim](http://download.kiwix.org/zim/wikipedia/wikipedia_bm_all_maxi_2019-09.zim)

Category NS Alias: `Catégorie`

* Links to categories from bottom of articles are OK.
* `/U/Catégorie:Kan`
  * https://bm.wikipedia.org/wiki/Cat%C3%A9gorie:Kan
  * Under **J** should be only one link to sub-category **Jamu** (`/U/Catégorie:Jamu`)
  * Instead we have five links to that sub category.
  * Links to Pages are OK.
* `/U/Catégorie:Dunia`
  * https://bm.wikipedia.org/wiki/Cat%C3%A9gorie:Dunia
  * Links to sub-categories are OK (single sub appears twice instead of once though)
  * Pages lists the 2 pages present online plus a link to the sub-category.
  * That link is wrong (`/A/` instead of `/U/`)
* `/U/Catégorie:Afrika`
  * https://bm.wikipedia.org/wiki/Cat%C3%A9gorie:Afrika
  * Link in the text intro (not sub-categories section) goes to `/U/` instead of `/A/`
  * Links to sub-categories with no page (only sub categories) are absent

I also tested on [this ar zim](http://download.kiwix.org/zim/other/mawsouaa_ar_all_2019-09.zim) corresponding to [mawsouaa.tn](http://www.mawsouaa.tn)

[...]

Issue #2614, opened 2026-02-11
Use a proper i18n library


Markus-Rost:
We aren't currently using any i18n library for our translations. This is working fine for the few simple strings currently in mwoffliner, but will cause issues once we need more complex translations like in #2245 which requires support for [plural rules](https://translatewiki.net/wiki/Plural). As such we should start looking into adding a proper i18n library.

kelson42:
Do we use one for phets or kiwix-js? Maybe we could benefit of their experience with this challenge.

benoit74:
Phet doesn't use a localization library, it switches labels provided by the website.

Kiwix-js uses [i18next](https://www.i18next.com/) format but doesn't use the library.

https://github.com/kiwix/kiwix-js/blob/61c72f194f103b1dd5dcb70a1f2d130666728251/www/js/lib/translateUI.js#L25-L32

benoit74:
Note that restrictions identified by kiwix-js do not apply in mwoffliner, since we will not run localization in the client but scraper-side, when generating the ZIM

PR #2617, opened 2026-02-11
Support category pages without pagination

### LINKED ISSUE #762: Category graph should be simplified in case of subselection
Example to be given by @kelson42

### LINKED ISSUE #2749: Implement categories retrieval with simplification
Simplifications:
- no handling of pagination (#2620)
- simplified i18n (needs #2614 for real implementation)
- optionally support category graph simplification (#762)

Markus-Rost:
Start supporting (smaller) categories (#2245)

1. **getCategories**: This PR removes all of the old existing code relating to categories and `getCategories`, that option has been disabled for a while and starting from scratch seems like the easier approach for me to work on this.
   - Setting the option will add all category pages which have members inside the ZIM. The member lists on the category pages will also be limited to only pages inside the ZIM.
2. **categorieshtml**: Using the ActionParse renderer, *all articles* will now have the box listing the categories the article is a member of. Like normal, this will either link to the category pages (if they are part of the ZIM) or display them as plain text (if they aren't). image
3. **Category NS pages**: If there are category pages included in the ZIM, they will now have the expecting lists of their members (subcategories, pages and files). Files will display in a gallery, unless that has been disabled for the specific category page.
   - Pagination is not yet supported in this PR, category pages will only list the first 200 members of each type for larger categories. See also #2620
   - Localization of category pages is not fully implemented, see also #2614 

Fix #2749
Fix #762


benoit74 (commented):
I only have few questions for now:
- do I get it correctly that with this approach, we need to add category NS to the list of NS we want (if we want it) or specific category pages to the list of articles we want (if we want only few category pages) ? If I got this correctly, I really like the approach
- what about category pages with millions of subpages, can we expect the API to answer in a timely manner to the query to get this category data?
- do I get it correctly that should we drop the `getCategories` from parameter list?

Markus-Rost:
> I only have few questions for now:
> 
> * do I get it correctly that with this approach, we need to add category NS to the list of NS we want (if we want it) or specific category pages to the list of articles we want (if we want only few category pages) ? If I got this correctly, I really like the approach

Yes, that is correct.

> * what about category pages with millions of subpages, can we expect the API to answer in a timely manner to the query to get this category data?

The query itself is not expensive. Note the the request has the same limit of pages returned as the "allpages" query (500 by default, 5000 with `apihighlimits` permission). I assume on any wikis with millions of pages in a category, we would also be logged into an account with the higher limits anyway.

> * do I get it correctly that should we drop the `getCategories` from parameter list?

Yes, considering I have removed all the existing old category code.

benoit74:
I really enjoy this approach. 

I feel like we could even merge this quite soon without support for localization and without support for pagination of big categories. This is a long awaited feature, and I'm quite sure it will provide sufficient added value for users to accept to not have a fully-ready feature, it would already be very useful + since we need to add the Category NS / Category pages, it is still up to the user to decide if the feature is used or not.

It would also be a good opportunity to start getting feedback from real use-cases.

WDYT? Is there anything else significant which should make it in this PR before we merge, so that we do not break something?

Markus-Rost:
I can certainly make this ready without the complex changes like localization and pagination, adding them in a later PR.

Note that we probably wouldn't want to include the whole category NS in a ZIM, at least for the huge wikis. Articles are usually the minority of pages on a wiki, with file pages and other non-reader facing pages taking up the majority. This in turn means that the biggest categories with the most members are usually not something we would care about in a ZIM file (like file license categories).
I might look into adding back something like `getCategories` in the followup PR to make this more user friendly.

Markus-Rost:
Created #2620 as a follow-up issue to implement pagination.

kelson42:
@Markus-Rost Thank you very much for working on this feature. I'm so glad this happens after the way has been paved with the work of last year on MWoffliner! I have added myself as reviewer as I have quite a few strong opinion on this feature.

Here my remarks and questions:
* Can we please have a PR description describing what exactly bring this PR from a user perspective? And list clearly the limitations of current implementation.
* I don't think using the `--addNamespaces` is the most appropriate way to specify including category pages withing the ZIM. The reasons are (1) this option is primary needed in special case where the ZIM publisher estimates that the Mediawiki maintainer has not included properly all "content namespaces", but in normal case mwoffliner get them automatically. Here we want to add something which is very particular and pretty much not "content" (2) Including this options does not just add new pages, it modify the rendering of potentially all pages (3) This will be a very successful option, and requesting users they now the number of the namespace "Category" is not user friendly. This is why we need a dedicated option (but `--getCategories` is maybe not the best name?!)
* I understand this is the first PR of a train of others. We have a label for that at https://github.com/openzim/mwoffliner/issues?q=state%3Aopen%20label%3A%22category%22. I would like that the overviews on these ticket is managed.
* One question is, do you PR fixes #829 as well. What is the impact of this PR related category links and category listing at the bottom of any page?
* I have obviously a few other questions, but I guess, just putting the PR description will answer most of them.

Markus-Rost:
> @Markus-Rost Thank you very much for working on this feature. I'm so glad this happens after the way has been paved with the work of last year on MWoffliner! I have added myself as reviewer as I have quite a few strong opinion on this feature.
> 
> Here my remarks and questions:
> 
> * Can we please have a PR description describing what exactly bring this PR from a user perspective? And list clearly the limitations of current implementation.

Yes of course, I've added a more detailed description to this PR.

> * I don't think using the `--addNamespaces` is the most appropriate way to specify including category pages withing the ZIM. The reasons are (1) this option is primary needed in special case where the ZIM publisher estimates that the Mediawiki maintainer has not included properly all "content namespaces", but in normal case mwoffliner get them automatically. Here we want to add something which is very particular and pretty much not "content" (2) Including this options does not just add new pages, it modify the rendering of potentially all pages (3) This will be a very successful option, and requesting users they now the number of the namespace "Category" is not user friendly. This is why we need a dedicated option (but `--getCategories` is maybe not the best name?!)

I absolutely agree with this, some functionality like the old `getCategories` is definitely needed. However that's not exactly the focus of this PR. In this PR I'm just making category pages work as you would expect (listing the members, like on the wiki) so that the future PRs have a working base to build upon. 
As @benoit74 suggested, this is a basic PR with basically no major changes for users who don't include category pages manually, but allowing users to specifically include categories to test and provide feedback. This gives us time to properly implement the QoL options later on (I myself don't have the time to implement everything related to categories myself in a reasonable amount of time, so doing it in chunks helps to keep track)

> * I understand this is the first PR of a train of others. We have a label for that at https://github.com/openzim/mwoffliner/issues?q=state%3Aopen%20label%3A%22category%22. I would like that the overviews on these ticket is managed.

I've added the label to the existing issues and PRs relating to this and will create a few more follow-up issues over the next few days for features to implement later on.

> * One question is, do you PR fixes [Various problems with category pages #829](https://github.com/openzim/mwoffliner/issues/829) as well. What is the impact of this PR related category links and category listing at the bottom of any page?

I can't really comment on the problems related to paths, but those don't seem relevant anymore as that was specific to the old implementation which this PR removed completely.
The problems mentioned about respecting the sortkey and alphabetical order should be fixed by this PR, as it is getting the ordered list directly from the wiki API.

> * I have obviously a few other questions, but I guess, just putting the PR description will answer most of them.

Let me know if you have any more questions :)

kelson42:
@Markus-Rost Thank you for the clarification. All good on my side and good to merge once CI is green and @benoit74 OK with it.

kelson42:
@Markus-Rost I have tested this PR against a full scrape of https://bm.wikipedia.org.

As a result:
* I see the "Category" label in French. I would expect this label to be either translated in Bambara or in default language English
image

* Title of the page is in French as well (probably related reason)
image

* There is a problem with the gathering of categories items by letter, see (many time the letter "B" for example):
image

Markus-Rost:
@kelson42 the "Catégorie" appearances match the text on the wiki.

I'll look into the grouping issue this weekend.

Markus-Rost:
> @Markus-Rost I have tested this PR against a full scrape of https://bm.wikipedia.org.
> 
> As a result:
> 
> * I see the "Category" label in French. I would expect this label to be either translated in Bambara or in default language English
>
> * Title of the page is in French as well (probably related reason)

The fallback language chain for the wiki is `bm`, `fr`, `en`. The Bambara localization for MediaWiki is probably incomplete and missing those translations, so it's falling back to French.

> * There is a problem with the gathering of categories items by letter, see (many time the letter "B" for example):

Fixed, I forgot to actually update the last prefix 🤦‍♂️

kelson42:
@Markus-Rost Thanks! Makes sense, will do another test run tomorrow.

Markus-Rost:
@kelson42 I have re-implemented `getCategories` inside this PR. Setting the option will add all category pages which have members inside the ZIM. The member lists on the category pages will also be limited to only pages inside the ZIM.

kelson42:
@Markus-Rost Sorry for the delay, will come back to you ASAP on this.

benoit74 (commented):
Two questions:
- shouldn't we add small e2e tests with categories to check we retrieve the expected categories with expected content? maybe just a test with a selection of articles, retrieving the corresponding categories?
- WDYT about the --getCategories CLI argument? I'm not a big fan of the `get` prefix. In other arguments with already use `keep`, `add`, `without`, I feel like `--withCategories` would make more sense, but this is maybe more a personal taste ; or maybe even better `--withoutCategories` because the default should probably be to always fetch categories

I also ran with `--mwUrl=https://bm.wikipedia.org --getCategories --articleList Latin,Lionkan,Denba` and it looks like filtering is broken:
- both `Kan` and `Yiri` categories contains no pages, instead of the few selected ones
- `Kan` category mentions a subcategory but there is none

I tried to have a look but I failed to find the issue.

Markus-Rost:
> Two questions:
> 
> * shouldn't we add small e2e tests with categories to check we retrieve the expected categories with expected content? maybe just a test with a selection of articles, retrieving the corresponding categories?

Yes we definitely should add tests, I just didn't get to that yet.

> * WDYT about the --getCategories CLI argument? I'm not a big fan of the `get` prefix. In other arguments with already use `keep`, `add`, `without`, I feel like `--withCategories` would make more sense, but this is maybe more a personal taste ; or maybe even better `--withoutCategories` because the default should probably be to always fetch categories

I don't have any strong opinion on the name of the CLI argument, however a name like `--withoutCategories` might get confusing as you can still add category pages manually.

> I also ran with `--mwUrl=https://bm.wikipedia.org --getCategories --articleList Latin,Lionkan,Denba` and it looks like filtering is broken:
> 
> * both `Kan` and `Yiri` categories contains no pages, instead of the few selected ones
> * `Kan` category mentions a subcategory but there is none
> 
> I tried to have a look but I failed to find the issue.

I've fixed both of those issues.

benoit74:
I've been blocked in my tests by #2658, but this looks pretty promising. Should we merge this without tests and open another issue for that? I feel like we already have quite important added value in this PR, and it is not like we might release 2.0 without full category support since it is one of the most awaited feature of this release.

benoit74:
Just rebased on top of main, I will do one last test and I think we are good to merge this one as well.

benoit74:
I think I've found yet another issue: I've made a full ZIM of bm.wikipedia.org and looking at `Kan` category I notice that:
- display of subcategories is a bit broken (not sure we can / should fix this since this might be an upstream issue in the wiki)
- some articles are missing in the list despite being present in the ZIM and associated with the category

Online: https://bm.wikipedia.org/wiki/Cat%C3%A9gorie:Kan

In the ZIM:

image

Article `Solomana_Kante` is missing in the list but still correctly associated with the category in the article:

Screenshot 2026-03-24 at 10 35 03


@Markus-Rost Is this something you wanna fix now or should we merge this and fix afterwards?

kelson42:
`--withCategories` sounds a better name. I kind of have a struggle to test the full feature again. I'm also a bit unsure about the default behaviour (with or without the feature), but I guess you will make the right decision.

Markus-Rost:
> I think I've found yet another issue: I've made a full ZIM of bm.wikipedia.org and looking at `Kan` category I notice that:
> 
> * display of subcategories is a bit broken (not sure we can / should fix this since this might be an upstream issue in the wiki)

Not sure what happened there, I'm unable to reproduce it. Can you take a look at the CSS applied there?

> * some articles are missing in the list despite being present in the ZIM and associated with the category

This looks like the same issue kelson had and I already fixed (last latter missing). Are you sure you were using the most recent commit?

> @Markus-Rost Is this something you wanna fix now or should we merge this and fix afterwards?

I was planning to add some based tests before merging this, but my time is pretty limited at the moment so I didn't get to it yet. I'll leave it up to you if you want to merge as is or wait for the basic tests. I'll create follow-up PRs for pagination etc in any case.

Markus-Rost:
@benoit74 I fixed the display of lists with less than 6 articles (don't have them split into multiple columns) as well as resolved an issue with pages containing spaces not being listed (we really need to improve the situation of mwoffliner using "Page_name" to identify pages while MW uses "Page name").

benoit74:
@Markus-Rost I'm now moving my focus back to mwoffliner for few weeks to complete 2.0 release. Do you have time in coming days to add some tests in this PR or should I proceed with the merge and track the fact we wanna add tests in a follow-up issue? 

Btw, current agenda is to complete 2.0 release before end of July, which means having something mostly ready by mid-July so that we can test and fix remaining bugs. Is this an agenda which fits the fact you will complete all "category-related" PRs yourself or should we begin to consider changing the plan because your time is still very too limited? Note I expect this plan to be fragile anyway since I hope since 2 months I would be back working on mwoffliner, so I would not be surprised if I do not have sufficient time myself to complete the effort on time.

Markus-Rost:
@benoit74 Sadly my time is still very much limited. I'll see if I can make some time to work on the category issues, however I think probably the best to plan with that not being ensured.

benoit74:
@Markus-Rost no worries, you've already brought a lot to the scraper. And I prefer you spend the little time you have on helping us when we have hard questions about Mediawiki rather than writing tests others could write as well.

I will rebase and merge this PR as-is, work on other issues and then we will see if I complete work on category issues or you finally have some free time for that.

Thanks a lot for you feedback

benoit74:
Rebased

I also added some very basic tests. Nothing I'm extremely proud of, but at least it ensures we do not break categories retrieval by inadvertence.

benoit74 (approved):

Issue #2601, opened 2026-01-20
Recommend Docker image for using mwoffliner instead of installing locally


benoit74:
We need to recommend Docker image in `README.md` rather than installing locally which can cause all kind of headaches.

KunalSiyag:
I will help document the Docker image in the `README.md` as I build it in my own system to test it out. Can I be assigned this?

benoit74:
Please recommend to use the official image at `ghcr.io/openzim/mwoffliner`. More or less what we have at https://github.com/openzim/mindtouch

KunalSiyag:
> Please recommend to use the official image at `ghcr.io/openzim/mwoffliner`. More or less what we have at https://github.com/openzim/mindtouch

Done with the changes. Documented it as per how I ran it in my system.

benoit74:
Closed by #2618

PR #2618, opened 2026-02-16
docs: recommend Docker image as primary installation method (#2601)

KunalSiyag:


KunalSiyag:
> Thank you!
> 
> I feel like proposing three installation methods is too confusing for new comers, especially since the usage then assumes you are using the Docker installation. It also forces us to keep prerequisites section as-is while this is true only for non-Docker installation.
> 
> I propose to:
> 
> * mention only Docker (or Docker-based engine) and `amd64` in the `Prerequisites` (we do not support `arm64` ATM AFAIK) ; I assume image works fine on Windows as well
> * make one visible (no details) installation method in the `Installation` section, and keep other "hidden" with details like now
> * move prerequisites to the "Run software locally" details
> 
> WDYT?

Sounds good.  If regular people come to use it, they would be confused with the prerequisites. I am thinking of adding the usage with npm? It would be hidden but could be used depending on the user. Should we also mention a reference to Docker documentation for those who don't use it already?

benoit74:
> I am thinking of adding the usage with npm? 

Yes, this is indeed missing

> Should we also mention a reference to Docker documentation for those who don't use it already? 

Problem with that is that I'm not a fan of recommending Docker Desktop anymore given its licensing model, but I'm not that confident in another alternative (Rancher, ...) to recommend it. Let's live without it for now.

KunalSiyag:
> > I am thinking of adding the usage with npm?
> 
> Yes, this is indeed missing
> 
> > Should we also mention a reference to Docker documentation for those who don't use it already?
> 
> Problem with that is that I'm not a fan of recommending Docker Desktop anymore given its licensing model, but I'm not that confident in another alternative (Rancher, ...) to recommend it. Let's live without it for now.

Done with the changes and added a changelog entry also.  Let me know if there is some additional things to be revised here.

KunalSiyag:
Done with the changes. Squashed everything in a final commit.

PR #2621, opened 2026-02-18
Fix/2609 retry when malformed json

### LINKED ISSUE #2609: Retry when API response is malformed JSON
See https://farm.openzim.org/pipeline/991eaae5-e5c1-4b33-9856-73c39f2d9f6d/debug (https://farm.openzim.org/recipes/dwarffortresswiki.org_en_all)

See https://farm.openzim.org/pipeline/4d7fe54c-5c18-4fa2-a0da-80f033fab681/debug (https://farm.openzim.org/recipes/vikidia_fr)

Looks like we get into a condition where the connection terminates before all data is sent but connection is terminated normally. Not sure exactly what can cause that, but probably worth to catch and retry.

KunalSiyag:
Fixes #2609

## Changes

What changed:
- JSON requests are parsed strictly, so truncated/invalid JSON throws.
- Retry detection is limited to real parse failures:
  - direct SyntaxError
  - Axios ERR_BAD_RESPONSE with SyntaxError in cause
- Generic ERR_BAD_RESPONSE is not treated as malformed JSON.

Validation:
- Unit tests pass in test/unit/downloader.retry.test.ts.
- Real run against [dwarffortresswiki.org_en_all](https://farm.openzim.org/recipes/dwarffortresswiki.org_en_all) succeeded locally.
- Logs show malformed JSON warnings and immediate retries, then full completion with exit code 0 and no article/file failures.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,6 +2,7 @@ Unreleased:
 * NEW: Add support for ResourceLoader based JavaScript (@Markus-Rost #2310)
 * CHANGED: Keep only ActionParse renderer (@benoit74 #2210)
 * CHANGED: Keep inline scripts if all JS is allowed (@Markus-Rost #2555)
+* FIX: retry malformed JSON API responses (@kunalsiyag #2609)
 
 1.17.5:
 * FIX: Ignore articles returning permissiondenied error code (@kunalsiyag #2604)
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -71,6 +71,18 @@ interface CompressionData {
   data: any
 }
 
+function isMalformedJsonResponseError(err: unknown): boolean {
+  if (err instanceof SyntaxError) {
+    return true
+  }
+
+  if (!axios.isAxiosError(err) || err.code !== AxiosError.ERR_BAD_RESPONSE) {
+    return false
+  }
+
+  return err.cause instanceof SyntaxError
+}
+
 export class DownloadError extends Error {
   urlCalled: string | null
   httpReturnCode: number | null
@@ -179,6 +191,11 @@ class Downloader {
       failAfter: 10,
       retryIf: (err: any) => {
         const requestedUrl = err.urlCalled || err.config?.url || 'unknown'
+        if (isMalformedJsonResponseError(err)) {
+          logger.log(`Retrying ${requestedUrl} due to malformed JSON response`)
+          return true
+        }
+
         if (err instanceof AxiosError && err.code && !['ERR_BAD_REQUEST', 'ERR_BAD_RESPONSE'].includes(err.code)) {
           logger.log(`Retrying ${requestedUrl} URL due to ${err.code} error`)
           return true // retry all connection issues
@@ -247,6 +264,11 @@ class Downloader {
         'accept-encoding': 'gzip, deflate',
       },
       responseType: 'json',
+      transitional: {
+        // Parse failures must throw so backoff retries can be triggered.
+        silentJSONParsing: false,
+        forcedJSONParsing: true,
+      },
       method: 'GET',
     }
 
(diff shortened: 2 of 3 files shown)


KunalSiyag:
> Thank you.
> 
> Can you explain why we need to consider so many cases as a malformed JSON? Reading the doc, it looks like error is going to be `SyntaxError` since we've set `silentJSONParsing` to `false`. In particular, I'm worried about catching all `ERR_BAD_RESPONSE` as JSON errors, AFAIK this error arises for any 502 error, which might happen for many other reasons.
> 
> Did you tried your code on one of the cases of the issue?
> 
> I know the errors are transient, but https://farm.openzim.org/recipes/dwarffortresswiki.org_en_all repeatedly failed with this kind of error. Successfully making a ZIM of this (small, 17849 articles) wiki would prove your fix seems to work well.

Good point. I’ll narrow the detection so we only retry real JSON parse errors (SyntaxError, whether direct or wrapped as an Axios cause), and drop the generic ERR_BAD_RESPONSE catch.

Before I do a full local build for the Dwarf Fortress ZIM, I was wondering if there is an online pipeline I can use to run this integration test? Is there a bot command to trigger a farm test run directly from my PR branch, or could I push a test Docker image to my own Docker Hub registry for a temporary farm recipe to pull?

KunalSiyag:
I have tried a test run with the Dwarffortress and a ZIM have been made. I tried to use it with Kiwix (It works). The logs for the run are  [Log_Dwarffortress.log](https://github.com/user-attachments/files/25436455/Log_Dwarffortress.log)

Changelog
```diff
@@ -1,4 +1,7 @@
 Unreleased:
+* FIX: retry malformed JSON API responses (@kunalsiyag #2609)
```
benoit74, line 5: I don't think you've done so many modifications in this PR 😉

benoit74:
> Before I do a full local build for the Dwarf Fortress ZIM, I was wondering if there is an online pipeline I can use to run this integration test? Is there a bot command to trigger a farm test run directly from my PR branch, or could I push a test Docker image to my own Docker Hub registry for a temporary farm recipe to pull?

We do not have such things. We do build `main` branch continuously as a `dev` Docker image, and I will indeed test this with Dwarf Fortress Wiki in the Zimfarm before making a release.

Changelog
KunalSiyag: Yes, definitely not a part of this PR 😓

src/Downloader.ts
```diff
@@ -74,6 +74,18 @@ interface CompressionData {
   data: any
 }
 
+function isMalformedJsonResponseError(err: unknown): boolean {
+  if (err instanceof SyntaxError) {
+    return true
+  }
+
+  if (!axios.isAxiosError(err) || err.code !== AxiosError.ERR_BAD_RESPONSE) {
```
benoit74, line 79: I still don't get why we would need this.

Why not simply having something like `return (err instanceof SyntaxError) || (axios.isAxiosError(err) && err.cause instanceof SyntaxError)` ?

Even if I'm not even convinced it needs to be that complex, maybe just `return err instanceof SyntaxError` is enough.

KunalSiyag: Good question.  Because Axios does not always expose malformed JSON as a plain SyntaxError. Depending on version / adapter, we may get an ERR_BAD_RESPONSE with the parsing error stored in cause, and relying only on instanceof SyntaxError can miss those cases. I wanted to tighten the cases based on the previous feedback. The checks simply ensure retries trigger consistently while avoiding unrelated SyntaxErrors. 

Happy to simplify if you’d prefer the shorter form.

benoit74: OK got it.

KunalSiyag:
Done! I've squashed the commits into a single one. Ready for merge. Thank you for the feedbacks.

kelson42:
Please remove lasr merge commit. This is not allowed. Do a rebase instead.

KunalSiyag:
Rebasing done and merge commit removed. Force-pushed the updated branch.

KunalSiyag:
Done with the formatting.

PR #2622, opened 2026-02-19
Keep only ActionParse renderer, drop other ones

### LINKED ISSUE #2210: Do we still need multiple renderers ?

### LINKED ISSUE #2585: CI (again) failing on unit and e2e tests
Would be good to fix it ASAP so we don't get use of having it failing. 

See
https://github.com/openzim/mwoffliner/actions/workflows/ci.yml

benoit74:
Fix #2210
Fix #2585

Changes:
- drop all renderers except `ActionParse`
- simplify some tests which were made too complex due to the presence of many renderers
- drop `--mwRestApiPath` since it is absolutely not used anymore (other ones will be dropped in https://github.com/openzim/mwoffliner/issues/2489)
- upgrade version to `2.0.0-dev0` to clearly indicate we are now on the way to a major release

### DIFF
--- a/test/e2e/en.e2e.test.ts
+++ b/test/e2e/en.e2e.test.ts
@@ -69,6 +69,7 @@ await testAllRenders('en-wikipedia', parameters, async (outFiles) => {
       const expectedClasses = [
         'client-nojs',
         'skin-theme-clientpref-os',
+        'skin-theme-clientpref-thumb-standard',
         'vector-feature-appearance-pinned-clientpref-0',
         'vector-feature-custom-font-size-clientpref-1',
         'vector-feature-language-in-header-enabled',


benoit74:
@kelson42 mentioned that fixing CI should be our top priority, before merging any other PR.

I believed that moving to self-hosted runner would solve throttling issues, or increasing default requests interval would help, but **none of it really helped**.

The proper solution seems to remove all renderers but ActionParse.

I propose to now merge this PR (if it suits you). Also because it will make any other PR development way easier since we won't have to care about the various unused renderer. And then rebase all "stalled" PRs on top of it so that I can review them more easily with CI passing.

src/renderers/rendering.context.ts
```diff
@@ -29,9 +29,7 @@ class RenderingContext {
       this.articlesRenderer = renderer
     } else {
       this.mainPageRenderer = await rendererBuilder.createRenderer({ renderType: 'desktop' })
```
Markus-Rost: Is the separate mainPageRenderer still needed? Seems like that one was specific to the WikimediaMobile renderer and should be removed in this PR.

benoit74: This would indeed simplify the codebase. I thought we might need it again at some point should we decide to introduce yet another renderer which needs something special on main article. But I'm not convinced we will need another renderer, and even less convinced it might need a special renderer for the main page.

Let's remove it to keep code easier to maintain, it is not like it would be hard to re-implement.

src/Downloader.ts
Markus-Rost: This function is now unused, as are `config.output.mw.js` and `config.output.mw.css`. They should probably be removed.

benoit74: It was used in a test you did not dropped in https://github.com/openzim/mwoffliner/pull/2483 ; not sure about the test added value, but I brought it back in https://github.com/openzim/mwoffliner/pull/2622/commits/6cc5b145ba65a50a729d2981d5d407537b78dc73, ready to work once we retrieve again JS config vars from the API.

Markus-Rost: All that test did was running `Downloader.getModuleDependencies('Potato')` once for each renderer (why, the test doesn't change at all depedning on renderer) and testing the output of that function. I didn't drop it in #2483 because it is unrelated to the PR (function not used by ActionParse, JS PR only for ActionParse), I only made a minor change to account for the updated typing of `moduleDependencies.jsConfigVars` (string to KVS).

benoit74: OK, so let's drop it as well then

benoit74:
@Markus-Rost do you feel like we can merge this one? 

Since it allows CI to pass, it would unblock all other PRs, especially the "Javascript support" one which I would love to merge asap.

benoit74:
Rebasing + simplifying commits

PR #2623, opened 2026-02-19
Increase default request interval for files download

### LINKED ISSUE #2585: CI (again) failing on unit and e2e tests

benoit74:
In an attempt to stabilize CI, it looks like file downloads causing many errors is causing many CI errors. This tries to stabilize the CI by providing a more meaningful default request interval.

Fix #2585


benoit74:
Still failing, even with 100ms default interval between each image fetch. e2e test works fine individually ran locally (and probably as well in the CI).

I have no clue about how to move this forward. e2e tests have been known to be risky in mwoffliner, looks like we have now reached the point where they are a burden.

benoit74:
This is a dead end, let's focus on https://github.com/openzim/mwoffliner/pull/2622 for the time being.

PR #2629, opened 2026-02-19
Issue-2481-support-third-party-youtube-videos

### LINKED ISSUE #2481: Support third-party hosted (Youtube) videos
See [openzim/zim-requests/issues/9#issuecomment-3195551625](../../zim-requests/issues/9#issuecomment-3195551625). Cyclowiki.ru has embedded videos (ex. [here](https://cyclowiki.org/wiki/%D0%94%D1%80%D0%B5%D0%B2%D0%BD%D0%B5%D0%B3%D1%80%D0%B5%D1%87%D0%B5%D1%81%D0%BA%D0%B8%D0%B9_%D1%8F%D0%B7%D1%8B%D0%BA)) that are not captured.

KunalSiyag:
Fixes #2481

Improves handling of YouTube videos embedded via <iframe> elements and other iframe elements.

Behavior

- Previously, all iframes were removed, causing YouTube embeds to disappear in ZIM output.
- Detects YouTube-specific iframe URLs (youtube.com, youtube-nocookie.com, youtu.be).
- Replaces YouTube iframes with a responsive external-video placeholder.
- Extracts the video ID to link to the canonical https://www.youtube.com/watch?v=<ID> and display the video thumbnail.
- If the video ID cannot be extracted, renders a generic external link instead of producing a broken embed.
- The rewrite occurs before media processing so thumbnails are handled by the existing asset pipeline.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,4 +1,5 @@
 Unreleased:
+* UPDATE: Replace iFrame with a placeholder card instead of deleting them (@kunalsiyag #2481)
 * NEW: Scrape only wikitext contentmodel articles by default (@triemerge #2445)
 * FIX: In nopic mode, fixed layout issues by preserving image layout using SVG placeholders (@kunalsiyag #1004)
 * NEW: Add support for ResourceLoader based JavaScript (@Markus-Rost #2310)
--- a/res/style.css
+++ b/res/style.css
@@ -115,3 +115,10 @@ summary.section-heading {
 body[data-useragent*='KAIOS'] h1 {
     display: none !important;
 }
+
+.mwo-external-video-card {
+    display: block;
+    margin: 1em 0;
+    border: 1px solid #a2a9b1;
+    background-color: #f8f9fa;
+}
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -608,6 +608,9 @@ export abstract class Renderer {
           return { url, path }
         }),
     )
+    // applyOtherTreatments must run before treatMedias so that iframe
+    // placeholders (e.g. YouTube thumbnails) are already in the DOM
+    // when treatMedias processes <img> tags for download into the ZIM.
     doc = this.applyOtherTreatments(doc, dump, articleId)
 
     const tmRet = await this.treatMedias(doc, dump, articleId)
@@ -785,12 +788,89 @@ export abstract class Renderer {
     element.parentElement.innerHTML = `${slices[0]}<!--htdig_noindex-->${element.outerHTML}<!--/htdig_noindex-->${slices[1]}`
   }
 
-  private removeIframeTags(parsoidDoc: DominoElement) {
-    // Remove all <iframe> tags
+  // Rewrite iframe tags into external placeholders
+  private processIframeTags(parsoidDoc: DominoElement) {
     const iframes: DominoElement[] = Array.from(parsoidDoc.getElementsByTagName('iframe'))
+
     for (const iframe of iframes) {
-      DU.deleteNode(iframe)
+      const src = iframe.getAttribute('src')
+      if (!src) {
+        DU.deleteNode(iframe)
+        continue
+      }
+
+      let fullSrc = src
+      try {
+        fullSrc = getFullUrl(src, MediaWiki.baseUrl)
+      } catch {
+        // Keep original src when URL parsing fails
+      }
+
+      const placeholder = this.isYouTubeIframeUrl(fullSrc)
+        ? this.createYouTubePlaceholder(parsoidDoc, iframe, fullSrc, this.extractYouTubeVideoId(fullSrc))
+        : this.createExternalIframePlaceholder(parsoidDoc, iframe, fullSrc)
+      if (iframe.parentNode) {
+        iframe.parentNode.replaceChild(placeholder, iframe)
+      } else {
+        DU.deleteNode(iframe)
+      }
     }
+  }
+  private isYouTubeIframeUrl(url: string) {
+    const lowerUrl = url.toLowerCase()
+    return lowerUrl.includes('youtube.com/embed/') || lowerUrl.includes('youtube-nocookie.com/embed/') || lowerUrl.includes('youtu.be/') || lowerUrl.includes('youtube.com/watch')
+  }
+
+  private extractYouTubeVideoId(url: string): string | null {
+    const match = url.match(/(?:embed\/|v=|\.be\/)([a-zA-Z0-9_-]+)/)
+    return match ? match[1] : null
+  }
+
+  private createExternalIframePlaceholder(parsoidDoc: DominoElement, iframe: DominoElement, originalUrl: string): DominoElement {
+    const card = parsoidDoc.createElement('div')
+    this.copyNodeAttributes(iframe, card)
+    const iframeClass = card.getAttribute('class')
+    card.setAttribute('class', iframeClass ? `${iframeClass} mwo-external-video-card` : 'mwo-external-video-card')
+
+    const link = parsoidDoc.createElement('a')
+    link.setAttribute('class', 'external')
+    link.setAttribute('href', originalUrl)
+    link.textContent = 'Open embedded content (external)'
+    card.appendChild(link)
+    return card
+  }
+
+  private copyNodeAttributes(sourceNode: DominoElement, targetNode: DominoElement) {
+    for (let attributeIndex = 0; attributeIndex < sourceNode.attributes.length; attributeIndex++) {
+      const attribute = sourceNode.attributes.item(attributeIndex)
+      if (!attribute) {
+        continue
+      }
+      targetNode.setAttribute(attribute.name, attribute.value)
+    }
+  }
+
+  private createYouTubePlaceholder(parsoidDoc: DominoElement, iframe: DominoElement, originalUrl: string, videoId: string | null): DominoElement {
+    const link = parsoidDoc.createElement('a')
+    link.setAttribute('class', 'external')
+    link.setAttribute('href', videoId ? `https://www.youtube.com/watch?v=${videoId}` : originalUrl)
+
+    if (!videoId) {
+      link.textContent = 'View this content externally'
+      link.setAttribute('aria-label', 'View this content externally')
+      return link
+    }
+
+    const thumbnail = parsoidDoc.createElement('img')
+    this.copyNodeAttributes(iframe, thumbnail)
+    // Set the online thumbnail URL here; treatMedias will process this
+    // <img> tag later to download the image and rewrite src to a local path in the ZIM.
+    thumbnail.setAttribute('src', `https://img.youtube.com/vi/${videoId}/hqdefault.jpg`)
+    thumbnail.setAttribute('alt', 'Video thumbnail')
+    thumbnail.setAttribute('loading', 'lazy')
+
+    link.appendChild(thumbnail)
+    return link
   }
 
   private removeCitations(parsoidDoc: DominoElement) {
@@ -917,7 +997,7 @@ export abstract class Renderer {
   }
 
   private applyOtherTreatments(parsoidDoc: DominoElement, dump: Dump, articleId: string) {
-    this.removeIframeTags(parsoidDoc)
+    this.processIframeTags(parsoidDoc)
 
     if (dump.nodet) {
       this.removeCitations(parsoidDoc)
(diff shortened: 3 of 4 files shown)


Markus-Rost (changes requested):
You are currently discarding the size of the iframe (`width` and `height` attribute) and replacing it a generic div. You should keep the existing `width` and `height` attributes if possible.
Any existing classes and inline styles might also be worth keeping to ensure your div takes up the intended place of the iframe properly.

Otherwise this is a great way to solve the issue and might be a great idea for all iframes (replacing the iframe with an external link to the src), not just youtube embeds. 
Though of course with the special handling of youtube for the thumbnail etc.

KunalSiyag:
> You are currently discarding the size of the iframe (`width` and `height` attribute) and replacing it a generic div. You should keep the existing `width` and `height` attributes if possible. Any existing classes and inline styles might also be worth keeping to ensure your div takes up the intended place of the iframe properly.
> 
> Otherwise this is a great way to solve the issue and might be a great idea for all iframes (replacing the iframe with an external link to the src), not just youtube embeds. Though of course with the special handling of youtube for the thumbnail etc.

Good catch on the attributes, I'll update the code to retain width, height, class, and style.

Regarding the approach to make this generic for all iframes, my only concern is that some iframes embed local articles that we want to keep available offline. If we convert all of them to external links, we might break that offline functionality. WDYT?

Markus-Rost:
> Regarding the approach to make this generic for all iframes, my only concern is that some iframes embed local articles that we want to keep available offline. If we convert all of them to external links, we might break that offline functionality. WDYT?

Turning them into an external link, even if they are technically part of the ZIM as well, would still be better than the current version which just deletes all iframes

benoit74:
Plus we already have code in place to know if a given URL is going to be part of the ZIM or not, this is what we use when rewriting `<a href=...>` tags, and probably also `` (not sure about that one) and so on ; we can use same logic for iframe to detect internal links (and rewrite them correctly) and external links.

KunalSiyag:
> Plus we already have code in place to know if a given URL is going to be part of the ZIM or not, this is what we use when rewriting `<a href=...>` tags, and probably also `` (not sure about that one) and so on ; we can use same logic for iframe to detect internal links (and rewrite them correctly) and external links.

I have modified the code to support the external links while keeping the attributes same. Trying to recreate the build in my system to check it now.

KunalSiyag:
I have tried to check if it is working by fetching the article at [link](https://wiki.teamfortress.com/wiki/Meet_the_Medic_(taunt)) and it works well

However, we need to enhance how it handles lazy-loaded iframes—specifically, the div elements containing scripts that only render the iframe upon being clicked.

res/style.css
```diff
@@ -115,3 +115,49 @@ summary.section-heading {
...
+
+
+
+.external-video-card {
+    margin: 1em 0;
+    border: 1px solid #a2a9b1;
+    background-color: #f8f9fa;
+}
+
+.external-video-link {
```
Markus-Rost: The CSS classes should probably use the `mwo-` prefix to avoid conflicts with actual elements on the wiki

src/renderers/abstract.renderer.ts
```diff
@@ -680,13 +680,108 @@ export abstract class Renderer {
...
+  private createExternalIframePlaceholder(parsoidDoc: DominoElement, iframe: DominoElement, originalUrl: string): DominoElement {
+    const card = parsoidDoc.createElement('div')
+    this.copyNodeAttributes(iframe, card)
+    const iframeClass = card.getAttribute('class')
+    card.setAttribute('class', iframeClass ? `${iframeClass} external-video-card` : 'external-video-card')
+
+    const link = parsoidDoc.createElement('a')
+    link.setAttribute('class', 'external external-video-link')
+    link.setAttribute('href', originalUrl)
+    link.setAttribute('onclick', "return window.confirm('You are about to leave this ZIM and open an external website. Continue?')")
```
Markus-Rost: The onclick alert seems excessive, such warnings are for the reader to implement. The `a.external` element should already receive the normal external link icon from the existing wiki styles, like all other external links.

Markus-Rost (changes requested):

res/style.css
KunalSiyag: I missed that. Thanks for pointing out. Will rephrase the classes.

src/renderers/abstract.renderer.ts
KunalSiyag: Yeah I guess alerts should be handled by reader. I initally had a raw output that didn't alert when clicked on the image, so I figured an alert would be good when clicked on the thumbnail by mistake.

Will remove it and make it behave normally. Before I proceed though:

This is the **current setup** wherein we also display a textcontent just below the thumbail (it could disorder the layout in some cases ):

image
<hr>

This was the previous one (Adding the overlay text was breaking the CSS because JS scripts were not working). 



image

What I am thinking of now is to remove the textcontent since the icon alone signfies that clicking the  thumbnail will lead to a external source.

What do you think about that?

benoit74: I indeed agree that we do not want an alert to be implemented in the ZIM, ZIM reader will handle that based on user prefs, and I agree we probably do not need any textcontent, the icon alone is our convention so far (and allows for simpler logic no matter what the iframe is).

src/renderers/abstract.renderer.ts
```diff
@@ -680,13 +680,108 @@ export abstract class Renderer {
...
+      link.textContent = 'Watch on YouTube (external)'
+      card.appendChild(link)
+      return card
+    }
+
+    const thumbnailWrapper = parsoidDoc.createElement('div')
+    thumbnailWrapper.setAttribute('class', 'external-video-thumbnail')
+
+    const thumbnail = parsoidDoc.createElement('img')
+    thumbnail.setAttribute('src', `https://img.youtube.com/vi/${videoId}/hqdefault.jpg`)
```
benoit74, line 868: The fact the thumbnail is external is an issue for me.

I recommend to either drop the thumbnail feature for now (would simplify maintenance as well) or to implement logic to download the thumbnail and add it inside the ZIM.

KunalSiyag: I don't think removing the thumbnail would do any good as we would have to put a placeholder in place eitherway with something to fill that gap.
So I will proceed to implement a logic to download and store the thumbnail internally.

src/renderers/abstract.renderer.ts
KunalSiyag: Fair enough, I am also of the same mind.

KunalSiyag:
Verified by generating a ZIM locally.

The YouTube placeholder rendering and offline thumbnail download are working as expected.

Example screenshot:
image

One edge case I’m unsure about: when the videoId is successfully extracted but the thumbnail image fails to download. In that scenario the placeholder may render without the image. Not yet tested though.

Should this fall back to the generic external link placeholder instead, or is the current behavior acceptable?

benoit74:
Current behavior regarding placeholder which fails to download is acceptable. This is a generic issue for all images which fails to download.

KunalSiyag:
Got it, thanks for confirming! That makes sense. I’ll leave it as is then.

benoit74:
I tried this on cyclowiki.org and besides that you need to rebase + run without js (`--javaScript=none`) to work around other bugs, it works quite well.

Main concern is that img does not fit inside parent size:

image

vs

Screenshot 2026-03-09 at 17 16 16

would it make sense to always set this to the image:

```css
  object-fit: contain;
  width: 100%;
  height: 100%;
```

(not sure we need the three)

KunalSiyag:
Thanks for the suggestion. the `--javaScript=none` tip helped a lot. I was running into the same issue with older MediaWiki instances where the dump fails while fetching the startup module:

```
Failed to get module with url https://cyclowiki.org/w/load.php?lang=ru&modules=startup&only=scripts&raw=1&skin=vector
TypeError: Cannot read properties of null (reading '1')
at downloadAndSaveStartupModule (dump.ts:207)
```

Running without JavaScript works around that problem on Cyclowiki.

Regarding the layout issue, I’ll update the styling though I think setting `height: auto` would avoid stretching the image.

Markus-Rost:
Arg, looks like I did a bad job in my JS PR to test for older Mw versions. Sorry about that, I've created #2650 as a patch for the startup issue.

KunalSiyag:
Quick update: the thumbnail stretching issue appears specific to certain wiki outputs (e.g., Cyclowiki), but the site is currently down/unreliable, so I’m unable to validate against it right now after the fix. Will find some other ways to test.

KunalSiyag:
While testing this across a few different sites, I noticed that many pages use lazy-loading patterns where the <iframe> is not present in the initial HTML. Instead, a preview element (often a div with a background image or an img thumbnail) is rendered first and the actual iframe is injected via JavaScript only when the user clicks on it. (ex [lurkmore](https://lurkmore.media/YouTube))

Since JavaScript is disabled/stripped in mwoffliner output, those iframes never appear in the DOM, which means the current logic cannot detect them and they remain as non-functional placeholders.

I’m thinking about extending the detection logic to also look for these lazy video preview patterns and apply the same thumbnail → external link replacement strategy used for iframes. That way the behavior stays consistent even when the iframe is generated dynamically.

This is probably out of scope for this PR, but I wanted to mention it since it came up during testing.

benoit74:
@KunalSiyag can you please complete this effort before moving on to other issues?

KunalSiyag:
> @KunalSiyag can you please complete this effort before moving on to other issues?

Yes, I’m already working on it. I’ll complete this first and share an update soon.

KunalSiyag:
The issue has been resolved, it was due to the wrapper div being implemented not the css properties. However, the external link icon for the anchor tag is right-aligned, causing it to go out of frame in this case. This is due to the .external style which was already being used:
```
 {
    background-image: url(data:image/svg+xml,...); /* external link icon SVG */
    background-repeat: no-repeat;
    background-position: center right;
    padding-right: 13px;
}
```
image
&nbsp&nbsp&nbsp&nbsp&nbsp
image

Changelog
```diff
@@ -1,4 +1,5 @@
 Unreleased:
+* FIX: Support third-party hosted (Youtube) videos (@kunalsiyag #2481)
```
benoit74: This PR is now doing more than only handling Youtube, it handles all iframes, we should update that changelog

Changelog
```diff
@@ -9,6 +10,7 @@ Unreleased:
 * FIX: Improve mwUrl and adminEmail input validation and expanding test coverage for input sanitation (@RishabhThakur-19 #2657)
 * DEL: Remove deprecated CLI parameters `mwWikiPath`, `mwIndexPhpPath`, `keepEmptyParagraphs` (@Markus-Rost #2489)
 
+
```
benoit74: Remove this double blank line

res/style.css
```diff
@@ -115,3 +115,12 @@ summary.section-heading {
 body[data-useragent*='KAIOS'] h1 {
     display: none !important;
 }
+
```
benoit74, line 118: No need for so many blank lines

src/renderers/abstract.renderer.ts
```diff
@@ -813,7 +888,7 @@ export abstract class Renderer {
   }
 
   private applyOtherTreatments(parsoidDoc: DominoElement, dump: Dump, articleId: string) {
```
benoit74, line 999: Probably worth to add a comment in caller (`processHtml`) that `applyOtherTreatments` needs to run before `treatMedias`

src/renderers/abstract.renderer.ts
```diff
@@ -681,14 +681,89 @@ export abstract class Renderer {
...
+
+    if (!videoId) {
+      link.textContent = 'View this content externally'
+      link.setAttribute('aria-label', 'View this content externally')
+      return link
+    }
+
+    const thumbnail = parsoidDoc.createElement('img')
+    this.copyNodeAttributes(iframe, thumbnail)
+    thumbnail.setAttribute('src', `https://img.youtube.com/vi/${videoId}/hqdefault.jpg`)
```
benoit74, line 868: Please add a comment explaining that we need to set online URL here because it will be processed afterwards to download it to the ZIM

res/style.css
KunalSiyag: Reduced the lines. Mistake on my part.

src/renderers/abstract.renderer.ts
KunalSiyag: That would be wise, Yes.

src/renderers/abstract.renderer.ts
KunalSiyag: Done with that.

PR #2630, opened 2026-02-21
add --customCss CLI option for injecting static CSS files

### LINKED ISSUE #2448: Automatically add "static" CSS scripts used on all wiki pages
We've just produced a ZIM of https://wiki.gentoo.org

This wiki is using a custom `Tyrian` skin which scraper does not support, so we've used fallback skin.

The result is however rather deceptive in terms of visual result: https://dev.library.kiwix.org/viewer#gentoo_en_all_maxi_2025-07

I suspect this is largely caused by the fact that many CSS does not comes from regular Mediawiki modules but from two "statically added" CSS scripts which seems to be used on all pages.

Many questions:
- is it possible to retrieve the list of such scripts? via API?
- is it sufficient to add them to fix most styling issues?
- should we do it systematically on all skins (because even someone using a supported skin like vector-2022 could add its own "static" CSS on top of it), or only when using fallback skin?

ziaddevv:
fixes #2448

**What this PR does**
Adds a --customCss option to let users include their own CSS files when generating a ZIM. This fixes pages that look broken because some wikis use external CSS not captured by default

**How it works**
You just pass a comma-separated list of URLs:
`mwoffliner --customCss=cssUrl,cssURL2`

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -25,6 +25,7 @@ import {
   downloadAndSaveModule,
   downloadAndSaveStartupModule,
   getModuleDependencies,
+  downloadAndSaveCustomCss,
   genCanonicalLink,
   genHeaderCSSLink,
   genHeaderScript,
@@ -56,6 +57,7 @@ import { articleListHomeTemplate, htmlRedirectTemplateCode } from './Templates.j
 import { downloadFiles, saveArticles } from './util/saveArticles.js'
 import { getCategoriesForArticles, trimUnmirroredPages } from './util/categories.js'
 import urlHelper from './util/url.helper.js'
+import { parseCustomCssUrls, customCssUrlToFilename } from './util/customCss.js'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
@@ -103,6 +105,7 @@ async function execute(argv: any) {
     forceRender,
     forceSkin,
     langVariant,
+    customCss,
   } = argv
 
   let { articleList, articleListToIgnore } = argv
@@ -114,6 +117,14 @@ async function execute(argv: any) {
   // TODO: Move it to sanitaze method
   if (articleList) articleList = String(articleList)
   if (articleListToIgnore) articleListToIgnore = String(articleListToIgnore)
+
+  // Parse --customCss and populate Downloader so renderers can use the list
+  if (customCss) {
+    Downloader.customCssUrls = parseCustomCssUrls(String(customCss))
+    if (Downloader.customCssUrls.length > 0) {
+      logger.log(`Custom CSS URLs configured: ${Downloader.customCssUrls.join(', ')}`)
+    }
+  }
   const publisher = _publisher || config.defaults.publisher
 
   // TODO: Move it to sanitaze method
@@ -435,6 +446,23 @@ async function execute(argv: any) {
     if (!mainPage) {
       logger.log('Checking Main Page rendering')
       await createIndexPage(dump, zimCreator, true)
+    }
+
+    // Download and save custom CSS files
+    if (Downloader.customCssUrls.length > 0) {
+      logger.log(`Downloading ${Downloader.customCssUrls.length} custom CSS file(s)`)
+      const cssErrors: string[] = []
+      for (const cssUrl of Downloader.customCssUrls) {
+        try {
+          const filename = customCssUrlToFilename(cssUrl)
+          await downloadAndSaveCustomCss(zimCreator, cssUrl, filename)
+        } catch (err) {
+          cssErrors.push(`  - Failed to download [${cssUrl}]: ${err}`)
+        }
+      }
+      if (cssErrors.length > 0) {
+        throw new Error(`Failed to download custom CSS file(s):\n${cssErrors.join('\n')}`)
+      }
     }
 
     logger.log('Getting articles')
--- a/src/renderers/action-parse.renderer.ts
+++ b/src/renderers/action-parse.renderer.ts
@@ -7,6 +7,7 @@ import { genCanonicalLink, genHeaderScript, genHeaderCSSLink, getStaticFiles, ge
 import MediaWiki from '../MediaWiki.js'
 import { htmlVectorLegacyTemplateCode, htmlVector2022TemplateCode, htmlFallbackTemplateCode, javaScriptTemplateCode } from '../Templates.js'
 import Downloader, { DownloadError } from '../Downloader.js'
+import { customCssUrlToFilename } from '../util/customCss.js'
 import Gadgets from '../Gadgets.js'
 import { extractJsConfigVars, extractBodyCssClass, extractHtmlCssClass, getMainpageTitle } from '../util/articles.js'
 
@@ -85,6 +86,14 @@ export class ActionParseRenderer extends Renderer {
       .replace('__ARTICLE_JS_MODULES__', jsonStringify(jsDependenciesList))
       .replace('__ARTICLE_CSS_STATE__', jsonStringify(articleCssState))
 
+    // Generate custom CSS links from --customCss option
+    const customCssLinks = (Downloader.customCssUrls || [])
+      .map((cssUrl: string) => {
+        const filename = customCssUrlToFilename(cssUrl)
+        return genHeaderCSSLink(config, filename, articleId, config.output.dirs.res)
+      })
+      .join('\n    ')
+
     const htmlTemplateString = this.#htmlTemplateCode()
       .replace(/__ARTICLE_LANG_DIR__/g, articleLangDir)
       .replace('__ARTICLE_CANONICAL_LINK__', genCanonicalLink(config, MediaWiki.webUrl.href, articleId))
@@ -94,6 +103,7 @@ export class ActionParseRenderer extends Renderer {
       .replace('__ARTICLE_JS_STARTUP__', articleJsStartup)
       .replace('__ARTICLE_CSS_AFTER_META__', articleCssAfterMeta)
       .replace('__ARTICLE_CSS_NOSCRIPT__', articleCssNoscript)
+      .replace('__CUSTOM_CSS__', customCssLinks)
       .replace(/__ASSETS_DIR__/g, config.output.dirs.assets)
       .replace(/__RES_DIR__/g, config.output.dirs.res)
       .replace(/__MW_DIR__/g, config.output.dirs.mediawiki)
--- a/src/util/customCss.ts
+++ b/src/util/customCss.ts
@@ -1,0 +1,66 @@
+import * as fs from 'fs'
+import crypto from 'crypto'
+
+export function parseCustomCssUrls(rawValue: string): string[] {
+  if (!rawValue) {
+    return []
+  }
+
+  const result: string[] = []
+  const errors: string[] = []
+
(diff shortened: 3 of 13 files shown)


ziaddevv:
Hey! Just an update: I've expanded the --customCss argument so it can now read URLs directly from a local .csv
 or .txt file (using the exact same fs.existsSync pattern we already use for --articleList).

 I also fixed a tiny logging bug where invalid URL warnings were being swallowed because they fired before the logger was officially initialized. I'll post once i finish

ziaddevv:
Update : I’ve finished the remaining changes and the PR is now ready for review

src/renderers/action-parse.renderer.ts
```diff
@@ -53,13 +54,22 @@ export class ActionParseRenderer extends Renderer {
       .map((oneCssDep: string) => genHeaderCSSLink(config, oneCssDep, articleId, config.output.dirs.mediawiki))
       .join('\n    ')
 
+    // Generate custom CSS links from --customCss option
+    const customCssLinks = (Downloader.customCssUrls || [])
+      .map((cssUrl: string) => {
+        const filename = customCssUrlToFilename(cssUrl)
+        return genHeaderCSSLink(config, `custom.${filename}`, articleId, config.output.dirs.mediawiki)
```
Markus-Rost: Those CSS files probably shouldn't be in the mediawiki directory. That directory is for files directly from the wiki and the css modules from the wiki can have any names. It wouldn't seem impossible for a wiki to add a custom css module named using the same `custom.` format. Using the `_res_` directory with the `custom.` prefix might be more appropriate.

src/util/customCss.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function customCssUrlToFilename(url: string): string {
+  try {
+    const pathname = new URL(url).pathname
+    const filename = pathname.split('/').pop() || 'custom_style'
+    return filename.replace(/\.css$/, '').replace(/[^a-zA-Z0-9._-]/g, '_')
+  } catch {
+    return 'custom_style'
+  }
+}
```
Markus-Rost, line 66: This function is likely return the same filename for different urls. Later code has no handling for such conflicts.

src/util/customCss.ts
```diff
@@ -0,0 +1,62 @@
+import * as logger from '../Logger.js'
+import * as fs from 'fs'
+
+export function parseCustomCssUrls(rawValue: string): string[] {
+  if (!rawValue) {
+    return []
+  }
+  const seen = new Set<string>()
+  const result: string[] = []
```
Markus-Rost: `seen` and `result` always have the same content, this should just be one variable.

src/util/customCss.ts
ziaddevv: > This function is likely return the same filename for different urls. Later code has no handling for such conflicts.

Fixed ,  the function now prefixes each filename with an 8-character MD5 hash of the full URL, guaranteeing uniqueness even when two URLs share the same basename like both ends with style.css. Added a test for this case too

src/util/customCss.ts
ziaddevv: > `seen` and `result` always have the same content, this should just be one variable.

so I dropped the Set and now use a single result array with result.includes() for deduplication. I realized that The O(1) lookup of a Set is irrelevant at this small scale that is an extra optimization

src/renderers/action-parse.renderer.ts
ziaddevv: > Those CSS files probably shouldn't be in the mediawiki directory. That directory is for files directly from the wiki and the css modules from the wiki can have any names. It wouldn't seem impossible for a wiki to add a custom css module named using the same `custom.` format. Using the `_res_` directory with the `custom.` prefix might be more appropriate.

done I moved it

src/util/customCss.ts
```diff
@@ -0,0 +1,66 @@
+import * as logger from '../Logger.js'
+import * as fs from 'fs'
+import crypto from 'crypto'
+
+export function parseCustomCssUrls(rawValue: string): string[] {
```
benoit74, line 4: Any error while parsing the custom CSS URL should fail the scrape. User has asked for a custom URL, if the scraper does not achieve to pull it, we better fail the scraper early so that user has a chance to fix that. It would be very deceptive to let the scraper run (potentially for hours) to finally have a non-working ZIM because custom CSS is still missing. Probably worth to check all parts before exiting, so that if we have multiple parts and all are bad, we display all errors before exiting.

src/util/customCss.ts
```diff
@@ -0,0 +1,66 @@
...
+        logger.warn(`Skipping invalid CSS URL [${trimmed}]: unsupported protocol`)
+        continue
+      }
+    } catch {
+      logger.warn(`Skipping invalid CSS URL [${trimmed}]`)
+      continue
+    }
+
+    if (result.includes(trimmed)) {
+      logger.log(`Ignoring duplicate CSS URL [${trimmed}]`)
```
benoit74: This should be an error as well, there is no reason for the user to give duplicate URLs, if so we should let him fix

src/util/customCss.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return result
+}
+
+export function customCssUrlToFilename(url: string): string {
+  try {
+    const pathname = new URL(url).pathname
+    const basename = pathname.split('/').pop() || 'custom_style'
+    const stem = basename.replace(/\.css$/, '').replace(/[^a-zA-Z0-9._-]/g, '_')
```
benoit74: Why dropping the `.css` suffix? I would prefer we ensure there is always a `.css` suffix even if it was missing first-hand ; this brings lots of clarity in `zimdump` outputs for instance

src/util/dump.ts
```diff
@@ -159,6 +159,20 @@ export async function downloadAndSaveModule(zimCreator: Creator, module: string,
   }
 }
 
+// Download a custom CSS file and save it into the ZIM.
+export async function downloadAndSaveCustomCss(zimCreator: Creator, cssUrl: string, filename: string): Promise<void> {
+  try {
+    logger.log(`Downloading custom CSS [${cssUrl}]`)
+    const { data: cssBody } = await Downloader.request<string>({ url: cssUrl, method: 'GET', responseType: 'text', timeout: 30000 })
```
benoit74: please prefer to use `Downloader.downloadContent` which is more appropriate in that it ensures to use the same logic we use to download any other content (including automated retry rules, ...)

src/util/dump.ts
```diff
@@ -159,6 +159,20 @@ export async function downloadAndSaveModule(zimCreator: Creator, module: string,
...
+    const processedCss = processStylesheetContent(cssUrl, '', cssBody)
+    const zimPath = cssPath(`custom.${filename}`, config.output.dirs.res)
+    await zimCreatorMutex.runExclusive(() => zimCreator.addItem(new StringItem(zimPath, 'text/css', null, { FRONT_ARTICLE: 0 }, processedCss)))
+    logger.log(`Saved custom CSS [${cssUrl}] at ${zimPath}`)
+  } catch (err) {
+    logger.warn(`Failed to download custom CSS [${cssUrl}]: ${err}`)
```
benoit74: This should be a fatal error as well

src/mwoffliner.lib.ts
```diff
@@ -455,6 +466,15 @@ async function execute(argv: any) {
       }),
     )
 
+    // Download and save custom CSS files
```
benoit74: We should do it way earlier, at least before the "Get articles" phase which could take very long. If for any reason we fail to get the custom CSS, we need to raise the error to the scraper user asap.

benoit74 (changes requested):
See inline comment + you need to add the new `customCss` CLI argument to `offliner-definition.json` (this file is used by Zimfarm to know which argument is needed)

RishabhThakur-19:
@ziaddevv if you want I can help you in resolving the errors and requested changes ( because I mistakenly started working on this issue too ).let me know what you think

src/util/customCss.ts
Markus-Rost: The catch still always returns the same file name.

benoit74: And I barely see a reason for a try/catch block here. I barely see a reason why we would not achieve to run this logic, and should this happen I would prefer we fail the scrape rather and fix the logic rather than having a weird fallback logic apply.

ziaddevv:
> @ziaddevv if you want I can help you in resolving the errors and requested changes ( because I mistakenly started working on this issue too ).let me know what you think

Hi @RishabhThakur-19 , I'm always open to collaborating even on simple issues as it usually leads to better results
If you have any comments or think we can come up with a better solution, please feel free to let me know
I'm about to post the update

src/util/dump.ts
ziaddevv: > please prefer to use `Downloader.downloadContent` which is more appropriate in that it ensures to use the same logic we use to download any other content (including automated retry rules, ...)

I missed this one , thanks for referring

ziaddevv:
Thanks @Markus-Rost and @benoit74 ,  all feedback addressed

My initial approach used silent warnings and fallbacks to keep the scrape running even when something went wrong. Based on your feedback I've moved to a strict fail-fast philosophy all errors are collected and reported at once, and there are no silent fallbacks anywhere in the custom CSS flow

benoit74:
Nota: the fact that `customCss` is declared as `type: string` in offliner-defintion.json is intentional on my side. We might have used `type: blob` and `kind: css` but I'm not sure at all this will be the core usage in the Zimfarm (input our own CSS) + I'm quite sure we might need to input multiple comma-separated URLs which is not possible with `type: blob` ATM.

src/util/dump.ts
```diff
@@ -159,6 +159,16 @@ export async function downloadAndSaveModule(zimCreator: Creator, module: string,
   }
 }
 
+// Download a custom CSS file and save it into the ZIM.
+export async function downloadAndSaveCustomCss(zimCreator: Creator, cssUrl: string, filename: string): Promise<void> {
+  logger.log(`Downloading custom CSS [${cssUrl}]`)
+  const { content: cssBody } = await Downloader.downloadContent(cssUrl, 'css')
+  const processedCss = processStylesheetContent(cssUrl, '', cssBody.toString())
+  const zimPath = cssPath(filename, config.output.dirs.res)
```
Markus-Rost, line 216: `processStylesheetContent` stores any files inside the CSS in the `_mw_` directory while also assuming that this is the same directory the css file being processed is stored at. 
While rewriting `processStylesheetContent` to accept the directory as a parameter would be possible, this would risk us having duplicate files between `_mw_` and `_res_` directory.

There also seems to be a bigger underlying issue here caused by `processStylesheetContent` using a different path logic compared to normal `` elements.
This can result in:
a) the same file being included twice in the ZIM, once at `_assets_` using by an `` element and once in `_mw_` used in a css file.
b) name conflicts between files using css with the same base name but different urls
I'll create a followup issue for this in a moment

Markus-Rost: I've created #2647

Markus-Rost: Once #2648 is merged to fix the `processStylesheetContent` issue, no further changes in this PR will be needed.

benoit74:
I've squashed + rebased on top of main, waiting for tests results and I will merge hopefully.

PR #2631, opened 2026-02-25
fix: resize images to HTML-defined display width (#2192)

### LINKED ISSUE #2192: Detect image size from HTML attribute instead of URL regex and recompress
Following https://phabricator.wikimedia.org/T360589

Wikimedia will start to serve images only at fixed width to save storage on their cache. 

For instance, if the Wikitext requests the image thumbnail at 900px, in fact the image served will be at 1000px (because this is the smaller largest resolution available among the fixed set of resolutions they have chosen).

The consequence for us is that the image URL will contain this 1000px information, and we will hence retrieve and store image at this resolution, while in fact we could know that it is not used at any resolution larger than 900px.

Do we want to modify the scraper to use this 900px information and automatically resize the image retrieved from the mediawiki ?

KunalSiyag:
Fixes #2192

Images are resized to their actual display width (from HTML `width` attributes) before compression instead of being stored at full resolution. This reduces ZIM size without visible quality loss.

### Changes

- **`abstract.renderer.ts`**
  - Attach `width` from `getRequestedImageWidths` to `imageDependencies`

- **`saveArticles.ts`**
  - Prefer HTML-derived `dep.width` for deduplication
  - Pass `width` to `downloadContent`

- **`Downloader.ts`**
  - Accept `requestedWidth`
  - Resize via `sharp` prior to `imagemin`
  - Introduce width-aware S3 cache keys (`webp-w300`)
  - Prevent cache collisions across resize variants

- **`processHtml.test.ts`**
  - Regression test for images missing `width`

### Design Decisions

- Skip GIF resizing (sharp removes animation frames)
- `withoutEnlargement: true` (downscale only)
- Graceful fallback on resize failure
- Width-aware S3 cache versioning

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -586,7 +586,7 @@ class Downloader {
   public async getJSON<T>(_url: string): Promise<T> {
     const url = urlHelper.deserializeUrl(_url)
     return new Promise<T>((resolve, reject) => {
-      this.backoffCall(this.getJSONCb, url, 'json', (err: any, val: any) => {
+      this.backoffCall(this.getJSONCb, url, 'json', undefined, (err: any, val: any) => {
         if (err) {
           const httpStatus = (err.response && err.response.status) || err.httpReturnCode
           logger.info(`Failed to get [${url}] [status=${httpStatus}]`)
@@ -618,7 +618,12 @@ class Downloader {
     return this.request({ url, data, method: 'POST', ...config })
   }
 
-  public async downloadContent(_url: string, kind: DonwloadKind, retry = true): Promise<{ content: Buffer | string; contentType: string; setCookie: string | null }> {
+  public async downloadContent(
+    _url: string,
+    kind: DonwloadKind,
+    retry = true,
+    requestedWidth?: number,
+  ): Promise<{ content: Buffer | string; contentType: string; setCookie: string | null }> {
     if (!_url) {
       throw new Error(`Parameter [${_url}] is not a valid url`)
     }
@@ -638,9 +643,9 @@ class Downloader {
           }
         }
         if (retry) {
-          this.backoffCall(this.getContentCb, url, kind, cb)
+          this.backoffCall(this.getContentCb, url, kind, requestedWidth, cb)
         } else {
-          this.getContentCb(url, kind, cb)
+          this.getContentCb(url, kind, requestedWidth, cb)
         }
       })
     } catch (err) {
@@ -687,7 +692,7 @@ class Downloader {
     return articleDetails
   }
 
-  private getJSONCb = <T>(url: string, kind: DonwloadKind, handler: (...args: any[]) => any): void => {
+  private getJSONCb = <T>(url: string, kind: DonwloadKind, _requestedWidth: number | undefined, handler: (...args: any[]) => any): void => {
     logger.info(`Getting JSON from [${url}]`)
     this.request<T>({ url, method: 'GET', ...this.jsonRequestOptions })
       .then((val) => {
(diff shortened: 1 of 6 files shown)


KunalSiyag:
### Verification

Behaviour validated by generating ZIM files before and after the change
with identical inputs and comparing extracted assets.

**Method**

- Same article list and flags
- Assets extracted via `zimdump dump`
- Sizes compared using `du` / `diff`

**Result**

Assets are identical; multiple images show reduced byte size.

Examples:

- 10_Downing_Street.jpg — 11342 → 6276 bytes  
- Aerial_Tower_of_London.jpg — 16880 → 12106 bytes  
- Canary_Wharf.Reza_04.jpg — 5680 → 2110 bytes  
- Notting_hill_carnival.jpg — 28462 → 15270 bytes  

Unchanged images are expected, confirming selective resizing.

**Full comparison attached:**

[old_sizes.txt](https://github.com/user-attachments/files/25557469/old_sizes.txt) [new_sizes.txt](https://github.com/user-attachments/files/25557473/new_sizes.txt)

src/Downloader.ts
```diff
@@ -596,10 +596,14 @@ class Downloader {
             resolve(val)
           }
         }
+        // Wrap getContentCb in a closure that captures requestedWidth, because
+        // backoff.call() treats the last positional arg as the callback —
+        // we can't append extra args after it.
+        const handler = (u: string, k: DonwloadKind, h: any) => this.getContentCb(u, k, h, requestedWidth)
```
benoit74: Why not changing `getContentCb` signature to `async (url: string, kind: DonwloadKind, requestedWidth?: number, handler: any)` (keeping handler last)? This code is already extremely complex to read, adding a closure to a handler callback makes it unnecessary complex I think

src/Downloader.ts
```diff
@@ -733,12 +753,18 @@ class Downloader {
     }
   }
 
-  private async downloadImage(url: string, handler: any) {
+  private async downloadImage(url: string, handler: any, requestedWidth?: number) {
+    // Build a width-aware cache version token so the same URL at different
+    // requested widths does not reuse an undersized cached image.
+    const cacheVersion = this.webp
```
benoit74: I'm a bit concerned about this approach because it means we might end-up with 10s of versions of the same image in the cache, leading to increase cache size / costs.

But I have to agree that I do not see a better approach

KunalSiyag: Yes though as per the dedup logic in `saveArticles.ts` , a single scrape run typically produces only one cached version per image (the max width), not multiple versions.
Multiple scrape runs with different article selections could produce additional versions, but it will be well limited to the number of distinct width that are present.

KunalSiyag:
> Thank you! I don't achieve to understand how the scraper will handle situation where one article request the image at a given resolution, and another article request the same image at another resolution. Can you please explain this?
> 
> Few other remarks inline.
> 
> Note that CI is currently broken and I have to fix this before we can merge any open PR.

Would be happy to clarify. As per my understanding, we have a logic embedded in the `saveArticles` (`currentDepIsHigherRes`)  that says to keep the file path of the highest resolution image possible when the same URL appears across multiple article with different width  (e.g., Article A wants 200px, Article B wants 400px), the maximum width wins. Only one entry per image path is stored in filesToDownload, and it keeps the highest resolution. So the image gets downloaded and resized to 400px, and both articles share that same file and are displayed appropriately.

Please let me know if I can clarify anything further.

Also, I completely understand about CI , no worries at all. I’ll continue exploring other issues in the meantime. I’m genuinely enjoying working on this project and learning a lot through the process.

src/Downloader.ts
KunalSiyag: This was my original thought process too, but I initially used a closure because I wanted to avoid changing the signatures of shared methods like` backoffCall` and `getJSONCb` (which also uses backoffCall but doesn't need a width parameter).

However, I agree that avoiding the closure makes the code much cleaner overall. I have reordered the signature to `(url, kind, requestedWidth?, handler) `as you suggested, and updated the dependent functions (passing `undefined` for width in the JSON call path).

benoit74:
> Please let me know if I can clarify anything further.

OK, got it, forgot about this bit.

KunalSiyag:
> Please fix prettier issue, the e2e test looks like transient issue and the unit-test failing is the usual culprit on external PRs AFAIK

Prettified the Downloader file. Also it was a small change, still though I squashed the commits and force-pushed with --force-with-lease. Is this the preferred workflow for small fixes like this?

benoit74:
> Also it was a small change, still though I squashed the commits and force-pushed with --force-with-lease. Is this the preferred workflow for small fixes like this?

For such small changes, since my review was already mostly over, and since we have important changes in `main` branch, yes this is ok.

benoit74:
@KunalSiyag our policy forbids merge commit from `main` branch to feature branch. I've made a rebase and will wait for CI checks to merge

benoit74:
@KunalSiyag please fix code quality issue(s)

KunalSiyag:
@benoit74 When I ran `npm run format-check`, it reported that there were no files to prettify, so it looks like the earlier push may not have included those changes. I’ve pushed them again now.

Also, the malformed JSON retry issues have been addressed in #2658 and should be resolved once that PR is merged.

benoit74:
@KunalSiyag I tried to rebase this PR but looks like I did something crazy, can you please have a look?

KunalSiyag:
> @KunalSiyag I tried to rebase this PR but looks like I did something crazy, can you please have a look?

That was a mistake on my part. I didn't check enough that the call on `backoffCall` method in the `download.retry.test.ts` needed to have 5 arguments since  the implementation of it in the `downloader.ts` file was modified to also include a `requestedwidth` argument. Fixed that and tested the unit cases locally, gtg.

On a side note, I again ran the `npm run format-check `but didn't find any discrepancy.

KunalSiyag:
It seems that in [23f0fd8](https://github.com/openzim/mwoffliner/commit/23f0fd817e0fcede670afb3808eaa038f95d0b80) branch, some of my commits have been dropped possibly due to conflict resolution. I will recover them locally, rebase on the latest branch, and reapply.

benoit74:
Very sorry if I dropped some or your commits, my bad

KunalSiyag:
> Very sorry if I dropped some or your commits, my bad

No worries at all, I think I also mixed in a wrong commit on my side, one of mine had the youtube message in it. I’ll clean up the branch and reapply only the relevant changes. Learning a lot by debugging what went wrong.

KunalSiyag:
This one should work. Took the latest branch, cherry picked the branch with the update and solved the merge conflicts.

benoit74:
Failing CI test is not relevant, it work on my end, flaky test.

PR #2641, opened 2026-03-08
feat: Use <meta name="description"> to populate ZIM LongDescription

### LINKED ISSUE #2384: Use `<meta name="description" content="">` DOM node to fill ZIM `LongDescription` Metadata
Per default (so if not specified differently) and if available on the MediaWiki main page, I think it would be good to take the value to fill the ZIM `LongDescription` metadata (which would be otherwise empty).

Sigmabrogz:
## Summary
Uses the `description` meta tag from the main page to populate the ZIM `LongDescription` metadata when it's not provided explicitly.

## Problem
Closes #2384
Currently, the `LongDescription` metadata defaults to empty if not specified via `--customZimLongDescription`.

## Solution
If `customZimLongDescription` is not set and a `mainPage` exists, we fetch the main page using `Downloader.downloadContent(mainPageUrl, 'data')`, parse it with `domino`, and extract the `<meta name="description" content="...">` tag to populate `LongDescription`.

## Testing
- Built successfully: `npm run build`
- Tests pass: `npm test`

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -229,6 +229,22 @@ async function execute(argv: any) {
     }
   }
 
+  let defaultLongDescription = ''
+  if (!customZimLongDescription && mainPage) {
+    const mainPageUrl = MediaWiki.webUrl + encodeURIComponent(mainPage)
+    try {
+      const { content } = await Downloader.downloadContent(mainPageUrl, 'data')
+      const htmlStr = typeof content === 'string' ? content : content.toString()
+      const doc = domino.createDocument(htmlStr)
+      const metaDesc = doc.querySelector('meta[name="description"]')
+      if (metaDesc) {
+        defaultLongDescription = metaDesc.getAttribute('content') || ''
+      }
+    } catch (err: any) {
+      logger.warn(`Failed to fetch main page to extract description: ${err.message}`)
+    }
+  }
+
   MediaWiki.apiCheckArticleId = mwMetaData.mainPage
   await MediaWiki.hasCoordinates()
   await MediaWiki.hasActionParseApi()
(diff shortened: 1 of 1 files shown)


benoit74 (commented):
I'm really not a big fan of downloading the main page ; we know this can trigger bot detection quite easily (we download a web page without downloading any of its resources). And since this is needed only to fill the long description which is not that an important ZIM metadata (no reader display it AFAIK) ... looks like we are putting the scraper at risk of being even more often blocked for a minimum added value.

Something to discuss with @kelson42 obviously

Sigmabrogz:
Thank you for the feedback @benoit74! You raise a valid concern about bot detection risk.

I understand the trade-off here - the LongDescription metadata isn't critical enough to justify potentially triggering bot detection.

A few alternative approaches we could consider:
1. **Skip the feature entirely** - if the risk outweighs the benefit, I'm happy to close this PR
2. **Use existing data** - is there any other metadata already being fetched that could populate LongDescription?
3. **Make it opt-in** - add a flag like `--fetchLongDescription` that's off by default

What would you prefer? I'm happy to go with whatever makes sense for the project's stability.

Thanks for maintaining this project!

benoit74:
Thanks for being open-minded.

On a personal basis, I've expressed my doubts about the usefulness of this feature since quite long ago, for the exact reasons mentioned here. 

This is hence now only @kelson42 call, not mine.

benoit74:
@kelson42 up

Aunirbhan:
Hey @Sigmabrogz @benoit74 @kelson42, I opened #2676 before seeing this was already in progress, sorry about the duplicate.

I noticed that `<meta name="description">` is absent from many standard MediaWikis (I tested Wikipedia, Wiktionary). So this would silently return nothing on every major wiki.

The same data is fetched using (`action=query&prop=description`), which can also avoid the bot detection concern @benoit74 mentioned. The HTML scraping could be kept as a fallback for non-Wikimedia wikis that do set the meta tag.

What do you think about this approach?

benoit74:
@kelson42 up

benoit74:
Moving to draft until we get conclusion about whether this is a good idea to move this PR forward or not.

kelson42:
@Sigmabrogz Thank you for attempting to implement this issue. I just have closed the issue and will close this PR as there is no obvious solution to the problem and therefore the enhancement can not be brought!

PR #2642, opened 2026-03-08
Scrape only wikitext contentmodel articles by default

### LINKED ISSUE #2445: Scrape only `wikitext` contentmodel articles by default
Currently, the scraper push to the ZIM all articles, no matter what their contentmodel is.

I feel like this is wrong, by default the scraper should scrape only `wikitext` contentmodel, log the list of articles found with other contentmodels since it might help, and an option would allow to add other contentmodel.

A bit like we do for namespaces, where by default we do not add all namespaces.

triemerge:
Fix #2445

- Filter articles by contentmodel during scraping, keeping only wikitext articles by default
- Add `--addContentModels` option to include additional content models (comma separated)
- Add `--scrapeAllContentModels` option to scrape all content models regardless
- Log skipped articles with their content model for transparency
- Add Changelog entry

### DIFF
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -96,7 +96,7 @@ async function saveToStore(
   }
 }
 
-export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], continueLimit?: number): Promise<void> {
+export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], allowedContentModels: string[] = ['wikitext'], continueLimit?: number): Promise<void> {
   // eslint-disable-next-line no-async-promise-executor
   return new Promise(async (resolve, reject) => {
     let totalArticles = 0
@@ -139,6 +139,20 @@ export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], conti
             articleIdsToIgnore = []
           }
           articleIdsToIgnore.push(...newArticlesToIgnore)
+        }
+
+        // Filter articles by content model (#2445)
+        if (allowedContentModels) {
+          const articlesWithWrongModel = Object.values(chunk.articleDetails)
+            .filter((a) => !allowedContentModels.includes(a.contentmodel))
+            .map((article) => article.title)
+          if (articlesWithWrongModel.length > 0) {
+            logger.info(`Skipping articles with non-matching content model: ${articlesWithWrongModel.join(', ')}`)
+            if (!articleIdsToIgnore) {
+              articleIdsToIgnore = []
+            }
+            articleIdsToIgnore.push(...articlesWithWrongModel)
+          }
         }
 
         if (articleIdsToIgnore) {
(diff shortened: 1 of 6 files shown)


src/util/mw-api.ts
```diff
@@ -141,6 +162,11 @@ export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], conti
           articleIdsToIgnore.push(...newArticlesToIgnore)
         }
 
+        // Filter articles by content model (#2445)
+        if (allowedContentModels) {
+          chunk.articleDetails = filterByContentModel(chunk.articleDetails, allowedContentModels)
```
Markus-Rost: Using a dedicated function here seems unnecessary and the function is more complex than it needs to be, this can just use the same logic as `articleIdsToIgnore` below (possibly even the same loop).

src/util/mw-api.ts
```diff
@@ -7,7 +7,23 @@ import RedisStore from '../RedisStore.js'
...
+      skipped.push(`${detail.title} (${detail.contentmodel})`)
+    }
+  }
+  if (skipped.length) {
+    logger.info(`Skipping articles with non-matching content model: ${skipped.join(', ')}`)
+  }
+  return filtered
+}
+
+export async function getArticlesByIds(articleIds: string[], log = true, allowedContentModels: string[] | null = null): Promise<void> {
```
Markus-Rost: This function takes a user provided list of pages, no filtering by content model needed here as it can be assumed that a listed page should always be included.

triemerge: Right, removed the filtering from getArticlesByIds entirely. If someone explicitly listed a page, it should get scraped regardless of its content model.

src/util/mw-api.ts
triemerge: Done. Dropped the separate function and moved the filtering into getArticlesByNS using the same pattern as articleIdsToIgnore. Also removed scrapeAllContentModels since listing the models in --addContentModels covers that use case.

src/util/mw-api.ts
```diff
@@ -96,7 +96,7 @@ async function saveToStore(
   }
 }
 
-export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], continueLimit?: number): Promise<void> {
+export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], continueLimit?: number, allowedContentModels: string[] = ['wikitext']): Promise<void> {
```
Markus-Rost: Move allowedContentModels before continueLimit in the parameter list and update any usages adding continueLimit.

triemerge: Done, moved it before continueLimit. Updated the test in mwApi.test.ts too since it was passing continueLimit.

triemerge:
Added it to offliner-definition.json right next to addNamespaces. Also handled the parameter reorder from the inline comment.

benoit74:
Could we get into a mediawiki where we do not want the `wikitext` content model? I'm really unsure it might exist into the wild, but if so we are "stuck" with the `--addContentModel` argument, we should better directly use `--allowedContentModels` which by default is `wikitext` and could be overridden to anything else (which must include wikitext if we want it). But this is annoying for the user if `wikitext` is always needed, because it means he must not forget to add wikitext to his list.

triemerge:
The current approach follows the same pattern as `addNamespaces` from the original issue. Switching to a full override would be more flexible but means users have to remember to include wikitext every time. Happy to change it if you and @Markus-Rost prefer that.

Markus-Rost:
I doubt we would ever want to not include the `wikitext` content model. Following the same logic as us always wanting to include the content namespaces.

PR #2643, opened 2026-03-08
Text overlaps on other other text or even images

### LINKED ISSUE #1004: Text overlaps on other other text or even images
Most issues can be seen in titles. Affects all range of Wikis.

http://library.kiwix.org/wikipedia_gsw_all_nopic_2019-06/A/Wikipedia:Houptsyte
GSW
http://library.kiwix.org/wikibooks_gl_all_nopic_2019-07/A/Portada
GL
http://library.kiwix.org/wikipedia_scn_all_nopic_2019-04/A/P%C3%A0ggina_principali
SCN
http://library.kiwix.org/wikinews_bs_all_nopic_2019-07/A/Po%C4%8Detna_strana
BS
http://library.kiwix.org/wikipedia_tt_all_nopic_2019-04/A/%D0%91%D0%B0%D1%88_%D0%B1%D0%B8%D1%82
TT
http://library.kiwix.org/wikinews_bs_all_novid_2019-07/A/Po%C4%8Detna_strana
BS
http://library.kiwix.org/wikipedia_eu_all_novid_2019-06/A/Azala
EU
http://library.kiwix.org/wikisource_pt_all_novid_2019-07/A/Wikisource:P%C3%A1gina_principal
PT
http://library.kiwix.org/wikiversity_fr_all_nopic_2019-07/A/Wikiversit%C3%A9:Accueil
FR
http://library.kiwix.org/granbluefantasy_en_all_all_novid_2018-10/A/Main_Page.html
EN
http://library.kiwix.org/wikipedia_tt_all_novid_2019-04/A/%D0%91%D0%B0%D1%88_%D0%B1%D0%B8%D1%82
TT
http://library.kiwix.org/wikipedia_vec_all_nopic_2019-04/A/Pajina_prinsipa%C5%82e
VEC
http://library.kiwix.org/wikipedia_it_all_nopic_2019-10/A/Pagina_principale
IT
http://library.kiwix.org/wikipedia_lez_all_nopic_2019-05/A/%D0%9A%D1%8C%D0%B8%D0%BB%D0%B8%D0%BD_%D1%87%D1%87%D0%B8%D0%BD
LEZ
http://library.kiwix.org/wikipedia_xal_all_nopic_2019-04/A/%D0%9D%D2%AF%D1%80_%D1%85%D0%B0%D0%BB%D1%85
XAL
http://library.kiwix.org/wikipedia_ce_all_novid_2019-04/A/%D0%9A%D0%BE%D1%8C%D1%80%D1%82%D0%B0_%D0%B0%D0%B3%D3%80%D0%BE
CE
http://library.kiwix.org/wikipedia_sr_all_nopic_2018-10/A/%D0%93%D0%BB%D0%B0%D0%B2%D0%BD%D0%B0_%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%B0.html
SR
http://library.kiwix.org/wikipedia_gan_all_nopic_2019-06/A/%E5%B0%81%E9%9D%A2
GAN
http://library.kiwix.org/wikisource_gl_all_novid_2019-07/A/Portada
GL
http://library.kiwix.org/wikipedia_fy_all_nopic_2019-06/A/Haadside
FY
http://library.kiwix.org/wikipedia_os_all_nopic_2019-04/A/%D0%A1%C3%A6%D0%B9%D1%80%D0%B0%D0%B3_%D1%84%D0%B0%D1%80%D1%81
OS
http://library.kiwix.org/wikipedia_stq_all_nopic_2019-03/A/Haudsiede
STQ
http://library.kiwix.org/wikipedia_pms_all_novid_2019-04/A/Intrada
PMS
http://library.kiwix.org/wikisource_pt_all_nopic_2019-07/A/Wikisource:P%C3%A1gina_principal
PT
http://library.kiwix.org/wikinews_ca_all_nopic_2019-07/A/Portada
CA
http://library.kiwix.org/wikipedia_pms_all_nopic_2019-04/A/Intrada
PMS
http://library.kiwix.org/wikipedia_ce_all_nopic_2019-04/A/%D0%9A%D0%BE%D1%8C%D1%80%D1%82%D0%B0_%D0%B0%D0%B3%D3%80%D0%BE
CE
http://library.kiwix.org/wikisource_fi_all_novid_2019-07/A/Etusivu
FI
http://library.kiwix.org/wikiversity_it_all_nopic_2019-07/A/Pagina_principale
IT
http://library.kiwix.org/wikipedia_frr_all_nopic_2019-06/A/Wikipedia:Hoodsid
FRR
http://library.kiwix.org/wikipedia_lo_all_novid_2019-05/A/%E0%BB%9C%E0%BB%89%E0%BA%B2%E0%BA%AB%E0%BA%BC%E0%BA%B1%E0%BA%81
LO

KunalSiyag:
Fix #1004 by preserving image layout in nopic mode using SVG placeholders

- Update treatImage() to replace img src with inline SVG placeholders instead of removing images
- Preserve img elements and dimensions in nopic
- Prevent image wrappers (figure/span) from being removed when dump.nopic is true
- Render meaningful alt text inside placeholder SVG, otherwise use transparent placeholder
- Do not add placeholder images to imageDependencies
- Keep existing math fallback and novid/nodet behavior unchanged
- Parsing of dimensions from inline style
- Reducing the fallback svg to 1*1 to handle images with no size metadata

Tests:
- Update media.treatment tests to expect img with data:image/svg+xml src and empty imageDependencies
- Update e2e nopic assertion to expect images without downloaded assets

Validation:

Made a zim file of https://it.wikiversity.org/wiki/Pagina_principale#/

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,6 @@
 Unreleased:
 * NEW: Scrape only wikitext contentmodel articles by default (@triemerge #2445)
+* FIX: In nopic mode, fixed layout issues by preserving image layout using SVG placeholders (@kunalsiyag #1004)
 * NEW: Add support for ResourceLoader based JavaScript (@Markus-Rost #2310)
 * CHANGED: Keep only ActionParse renderer (@benoit74 #2210)
 * CHANGED: Keep inline scripts if all JS is allowed (@Markus-Rost #2555)
--- a/src/renderers/abstract.renderer.ts
+++ b/src/renderers/abstract.renderer.ts
@@ -339,7 +339,7 @@ export abstract class Renderer {
   private treatImageFrames(dump: Dump, parsoidDoc: DominoElement, imageNode: DominoElement) {
     const image = imageNode.getElementsByTagName('img')[0] || imageNode.getElementsByTagName('video')[0]
 
-    if (!this.shouldKeepNode(dump, imageNode, image)) {
+    if (!this.shouldKeepNode(imageNode, image)) {
       DOMUtils.deleteNode(imageNode)
       return
     }
@@ -376,8 +376,13 @@ export abstract class Renderer {
   private async treatImage(dump: Dump, srcCache: KVS<boolean>, articleId: string, img: DominoElement): Promise<{ imageDependencies: string[] }> {
     const imageDependencies: string[] = []
 
-    if (!this.shouldKeepImage(dump, img)) {
+    if (!this.shouldKeepImage(img)) {
       DOMUtils.deleteNode(img)
+      return { imageDependencies }
+    }
+
+    if (dump.nopic && !this.isMathFallbackImage(img)) {
+      this.convertImageToPlaceholder(img)
       return { imageDependencies }
     }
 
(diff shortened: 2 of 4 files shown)


KunalSiyag:
The fix worked for the validation ZIM, but further testing revealed some failures. I will investigate this further.

So far, on two different ZIM builds I encountered the following errors:

1. During indexing:

```
set index
terminate called after throwing an instance of 'Xapian::InvalidOperationError'
```

2. While rendering an article (`Meet_the_Medic_(taunt)`):

```
Error downloading/rendering article Meet_the_Medic_(taunt)
Failed to run mwoffliner after [7s]:
TypeError: Cannot read properties of null (reading '1')
    at extractJsConfigVars (/tmp/mwoffliner/src/util/articles.ts:43:81)
```

Markus-Rost:
> 2. While rendering an article (`Meet_the_Medic_(taunt)`):
> 
> ```
> Error downloading/rendering article Meet_the_Medic_(taunt)
> Failed to run mwoffliner after [7s]:
> TypeError: Cannot read properties of null (reading '1')
>     at extractJsConfigVars (/tmp/mwoffliner/src/util/articles.ts:43:81)
> ```

That's my fault, the JavaScript PR introduced that crash when used on older MW versions. I've added a fix in #2644

benoit74:
> set index
terminate called after throwing an instance of 'Xapian::InvalidOperationError'

This is a known Xapian issue which happens sporadically, probably linked to some race condition. Unless it happens on all runs, you can ignore this error.

Changelog
```diff
@@ -1,4 +1,5 @@
 Unreleased:
+* FIX: Text overlaps on other other text or even images (@kunalsiyag #1004)
```
benoit74: Please change this to a more meaningful description of the final change, something along the lines of "in nopic ZIM, replace images by SVG placeholders to preserve original layout"

KunalSiyag: Will do ASAP

src/renderers/abstract.renderer.ts
```diff
@@ -412,14 +417,105 @@ export abstract class Renderer {
...
+  }
+
+  private getMeaningfulAltText(img: DominoElement) {
+    const alt = img.getAttribute('alt')
+    if (!alt) {
+      return null
+    }
+    const compactAlt = alt.replace(/\u00a0/g, ' ').trim()
+    return compactAlt.length > 0 ? compactAlt : null
+  }
```
Markus-Rost, line 505: MediaWiki likes/liked to use the file name as the alt text for all images. I'm unsure if we should consider those default alt texts as meaningful in this context, as it would mean that basically every image will have one. @benoit74 what do you think?

```html
<img alt="Heavyrespawns.png" src="/w/images/thumb/4/44/Heavyrespawns.png/300px-Heavyrespawns.png" width="300" height="526">
```

KunalSiyag: Maybe we could parse the alt text to make it more specific in this case. For example, we could use stricter case and use regex to only keep the part before the file extension (e.g., Heavyrespawns instead of Heavyrespawns.png). WDYT?

benoit74: Thanks for raising this point, I appreciate!

From my PoV, this is still the "official" alt-text proposed by Mediawiki. I agree it is not very appealing, but I tend to consider that anytime we try to be wiser than what Mediawiki has to offer by fixing its flaws, we also take the risk of creating a half-broken feature which will bite us at some occasion.

I would suggest to consider any alt-text provided by Mediawiki as an alt-text, impact on ZIM size is marginal, UI/UX will probably not be "that awful", and should this become a real issue we can fix that in the aftermath. I'm quite confident we will not create broken (like way too hard to read) ZIMs with these alt-text.

I suggest we merge "as-is" and keep an eye on that topic in ZIMs we will test with 2.0.

Is that ok for you?

Markus-Rost: How does the PR currently solve the alt text for images smaller than the alt text? My main concern with this is small inline icons (like 16x16px) inside a block of text.

Also it seems like newer MW versions stopped adding the file name as alt text, I dont know the reasoning behind this change. However I agree that we should just use the alt text as-is, so long as it won't cause small images to exceed their original size.

benoit74: Indeed, I assumed the SVG would crop the text, but this needs to be checked.

KunalSiyag: Yes, It doesn't crop the text as of now. I will work on to extend the codebase to include case that works on to crop the alt as per the image size.

KunalSiyag:
I worked on the crop part and this is what I get as output:

Before

image

</br>

After

image
</br>

The layout is in place but the cropping of alt text seems to make it not understable though. It  does seem to be a positive step for now. Should we go ahead with this?

benoit74:
To me this is already very good compared to what it was so far.

I feel like we should merge and test this is the wild. The way it is implemented it is seems quite safe to "deploy".

Once we will have gathered experience on the quirks of this implementation, it will be easier to decide if it is worth to invest in an even more reliable solution, and what are the edge cases we wanna handle.

I consider as highly possible the fact that we finally decide to not display any alt text because it create too much "noise".

PR #2644, opened 2026-03-09
Fix extractJsConfigVars for older MW versions

Markus-Rost:
Fix crash on extractJsConfigVars with older MW versions.

Changed javaScript.html to use the old version with `RLQ` for backwards compatibility
Added tests for extractJsConfigVars using both old and current headHtml examples


Markus-Rost:
@benoit74 please review and merge ASAP

PR #2646, opened 2026-03-09
Remove deprecated CLI parameters

### LINKED ISSUE #2489: Remove mwWikiPath and mwIndexPhpPath options

Markus-Rost:
Remove deprecated CLI parameters `mwWikiPath`, `mwIndexPhpPath`, `keepEmptyParagraphs`

Fix #2489

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -140,23 +140,17 @@
       "title": "Add categories",
       "description": "[WIP] Download category pages"
     },
-    "keepEmptyParagraphs": {
+    "keepEmptySections": {
       "type": "boolean",
       "required": false,
-      "title": "Keep empty paragraphs",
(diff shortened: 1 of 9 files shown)


offliner-definition.json
```diff
@@ -140,24 +140,12 @@
       "title": "Add categories",
       "description": "[WIP] Download category pages"
     },
-    "keepEmptyParagraphs": {
```
benoit74, line 143: Rather than removing it from this file, we should replace it with `keepEmptySections` (we did not introduced the new setting yet to avoid confusion)

Markus-Rost: Done

PR #2648, opened 2026-03-09
Make processStylesheetContent use getMediaBase

### LINKED ISSUE #2647: `processStylesheetContent` has bad url rewrite logic
`processStylesheetContent` has very bad url rewrite logic. It just takes the base name of a url:
https://github.com/openzim/mwoffliner/blob/7372b29be321c2bc06621a1b17e8de6c2893381e/src/util/dump.ts#L40-L42
And stores it at `_mw_/FILENAME`
https://github.com/openzim/mwoffliner/blob/7372b29be321c2bc06621a1b17e8de6c2893381e/src/util/dump.ts#L69

This can result in:
1. the same file being included twice in the ZIM, once at `_assets_` used by an `` element and once in `_mw_` used in a css file.
2. name conflicts between files using css with the same base name but different urls

`processStylesheetContent` should use the same logic as `` elements and store any files in `_assets_` to avoid these issues.

Markus-Rost:
Fix #2647

### DIFF
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -1,9 +1,8 @@
 import { fileURLToPath } from 'url'
-import * as pathParser from 'path'
 import * as logger from '../Logger.js'
 import Downloader from '../Downloader.js'
 import RedisStore from '../RedisStore.js'
-import { getFullUrl, jsPath, cssPath, getRelativeFilePath } from './index.js'
+import { getFullUrl, jsPath, cssPath, getRelativeFilePath, getMediaBase } from './index.js'
 import { config } from '../config.js'
 import MediaWiki from '../MediaWiki.js'
 import { Creator, StringItem } from '@openzim/libzim'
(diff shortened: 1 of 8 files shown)


test/e2e/bm.e2e.test.ts
```diff
@@ -29,7 +29,7 @@ await testAllRenders('bm-wikipedia', parameters, async (outFiles) => {
         // nopic has enough files (this is just an estimate and can change
         // with time, as new Mediwiki versions are released).
         expect(dump.status.files.success).toBeGreaterThanOrEqual(3)
```
benoit74: We should update the minimum value to be closer to what we expect.

src/util/dump.ts
```diff
@@ -33,43 +32,22 @@ export function processStylesheetContent(cssUrl: string, linkMedia: string, body
...
-          Downloader.cssDependenceUrls[url] = true
-          filesToDownloadXPath.set(config.output.dirs.mediawiki + '/' + decodedFilename, { url: urlHelper.serializeUrl(url), kind: 'media' })
-        }
-      } else {
-        logger.warn(`Skipping CSS [url(${url})] because the pathname could not be found [${filePathname}]`)
+      const fullurl = getFullUrl(url, cssUrl)
+      const filepath = getMediaBase(fullurl, true)
+
+      /* Rewrite the CSS */
+      const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
```
benoit74, line 43: I'm sorry but I don't get the logic here. 

Why when we are treating a JS file we need `__RELATIVE_FILE_PATH__` but when not (treating CSS, right?) we are sure to need only `../`.  Can't we have a CSS stored in a "subfolder" as well?

benoit74 (changes requested):
Small questions / remarks

src/util/dump.ts
Markus-Rost: If we have inline css in an article, we need to update the path relative to the article using getRelativeFilePath

If we have css inside a JS module, the ResourceLoader will insert the styles inline into the article. We need to get the relative file path in that case as well, but don't know it ahead of time. The ResourceLoader JS is patched to replace `__RELATIVE_FILE_PATH__` when inserting the styles dynamically.

Otherwise we have a normal CSS file, those are (as far as I know) all directly in the `_mw_` directory. That's also the assumption this function previously made. With the custom CSS PR coming now, we can also have CSS files in `_res_`, but those also don't use subdirectories so this change will work for both cases.

benoit74: Thank you!

test/e2e/bm.e2e.test.ts
Markus-Rost: I have increased all files.success values by ~80

benoit74:
(unit test failure is irrelevant, a transient failure most probably)

PR #2656, opened 2026-03-12
Recursively resolve CSS @import statements

### LINKED ISSUE #2649: Recursively import CSS stylesheets
See https://github.com/openzim/zim-requests/issues/1472

Looks like it miss at least:
- https://mtg.wiki/customizations/customizations.css
- https://mtg.wiki/index.php?title=MediaWiki:Universal.css&action=raw&ctype=text/css
- https://mtg.wiki/index.php?title=MediaWiki:Mbox.css&action=raw&ctype=text/css
- https://mtg.wiki/index.php?title=MediaWiki:Navbox.css&action=raw&ctype=text/css
- https://mtg.wiki/index.php?title=MediaWiki:Infobox.css&action=raw&ctype=text/css
- https://mtg.wiki/index.php?title=MediaWiki:Keyrune.css&action=raw&ctype=text/css

While I'm quite sure the first one is a typical case which needs to be solved with https://github.com/openzim/mwoffliner/issues/2448 (soon to be fixed), I'm a bit more confused about all the other ones because I don't achieve to find where they come from.

Are we missing another kind of modules we should load?

triemerge:
Fix #2649

`processStylesheetContent` now uses an iterative worklist to resolve `@import` statements: each CSS body is scanned for `@import` URLs, which are downloaded and added to the queue for further processing. This loops until no new imports remain.

- Handles both `@import url(...)` and `@import "..."` syntaxes
- `url()` references in imported stylesheets are resolved relative to each stylesheet's own URL
- Circular imports are handled via a `Set` (no depth limit needed)
- Function is now `async`; call sites in `downloadModule` and `applyOtherTreatments` updated accordingly

### DIFF
--- a/src/util/dump.ts
+++ b/src/util/dump.ts
@@ -15,44 +15,147 @@ import { zimCreatorMutex } from '../mutex.js'
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
 
-export function processStylesheetContent(cssUrl: string, linkMedia: string, body: string, articleId?: string, isJs?: boolean) {
+export async function processStylesheetContent(cssUrl: string, linkMedia: string, body: string, articleId?: string, isJs?: boolean) {
   // articleId is supposed to be passed only when we rewrite inline CSS for a given article and we hence have
   // to compute relative path to assets
 
   const { filesToDownloadXPath } = RedisStore
+  const importRegexp = /@import\s+(?:url\(\s*(['"]?)(.*?)\1\s*\)|(['"])(.*?)\3)\s*([^;]*);/gi
   const cssUrlRegexp = new RegExp('url\\([\'"]{0,1}(.*?)[\'"]{0,1}\\)', 'gi')
 
-  let rewrittenCss = `\n/* start ${cssUrl} */\n\n`
-  rewrittenCss += linkMedia ? `@media ${linkMedia}  {\n` : '\n'
-  rewrittenCss += `${body}\n`
-  rewrittenCss += linkMedia ? `} /* @media ${linkMedia} */\n` : '\n'
-  rewrittenCss += `\n/* end   ${cssUrl} */\n`
+  type CssPart = { type: 'css'; text: string }
+  type ImportPart = { type: 'import'; url: string; conditions: string }
+  type StylesheetParts = Array<CssPart | ImportPart>
 
-  /* Downloading CSS dependencies */
-  let match: any
-  // tslint:disable-next-line:no-conditional-assignment
-  while ((match = cssUrlRegexp.exec(body))) {
-    const url = match[1]
+  const wrapStylesheetContent = (css: string, url: string, conditions: string) => {
+    let wrappedCss = css
+    const rawConditions = conditions.trim()
 
-    /* Avoid 'data', so no URL dependency */
-    if (url && !url.match('^data')) {
-      const fullurl = getFullUrl(url, cssUrl)
-      const filepath = getMediaBase(fullurl, true)
+    if (rawConditions) {
+      const supportsMatch = rawConditions.match(/supports\s*\((.*?)\)/i)
+      const supportsCondition = supportsMatch?.[1]?.trim()
+      const mediaCondition = rawConditions.replace(/supports\s*\((.*?)\)/i, '').trim()
 
-      /* Rewrite the CSS */
-      const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
-      rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
-
-      /* Download CSS dependency, but avoid duplicate calls */
-      // eslint-disable-next-line no-prototype-builtins
-      if (!Downloader.cssDependenceUrls.hasOwnProperty(fullurl) && filepath) {
-        Downloader.cssDependenceUrls[fullurl] = true
-        filesToDownloadXPath.set(filepath, { url: urlHelper.serializeUrl(fullurl), kind: 'media' })
+      if (supportsCondition) {
+        wrappedCss = `@supports (${supportsCondition}) {\n${wrappedCss}\n}\n`
+      }
+      if (mediaCondition) {
+        wrappedCss = `@media ${mediaCondition} {\n${wrappedCss}\n}\n`
       }
     }
+
+    return `\n/* start ${url} */\n\n${wrappedCss}\n/* end   ${url} */\n`
   }
 
-  return rewrittenCss
+  const rewriteCssDependencies = (sourceCss: string, sourceUrl: string) => {
+    let rewrittenCss = sourceCss
+    let match: any
+
+    while ((match = cssUrlRegexp.exec(sourceCss))) {
+      const url = match[1]
+
+      /* Avoid 'data', so no URL dependency */
+      if (url && !url.match('^data')) {
+        const fullurl = getFullUrl(url, sourceUrl)
+        const filepath = getMediaBase(fullurl, true)
+
+        /* Rewrite the CSS */
+        const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
+        rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
+
+        /* Download CSS dependency, but avoid duplicate calls */
+        if (!Object.prototype.hasOwnProperty.call(Downloader.cssDependenceUrls, fullurl) && filepath) {
+          Downloader.cssDependenceUrls[fullurl] = true
+          filesToDownloadXPath.set(filepath, { url: urlHelper.serializeUrl(fullurl), kind: 'media' })
+        }
+      }
+    }
+
+    cssUrlRegexp.lastIndex = 0
+    return rewrittenCss
+  }
+
+  // Iterative worklist to resolve @import statements (no depth limit).
+  // Each entry carries its own URL so that url() references are resolved relative to each stylesheet.
+  const processedImportUrls = new Set<string>()
+  const worklist: Array<{ url: string; body: string }> = [{ url: cssUrl, body }]
+  const stylesheetParts = new Map<string, StylesheetParts>()
+
+  while (worklist.length > 0) {
+    const current = worklist.shift()!
+    const currentBody = current.body
+
+    const parts: StylesheetParts = []
+    let startIdx = 0
+
+    /* Find and queue @import URLs */
+    let importMatch: RegExpExecArray | null
+    const importRe = new RegExp(importRegexp.source, importRegexp.flags)
+    while ((importMatch = importRe.exec(currentBody)) !== null) {
+      if (importMatch.index > startIdx) {
+        parts.push({ type: 'css', text: currentBody.substring(startIdx, importMatch.index) })
+      }
+
+      const importUrl = importMatch[2] || importMatch[4]
+      const importConditions = (importMatch[5] || '').trim()
+
+      if (importUrl) {
+        const fullUrl = getFullUrl(importUrl, current.url)
+        parts.push({ type: 'import', url: fullUrl, conditions: importConditions })
+
+        if (!processedImportUrls.has(fullUrl)) {
+          processedImportUrls.add(fullUrl)
+          try {
+            const { content } = await Downloader.downloadContent(fullUrl, 'css')
+            worklist.push({ url: fullUrl, body: content.toString() })
+          } catch {
+            logger.warn(`Failed to download imported CSS: ${fullUrl}`)
+          }
+        }
+      }
+
+      startIdx = importRe.lastIndex
+    }
+
+    if (startIdx < currentBody.length) {
+      parts.push({ type: 'css', text: currentBody.substring(startIdx) })
+    }
+
+    stylesheetParts.set(current.url, parts)
+  }
+
+  const renderedImportedUrls = new Set<string>()
+  const renderStack = new Set<string>()
+
+  const renderStylesheet = (url: string, conditions = ''): string => {
+    if (renderedImportedUrls.has(url)) {
+      return ''
+    }
+    if (renderStack.has(url)) {
+      return ''
+    }
+
+    const parts = stylesheetParts.get(url)
+    if (!parts) {
+      return ''
+    }
+
+    renderedImportedUrls.add(url)
+    renderStack.add(url)
+    let currentBody = ''
+    for (const part of parts) {
+      if (part.type === 'css') {
+        currentBody += rewriteCssDependencies(part.text, url)
+      } else {
+        currentBody += renderStylesheet(part.url, part.conditions)
+      }
+    }
+    renderStack.delete(url)
+
+    return wrapStylesheetContent(currentBody, url, conditions)
+  }
+
+  return renderStylesheet(cssUrl, linkMedia)
 }
 
 export async function downloadModule(module: string, type: 'js' | 'css') {
@@ -141,9 +244,11 @@ export async function downloadModule(module: string, type: 'js' | 'css') {
       try {
         const cssParts: string[] = JSON.parse(embeddedCss[1])
         const processedCss = JSON.stringify(
-          cssParts.map((cssPart) => {
-            return processStylesheetContent(moduleApiUrl, '', cssPart, '', true)
-          }),
+          await Promise.all(
+            cssParts.map((cssPart) => {
+              return processStylesheetContent(moduleApiUrl, '', cssPart, '', true)
+            }),
+          ),
         ).replace(/__RELATIVE_FILE_PATH__/g, '"+RLCONF.zimRelativeFilePath+"')
         text = text.replace(embeddedCss[0], `,{"css":${processedCss}}`)
       } catch (e) {
@@ -153,7 +258,7 @@ export async function downloadModule(module: string, type: 'js' | 'css') {
   }
 
   if (type === 'css') {
-    text = processStylesheetContent(moduleApiUrl, '', text, '')
+    text = await processStylesheetContent(moduleApiUrl, '', text, '')
   }
 
   // Zimcheck complains about empty files, and it is too late to decide to not create this file
@@ -212,7 +317,7 @@ export function getModuleDependencies(oneModule: ResourceLoaderModule, allModule
 export async function downloadAndSaveCustomCss(zimCreator: Creator, cssUrl: string, filename: string): Promise<void> {
   logger.log(`Downloading custom CSS [${cssUrl}]`)
   const { content: cssBody } = await Downloader.downloadContent(cssUrl, 'css')
-  const processedCss = processStylesheetContent(cssUrl, '', cssBody.toString())
+  const processedCss = await processStylesheetContent(cssUrl, '', cssBody.toString())
   const zimPath = cssPath(filename, config.output.dirs.res)
   await zimCreatorMutex.runExclusive(() => zimCreator.addItem(new StringItem(zimPath, 'text/css', null, { FRONT_ARTICLE: 0 }, processedCss)))
   logger.log(`Saved custom CSS [${cssUrl}] at ${zimPath}`)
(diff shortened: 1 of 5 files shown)


src/util/dump.ts
```diff
@@ -15,44 +15,82 @@ import { zimCreatorMutex } from '../mutex.js'
...
-      /* Rewrite the CSS */
-      const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
-      rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
-
-      /* Download CSS dependency, but avoid duplicate calls */
-      // eslint-disable-next-line no-prototype-builtins
-      if (!Downloader.cssDependenceUrls.hasOwnProperty(fullurl) && filepath) {
-        Downloader.cssDependenceUrls[fullurl] = true
-        filesToDownloadXPath.set(filepath, { url: urlHelper.serializeUrl(fullurl), kind: 'media' })
+  const importRegexp = /@import\s+(?:url\(\s*['"]?\s*(.*?)\s*['"]?\s*\)|['"]([^'"]+)['"])[^;]*;/gi
```
Markus-Rost: `supports()` and any media queries in the import statement are ignored. They should instead be converted into `@supports` and `@media` rules.

src/util/dump.ts
```diff
@@ -15,44 +15,82 @@ import { zimCreatorMutex } from '../mutex.js'
...
+        /* Download CSS dependency, but avoid duplicate calls */
+        // eslint-disable-next-line no-prototype-builtins
+        if (!Downloader.cssDependenceUrls.hasOwnProperty(fullurl) && filepath) {
+          Downloader.cssDependenceUrls[fullurl] = true
+          filesToDownloadXPath.set(filepath, { url: urlHelper.serializeUrl(fullurl), kind: 'media' })
+        }
       }
     }
+
+    allCss += rewrittenCss
```
Markus-Rost: The CSS of the import statements is added below the CSS of the main file, instead of at the exact position of the import statement. This is changing the cascade of the styles, mostly likely resulting in broken styles when viewing the articles. Specially considering that import statements are only allowed at the very top of a file and therefore the main file in expected to easily overwrite previous styles from the import.

test/unit/util/dump.test.ts
```diff
@@ -87,4 +88,104 @@ describe('Download CSS or JS Module', () => {
...
+        return { content: Buffer.from('.from-b { color: green; }'), contentType: 'text/css', setCookie: null }
+      }
+      throw new Error(`Unexpected URL: ${url}`)
+    })
+
+    const rewrittenCSS = await processStylesheetContent('https://example.wiki/load.php?modules=site.styles', '', '@import url("/css/a.css");\n.main { color: blue; }', '')
+
+    expect(rewrittenCSS).toContain('.main { color: blue; }')
+    expect(rewrittenCSS).toContain('.from-a { color: red; }')
+    expect(rewrittenCSS).toContain('.from-b { color: green; }')
```
Markus-Rost, line 130: Test for the correct order of the styles: `.from-b` before `.from-a` before `.main`

src/util/dump.ts
triemerge: Thanks, fixed. <code>@import</code> conditions are now preserved by converting them to wrappers around imported CSS: <code>supports(...)</code> -&gt; <code>@supports (...) { ... }</code>, and remaining media condition -&gt; <code>@media ... { ... }</code>. Added dedicated test: <code>convert @import supports and media queries to wrappers</code>.

src/util/dump.ts
triemerge: Imported CSS is now injected at the exact import position, so cascade order is preserved.
Added explicit order assertions in tests for the nested case.

test/unit/util/dump.test.ts
triemerge: Done

benoit74:
I'm sorry but I don't get why we are replacing the import with imported code. To me this open the doors to too many subtle problems (like the supports, media, the place of the imported CSS, and potentially many others we did not yet identified).

Why not "simply" rewriting the URL to a proper "in-ZIM" URL, add the imported CSS url to the list of CSS to place into the ZIM, let this logic do its own work (hence resolving again potentially nested CSS url). This look way less fragile to me than current implementation, but I might miss something.

triemerge:
Switched away from inlining: `@import` now points to in-ZIM URLs, imported CSS is queued as CSS for the normal pipeline, and newly discovered imports are drained continuously so nested chains are resolved in the same run; tests were updated accordingly.

src/util/dump.ts
```diff
@@ -15,44 +15,87 @@ import { zimCreatorMutex } from '../mutex.js'
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
 
-export function processStylesheetContent(cssUrl: string, linkMedia: string, body: string, articleId?: string, isJs?: boolean) {
+export async function processStylesheetContent(cssUrl: string, linkMedia: string, body: string, articleId?: string, isJs?: boolean) {
```
Markus-Rost, line 18: Nothing async about this function anymore.

src/util/dump.ts
```diff
@@ -15,44 +15,87 @@ import { zimCreatorMutex } from '../mutex.js'
...
-  return rewrittenCss
+  const queueImportedStylesheets = (sourceCss: string, sourceUrl: string) => {
+    return sourceCss.replace(importRegexp, (fullMatch, urlQuote1, importUrlFromUrlFunc, urlQuote2, importUrlFromQuoted, rawConditions) => {
+      const importUrl = importUrlFromUrlFunc || importUrlFromQuoted
+      if (!importUrl || importUrl.match('^data')) {
+        return fullMatch
+      }
+
+      const dependency = getRelativePathForDependency(getFullUrl(importUrl, sourceUrl))
+      filesToDownloadXPath.set(dependency.filepath, { url: urlHelper.serializeUrl(dependency.fullUrl), kind: 'css' })
```
Markus-Rost: ~Imported CSS is not getting rewritten this way.~

src/util/dump.ts
```diff
@@ -15,44 +15,87 @@ import { zimCreatorMutex } from '../mutex.js'
...
-      const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
-      rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
-
-      /* Download CSS dependency, but avoid duplicate calls */
-      // eslint-disable-next-line no-prototype-builtins
-      if (!Downloader.cssDependenceUrls.hasOwnProperty(fullurl) && filepath) {
-        Downloader.cssDependenceUrls[fullurl] = true
-        filesToDownloadXPath.set(filepath, { url: urlHelper.serializeUrl(fullurl), kind: 'media' })
+  const getRelativePathForDependency = (fullUrl: string) => {
+    const filepath = getMediaBase(fullUrl, true)
```
Markus-Rost: With `getMediaBase` being made for image paths, the CSS urls will almost always use the hashed fallback. It also places the CSS files in the `_assets_` directory, where we so far don't have any CSS files. Unsure about this being issues that need fixing, but something I wanted to point out.

src/util/saveArticles.ts
```diff
@@ -156,9 +184,13 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
     await Downloader.downloadContent(fileToDownload.url, fileToDownload.kind, false)
       .then(async (resp) => {
         if (resp && resp.content && resp.contentType) {
+          let content = resp.content
+          if (fileToDownload.kind === 'css') {
+            content = await processStylesheetContent(urlHelper.deserializeUrl(fileToDownload.url), '', resp.content.toString(), '')
```
Markus-Rost: File name should be provided as `articleId` to ensure `processStylesheetContent` is correctly rewriting urls.

src/util/dump.ts
triemerge: Removed async from processStylesheetContent and cleaned up remaining await usage at call sites.

src/util/dump.ts
triemerge: Good point. Kept `getMediaBase` fallback behavior for consistency with the current asset-path strategy, and added edge-case tests for:
- MediaWiki raw CSS imports from `site.styles`
- protocol-relative `@import` URLs

src/util/saveArticles.ts
triemerge: Fixed.

When processing downloaded CSS files, `fileToDownload.path` is now passed as `articleId`, so nested CSS URL rewriting is computed from the actual CSS file location.

Also added a regression test covering nested imports from `_assets_` paths.

benoit74:
@triemerge please do not move too fast, I'm asking question to understand what we are doing and why, unless I request an explicit change you risk to have to rollback changes quite often...

> To me it seems like "simply" rewriting the URL to a proper "in-ZIM" URL is actually the more difficult solution. We don't currently have a proper queue for CSS files, so that needs to be added as well to ensure the imported CSS gets properly rewritten. Additionally we aren't able to get nice names for the imported CSS files, so they will all end up with hashed file names.

This argument makes a lot of sense to me. We do not have that many nested imports, and I now get correctly why the added complexity is in fact equivalent in both alternatives. Looking at latest changes (since you wrote them ...), I'm way less confident about the fact this is the correct way to handle the situation.

@Markus-Rost did you changed your mind now that you see the implementation, or are you ok with the fact that we move back to original implementation where imported CSS was resolved "in-place"?

Markus-Rost:
> @Markus-Rost did you changed your mind now that you see the implementation, or are you ok with the fact that we move back to original implementation where imported CSS was resolved "in-place"?

Both implementations work for me, though I prefer the original "in-place" version as it keeps CSS rewriting in one place instead of split across multiple files.

triemerge:
> @triemerge please do not move too fast, I'm asking question to understand what we are doing and why, unless I request an explicit change you risk to have to rollback changes quite often...
> 
> > To me it seems like "simply" rewriting the URL to a proper "in-ZIM" URL is actually the more difficult solution. We don't currently have a proper queue for CSS files, so that needs to be added as well to ensure the imported CSS gets properly rewritten. Additionally we aren't able to get nice names for the imported CSS files, so they will all end up with hashed file names.
> 
> This argument makes a lot of sense to me. We do not have that many nested imports, and I now get correctly why the added complexity is in fact equivalent in both alternatives. Looking at latest changes (since you wrote them ...), I'm way less confident about the fact this is the correct way to handle the situation.

Thanks for the clarification. I'll hold off on any further changes until we've talked through the approach together.

triemerge:
@benoit74 @Markus-Rost Switched back to the in-place approach for CSS `@import` handling.

src/util/saveArticles.ts
```diff
@@ -34,38 +35,57 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
   }
 
   let prevPercentProgress: string
+  let filesTotal = 0
+  const queuedFilePaths = new Set<string>()
```
benoit74: I don't get the added value of the addition of `queuedFilePaths` set, can you explain this a bit? Is that for deduplication purposes? I don't get how we could have duplicates here.

src/util/dump.ts
```diff
@@ -15,44 +15,166 @@ import { zimCreatorMutex } from '../mutex.js'
...
+
+    return wrappedCss
+  }
+
+  const rewriteCssDependencies = (sourceCss: string, sourceUrl: string) => {
+    let rewrittenCss = sourceCss
+    let match: any
+
+    // tslint:disable-next-line:no-conditional-assignment
+    while ((match = cssUrlRegexp.exec(sourceCss))) {
```
benoit74, line 54: You should be able to fix this line to remove tslint rule (see `except-parens` option doc at https://eslint.org/docs/latest/rules/no-cond-assign, I don't think we've set the option to `always`)

src/util/dump.ts
```diff
@@ -15,44 +15,166 @@ import { zimCreatorMutex } from '../mutex.js'
...
+        const fullurl = getFullUrl(url, sourceUrl)
+        const filepath = getMediaBase(fullurl, true)
+
+        /* Rewrite the CSS */
+        const relativePath = articleId ? getRelativeFilePath(articleId, filepath) : isJs ? `__RELATIVE_FILE_PATH__${filepath}` : `../${filepath}`
+        rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
+
+        /* Download CSS dependency, but avoid duplicate calls */
+        // eslint-disable-next-line no-prototype-builtins
+        if (!Downloader.cssDependenceUrls.hasOwnProperty(fullurl) && filepath) {
```
benoit74: I know this was here before you came, but we can fix this to remove the eslint ignore comment, see https://eslint.org/docs/latest/rules/no-prototype-builtins documentation

test/unit/util/dump.test.ts
```diff
@@ -87,4 +88,129 @@ describe('Download CSS or JS Module', () => {
...
+    })
+
+    const rewrittenCSS = await processStylesheetContent(
+      'https://example.wiki/load.php?modules=site.styles',
+      '',
+      '@import url("/customizations/custom.css");\n.main { color: blue; }',
+      '',
+    )
+
+    expect(rewrittenCSS).toContain('.main { color: blue; }')
```
benoit74, line 107: Why not testing for exact `rewrittenCSS` content? To me current structure makes reading the test way more complex, and I'm not even sure we have properly formatted CSS in the end. But I might miss the point of structuring tests like this, so please explain (especially since same pattern is used in all tests)

src/util/dump.ts
triemerge: Thanks, fixed. I removed the old `tslint` suppression here and kept the existing `while ((match = ...))` form, which is already compliant with the current lint rule configuration.

src/util/dump.ts
triemerge: Good catch, fixed as suggested. I removed the eslint ignore and replaced the direct `hasOwnProperty` call with `Object.prototype.hasOwnProperty.call(...)`.

src/util/saveArticles.ts
triemerge: Thanks for raising this. Yes, `queuedFilePaths` is for deduplication in the dynamic refill flow.

With the current architecture, `queuePendingFilesToDownload` runs repeatedly while downloads are in progress, and new paths can be added to the store during CSS processing (via `rewriteCssDependencies`). Without the set, a path read from the store but not yet deleted can be picked up again on a later call and pushed to a host queue twice. The set also keeps `filesTotal` stable, since the counter is incremented in the same function.

test/unit/util/dump.test.ts
triemerge: I used semantic assertions intentionally, because `rewrittenCSS` includes generated wrapper structure and formatting details (comment markers, injected newlines, and optional media wrappers), so exact string matching can become brittle.

The current checks focus on behavior: imported rules are present, `@import` is removed, and ordering is correct where relevant (explicitly checked in the nested-import test). Happy to switch to snapshots if you prefer stricter output validation.

src/util/saveArticles.ts
benoit74: OK, I got it.

The problem is that (if I get it correctly) it means that `queuedFilePaths` has the list of all files to download in memory, while we want to avoid keeping any big dataset in Node.JS memory, to move that responsibility to the Redis.

Maybe simply not flushing / deleting the `fileStore` would be sufficient?

test/unit/util/dump.test.ts
benoit74: OK, got it, thank you

src/util/saveArticles.ts
triemerge: Thanks for the valid point on memory. You’re right that `queuedFilePaths` grows with total files, which goes against the Redis-first approach. At the same time, simply not deleting from `fileStore` alone would not be sufficient: without dedup, each polling pass can re-read and re-queue the same paths.

One Redis-native option would be to extend `FileDetail` with a `queued` flag, mark entries as queued in `fileStore` instead of deleting them, and skip already-queued entries in `iterateItems`. This keeps state in Redis using the existing store, without introducing a new Redis keyspace. Would you like me to go with this approach?

benoit74: I don't think this will work if I get it correctly, because it means we will iterate over the whole list of files everytime we call `queuePendingFilesToDownload`, which is once everytime a file is downloaded ... this is not going to work well in terms of perf.

I think we have no choice but to create the queuedFilePaths in Redis (with only the path, no need for file details).

Btw, we probably do not need the `queuedPathsBatch` variable, we can flush the fileStore at every run, all files in the fileStore are processed each time.

WDYT?

triemerge: Agreed, this makes sense. 
I’ll replace the in-memory dedup with a Redis path-only dedup set and use `SADD` return value as the enqueue gate.

On the `flush()` suggestion: I think we should keep deleting only the processed batch, because a full flush right after scanning could drop paths added concurrently during downloads before they are scanned/marked. 

If you still prefer `flush()`, I can switch to it. I mainly wanted to avoid potentially losing newly-added paths.

benoit74: Indeed regarding parallel processing, got it, you are right. Then maybe we can add an argument to do a flush in first call (where we do not have anything running in parallel yet) and use the array in subsequent ones. Again, I would like to avoid store a too big list in memory.

triemerge: Makes sense. I'll add an `isFirstCall` parameter: `flush()` on the first call (before workers start), and `deleteMany` on the processed batch in subsequent calls. This keeps cleanup safe and keeps per-pass memory bounded. Will implement it.

triemerge:
> * all redis stores are created and maintained in the `RedisStore`, including "cleanup" at scraper startup.
> * you probably need to create a new `RedisSet` object, next to `RedisQueue` and `RedisKvs`, so that it is easier to reuse without having to mind too much about Redis paradigms

Replaced raw `client.sAdd` usage and the in-memory set with a `RedisSet` abstraction, alongside `RedisQueue` and `RedisKvs` under `RedisStore`, with the same lifecycle handling (`checkForExistingStores`, `close()`). Also added `deleteMany` for symmetry with `RedisKvs` and aligned the `RSET` typings. Unit tests included.

triemerge:
> With the switch back to the in-place approach, the changes in `saveArticles.ts` and everything related to RedisSet are no longer needed.
> 
> They should be removed or potentially moved to a dedicated PR as those changes are still being discussed and as such currently blocking the actual unrelated `@import` fix.

Makes sense. The `RedisSet` logic came out of the recent review feedback, but with the switch back to in-place resolution, those changes are no longer needed here. I'll remove them and keep this PR focused on just the `@import` fix.

triemerge:
Removed the `RedisSet` and `saveArticles.ts` changes. The PR is now focused on just the `@import` resolution fix in `dump.ts`.

Also fixed two issues noticed while revisiting the implementation:
- Nested stylesheets had their `url()` resources rewritten multiple times against the parent's base URL.
- Missing parentheses on `queue.flush()` in `test/unit/bootstrap.ts` caused test queues not to clear properly.

Will soon open a separate PR for the `RedisSet` utility.

src/util/dump.ts
```diff
@@ -15,44 +15,162 @@ import { zimCreatorMutex } from '../mutex.js'
...
+    /* Build CSS with comment markers */
+    let rewrittenCss = `\n/* start ${url} */\n\n`
+    rewrittenCss += isMain && linkMedia ? `@media ${linkMedia}  {\n` : '\n'
+    rewrittenCss += `${currentBody}\n`
+    rewrittenCss += isMain && linkMedia ? `} /* @media ${linkMedia} */\n` : '\n'
+    rewrittenCss += `\n/* end   ${url} */\n`
+
+    if (!isMain && importConditions) {
+      rewrittenCss = applyImportConditions(rewrittenCss, importConditions)
+    }
```
Markus-Rost: Move the comment markers inside the `applyImportConditions` function and provide `linkMedia` as parameter to the main `renderStylesheet` call. No need to have this wrapper logic twice. applyImportConditions might need a more neutral name in that case.

triemerge: Done, merged the comment markers into `wrapStylesheetContent` and passed `linkMedia` directly to `renderStylesheet`. Renamed it to reflect the combined responsibility.

PR #2659, opened 2026-03-16
Fix-2658: do not always retry malformed JSON

### LINKED ISSUE #2658: Do not always retry malformed JSON
https://github.com/openzim/mwoffliner/pull/2621 introduced a regression in error handling.

It is now retrying too many requests, because Mediawiki tends to often respond with an HTML document when an error occurs (429, 404, ...).

We need to find a solution to keep old error processing logic and only retry what was reported in https://github.com/openzim/mwoffliner/issues/2609 ; main problem seems to be that Axios always raises an error when JSON parsing fails.

@KunalSiyag do you wanna give this a look?

KunalSiyag:
This fixes #2658 by restoring raw-response handling for `getJSON()` requests and only retrying actual malformed JSON parse failures.

`#2621` made Axios parse API responses eagerly with strict JSON parsing. That caused HTML error pages returned by MediaWiki (for example `404`, `429`, `503`) to be misclassified as malformed JSON retry candidates before `mwoffliner` could inspect the raw status/body/content type.

This change switches JSON requests back to raw text handling, parses successful API responses manually, and keeps malformed JSON retries limited to true parse failures.

## Changes

- change `Downloader` JSON requests to use `responseType: "text"`
- remove Axios transitional strict JSON parsing from the JSON request options
- manually parse successful API responses in `getJSONCb()`
- retry only real malformed JSON parse failures (`SyntaxError`, direct or wrapped with `cause`)
- preserve raw non-`2xx` HTML/text responses for existing error handling
- fallback to ActionParse `jsconfigvars` when `headhtml` config cannot be parsed as strict JSON
- add unit coverage for:
  - raw text JSON parsing
  - malformed JSON retry and success-after-retry
  - generic `ERR_BAD_RESPONSE` not being treated as malformed JSON
  - raw HTML `404` responses being preserved
  - `extractJsConfigVars()` fallback for non-JSON config blobs

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -71,6 +71,10 @@ interface CompressionData {
   data: any
 }
 
+function isJsonContentType(contentType: string | null): boolean {
+  return typeof contentType === 'string' && /\b(?:application\/json|[\w.-]+\+json)\b/i.test(contentType)
+}
+
 function isMalformedJsonResponseError(err: unknown): boolean {
   if (err instanceof SyntaxError) {
     return true
@@ -81,6 +85,22 @@ function isMalformedJsonResponseError(err: unknown): boolean {
   }
 
   return err.cause instanceof SyntaxError
+}
+
+function parseJsonResponse<T>(url: string, status: number, contentType: string | null, body: unknown): T {
+  if (body !== null && typeof body === 'object') {
+    return body as T
+  }
+
+  if (typeof body !== 'string') {
+    return body as T
+  }
+
+  if (!isJsonContentType(contentType)) {
+    throw new DownloadError('Unexpected non-JSON response while calling API', url, status, contentType, body)
+  }
+
+  return JSON.parse(body) as T
 }
 
 export class DownloadError extends Error {
@@ -263,12 +283,7 @@ class Downloader {
         accept: 'application/json',
         'accept-encoding': 'gzip, deflate',
       },
-      responseType: 'json',
-      transitional: {
-        // Parse failures must throw so backoff retries can be triggered.
-        silentJSONParsing: false,
-        forcedJSONParsing: true,
-      },
+      responseType: 'text',
       method: 'GET',
     }
 
@@ -675,10 +690,13 @@ class Downloader {
     logger.info(`Getting JSON from [${url}]`)
     this.request<T>({ url, method: 'GET', ...this.jsonRequestOptions })
       .then((val) => {
-        if ((val.data as any).error) {
-          handler(new DownloadError(`Error returned while calling API`, url, val.status, val.headers['content-type'].toString(), val.data))
+        const contentType = val.headers['content-type']?.toString() || null
+        const data = parseJsonResponse<T>(url, val.status, contentType, val.data)
+
+        if ((data as any).error) {
+          handler(new DownloadError(`Error returned while calling API`, url, val.status, contentType, data))
         } else {
-          handler(null, val.data)
+          handler(null, data)
         }
       })
       .catch((err) => {
--- a/test/unit/downloader.retry.test.ts
+++ b/test/unit/downloader.retry.test.ts
@@ -20,13 +20,29 @@ const createDownloader = () => {
   return downloader
 }
 
+const stubImmediateBackoff = (downloader: DownloaderClass) => {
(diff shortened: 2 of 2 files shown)


KunalSiyag:
I validated the changes against the dwarffortresswiki scrape. During this process I discovered a separate issue where a config value was being parsed as JSON, leading to an error such as:

SyntaxError: Unexpected token '!', ..."akFrames":!0,"wgSepa"... is not valid JSON

The fallback added in articles.ts is not intended as the fix for that #2658 itself. Rather, it is an additional renderer-side robustness improvement that was identified during validation.

Markus-Rost:
> I validated the changes against the dwarffortresswiki scrape. During this process I discovered a separate issue where a config value was being parsed as JSON, leading to an error such as:
> 
> SyntaxError: Unexpected token '!', ..."akFrames":!0,"wgSepa"... is not valid JSON
> 
> The fallback added in articles.ts is not intended as the fix for that #2658 itself. Rather, it is an additional renderer-side robustness improvement that was identified during validation.

God damn minifier -.-

Could you remove that fix from this PR? I'll create an extra PR for that as it needs to be accounted for in a few more places.

test/unit/util/articles.test.ts
```diff
@@ -185,4 +185,24 @@ describe('articles utility', () => {
     }
     expect(extractJsConfigVars(fakeHeadHtml)).toEqual(result[mwVersion])
   })
+
+  test('extractJsConfigVars falls back to API jsconfigvars when headHtml config is not strict JSON', () => {
+    const headHtml = '<script>RLCONF={"wgPageName":"DF2014:Wild_boar","wgBreakFrames":!0,"wgUserName":null};RLSTATE={};</script>'
```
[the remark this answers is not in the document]
KunalSiyag: This was a valid suggestion. The test case didn't apply for the catch block and only the fallback was working. Handled that locally and pushed the updates.

KunalSiyag:
I’ve run the unit tests, and everything appears to be working correctly. there’s no longer any “Exceeded timeout of 5000 ms for a test” error being thrown as I saw in the PR checks before.
image

KunalSiyag:
> I feel like we are reinventing the wheel, can you develop why we cannot use Axios parsing logic?
> 
> I feel like `silentJSONParsing: false` was a correct setting, and that changing `forcedJSONParsing` from `true` to `false` could be sufficient to benefit from Axios logic to decide when content needs to be parsed from JSON.

Axios still parses whenever `responseType === 'json'` and with `silentJSONParsing: false` parse failures are still wrapped as `ERR_BAD_RESPONSE`. 

Why responseType: 'text' seems to be good

- It prevents Axios from entering the JSONRequested parse branch.
- We get raw body + status + content-type back.

Then mwoffliner decides:

- successful JSON response -> parse manually
- malformed successful JSON -> retry
- HTML 404/429/503 -> preserve old error handling

That is why I switched this path to raw text and moved JSON parsing into `getJSONCb()`: it restores access to the original body/status/content-type for old error handling, while still retrying true malformed JSON parse failures.

test/unit/util/articles.test.ts
```diff
@@ -222,4 +222,24 @@ describe('articles utility', () => {
     }
     expect(extractJsConfigVars(fakeHeadHtml)).toEqual(result[mwVersion])
   })
+
+  test('extractJsConfigVars falls back to API jsconfigvars when headHtml config is not strict JSON', () => {
+    const headHtml = '<script>;RLCONF={"wgPageName":"DF2014:Wild_boar","wgBreakFrames":!0,"wgUserName":null};RLSTATE={};</script>'
```
Markus-Rost: jsConfigVars using minified boolean values are already being tested above. Also this tests assumption about the API provided jsconfigvars is incorrect. (those are additional variables, not a fallback)

KunalSiyag: Good point. I checked out the extra test is redundant, we already cover that in 1.35.11, and the wording around `jsconfigvars` fallback is misleading since we don't have any API provided. I’ll drop that test.

benoit74:
I'll merge this now, @Markus-Rost feel free to speak up if you find something which needs to still be enhanced, but the fact this is failing other CIs is just too painful

Issue #2663, opened 2026-03-18
Improve test coverage for MW API, Redis utilities , and edge-case handling


RishabhThakur-19:
### Description

The current test suite lacks sufficient edge case coverage for several critical utility functions and components. This can lead to undetected bugs in uncommon but valid scenarios.

#### 1. RedisCompression
   

Redis keys are compressed as 
```
categories -> stored as c 
Title -> stored as n
and many more
```
in code we use full names ,if mapping breaks data gets stored incorrectly and some test may still pass ( because they dont check raw storage )

we need to verify : 
```
Data is storing redis using compressed keys
Data is retrieve using expanded keys 
Mapping is lossless
validate compression -decompression consistency  
```

#### 2. ArticleDetailXId

Specific Testcases to check it as a unit and not in aa integration test ,  
 verify : 
`set, get, getmany, delete,deletemany ,flush   `

#### 3. ExtractPagetitleFromHref:

currently  doesnt include  edge case that might be an problem:
add test cases for : 
```
Full Url
Encoded url 
Empty URl 
and other edge cases that has been not covered yet   
```

#### 4.GetDefaultSkin:
```
add Edge cases like:
when all skins are unusable 
skin array is empty
default skin exist but is unusable 
No default skin, usable skin exist
```

### Expected Outcome
* Improved reliability and robustness of the test suite
* Better handling of edge cases across API, Redis, and utility layers

PR #2667, opened 2026-03-20
fix (issue#2663) : improve test coverage  

RishabhThakur-19:
fixed issue #2663 

# Changes

- Created `redisCompression.test.ts` to validate Redis compression and ensure lossless mapping
- Added unit tests for ArticleDetailXId
- Covered edge cases for ExtractPageTitleFromHref
- Added edge case tests for GetDefaultSkin
- Improved GetDefaultSkin to ignore unusable default skins and log appropriate warnings
- Enhanced overall test reliability and robustness

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -14,6 +14,7 @@ Unreleased:
 * FIX: Handle categorylockdown error with placeholder instead of crashing (@triemerge #2150)
 * FIX: Improve mwUrl and adminEmail input validation and expanding test coverage for input sanitation (@RishabhThakur-19 #2657)
 * FIX: Recursively resolve CSS @import statements (@triemerge #2649)
+* FIX: Improve test coverage for MediaWiki.extractPageTitleFromHref and Rediscompression, also add unit tests for RedisStore.articleDetailXId (@RishabhThakur-19 #2663)
 * DEL: Remove deprecated CLI parameters `mwWikiPath`, `mwIndexPhpPath`, `keepEmptyParagraphs` (@Markus-Rost #2489)
 * NEW: Add `--customCss` CLI option to inject custom stylesheets
 
--- a/test/unit/mwApi.test.ts
+++ b/test/unit/mwApi.test.ts
@@ -128,6 +128,18 @@ describe('mwApi', () => {
     const absoluteIndexPathNoTitle = MediaWiki.extractPageTitleFromHref('/w/index.php?action=edit&redlink=1')
     // An index.php link missing the article title
     expect(absoluteIndexPathNoTitle).toBeNull()
+
+    const fullUrl = MediaWiki.extractPageTitleFromHref('https://en.wikipedia.org/wiki/Blue_whale')
+    // test for a complete  URL
+    expect(fullUrl).toBe('Blue_whale')
+
+    const encodedUrl = MediaWiki.extractPageTitleFromHref('/wiki/Radio_Rom%C3%A2nia_Actualit%C4%83%C8%9Bi')
+    // URL with special encoded characters
+    expect(encodedUrl).toBe('Radio_România_Actualități')
+
+    const emptyInput = MediaWiki.extractPageTitleFromHref('')
+    // Empty input should return null
+    expect(emptyInput).toBeNull()
   })
 })
 
--- a/test/unit/redis.test.ts
+++ b/test/unit/redis.test.ts
@@ -61,3 +61,129 @@ describe('Redis', () => {
     expect(flushedLen).toEqual(0)
   })
 })
+
+describe('RedisStore: articleDetailXId', () => {
+  beforeAll(startRedis)
+  afterAll(stopRedis)
+
+  const mockArticles = {
+    London: {
+      title: 'London',
+      categories: [{ title: 'City' }],
+      revisionId: 123,
+      coordinates: '51.5074,0.1278',
+    },
+    UK: {
+      title: 'United_Kingdom',
+      categories: [{ title: 'Country' }],
+      revisionId: 456,
+      thumbnail: { source: 'uk.png', height: 200, width: 200 },
+    },
+    TestArticle: {
+      title: 'TestArticle',
+      categories: [],
+      revisionId: 789,
+    },
+  }
+
+  beforeEach(async () => {
+    await RedisStore.articleDetailXId.flush()
+  })
+  //
+  test('set and get single article', async () => {
+    // if it fails that means serialization/deserialization is broken
+    const kvs = RedisStore.articleDetailXId
+
+    await kvs.set('London', mockArticles.London)
+    const result = await kvs.get('London')
+
+    expect(result).toBeDefined()
+    expect(result?.title).toBe('London')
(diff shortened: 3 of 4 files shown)


src/MediaWiki.ts
```diff
@@ -409,13 +409,26 @@ class MediaWiki {
...
-    if (defaultSkins.length == 0) {
-      throw new Error(`This wiki has no default skin:\n${JSON.stringify(skins)}`)
+    // Warn if any default skins are marked unusable
+    const unusableDefaults = skins.filter((s) => s.default && s.unusable)
+    if (unusableDefaults.length > 0) {
+      logger.warn(`Ignoring unusable default skin(s): ${unusableDefaults.map((s) => s.code).join(', ')}`)
     }
+
+    // Filter usable skins first
+    const usableSkins = skins.filter((skin) => !skin.unusable)
```
Markus-Rost: `getDefaultSkin` already receives a list filtered to only usable skins.

Should a wiki actually have their default skin marked as unusable, the default value would most likely overwrite the unusable skip. However that's a misconfiguration on the wiki side anyway which we shouldn't add special support for.

Multiple default skins are technically not possible.

RishabhThakur-19: @Markus-Rost  I dont think it will bring any problem , it will only make it more robust . and 
to answer the question that should a wiki actually have their default skin as unusable ,  it is uncommon in wiki production to have a default skin as unusable , but third party or misconfigured  Mediawiki instance may mark a default as unusable  

WDYT ???

Markus-Rost: It's running the unusable filter unnecessarily for a second time and if we actually wanted to support this misconfiguration of the wiki, the correct approach would be to return the default skin even if it is marked as "unusable" instead of throwing an error like we do right now. 

Note that "unusable" just means the skin is skipped in the user preferences list, so users can't decide to use it as their personal default skin. The skin might technically still work, though custom styling might be broken as the wiki editors don't expect it being used for reads.

RishabhThakur-19: Ok I got your point, and I will bring appropriate changes to it.

test/unit/mwApi.test.ts
```diff
@@ -210,7 +238,27 @@ describe('Mediawiki utils', () => {
         { code: 'modern', name: 'Modern', unusable: true },
       ],
     ],
-  ])('Get skin', (skins: SiteInfoSkin[]) => {
+    [
+      // All skins unusable
```
benoit74: This configuration does not exits "in the wild", it would make the wiki absolutely unusable, it makes no sense to adapt our code and tests for that.

test/unit/mwApi.test.ts
```diff
@@ -210,7 +238,27 @@ describe('Mediawiki utils', () => {
...
+    [
+      // Empty array
+      [],
+    ],
+  ])('Get skin (error cases)', (skins: SiteInfoSkin[]) => {
+    expect(() => MediaWiki.getDefaultSkin(skins)).toThrow()
+  })
+  test('no default but usable skins exist', () => {
+    const skins = [
+      { code: 'vector', name: 'Vector' },
```
benoit74: This configuration does not exits "in the wild", it would make the wiki unusable, it makes no sense to adapt our code and tests for that.

test/unit/mwApi.test.ts
```diff
@@ -197,6 +209,22 @@ describe('Mediawiki utils', () => {
...
+    [
+      // Ignore unusable default, pick valid one
+      [
+        { code: 'vector-2022', name: 'Vector (2022)', default: true, unusable: true },
+        { code: 'vector', name: 'Vector Legacy', default: true },
+      ],
+      'vector',
+    ],
+    [
+      // Default exists but is unusable
```
benoit74: This configuration does not exits "in the wild", it would make the wiki absolutely unusable, it makes no sense to adapt our code and tests for that.

test/unit/mwApi.test.ts
```diff
@@ -197,6 +209,22 @@ describe('Mediawiki utils', () => {
       ],
       'vector-2022',
     ],
+    [
+      // Ignore unusable default, pick valid one
```
benoit74: This configuration does not exits "in the wild", it would make the wiki absolutely unusable, it makes no sense to adapt our code and tests for that.

RishabhThakur-19:
> I let @Markus-Rost confirm since he has more experience than I do, but to me all the tests cases you've added around skins do not exist in the wild. A Mediawiki can have only one default skin, and it must be usable. Otherwise the wiki just doesn't work. So no-one will even try to scrape it. Displaying a message in the code when there are two defaults is already not something really required at all. We should not make that logic even more complex for no added value.

@Markus-Rost, @benoit74 Thank you for pointing out that my approach had gone beyond the intended scope. I will revert all the related changes.

@Markus-Rost also pointed out that multiple default skins are technically not possible. I had tried to justify my approach based on the existing logic that seemed to handle multiple defaults before creating this PR, which led me in that direction. I appreciate the clarification and will ensure that future work stays aligned with the project requirements.

test/unit/mwApi.test.ts
```diff
@@ -128,6 +128,18 @@ describe('mwApi', () => {
     const absoluteIndexPathNoTitle = MediaWiki.extractPageTitleFromHref('/w/index.php?action=edit&redlink=1')
     // An index.php link missing the article title
     expect(absoluteIndexPathNoTitle).toBeNull()
+
+    const fullUrl = MediaWiki.extractPageTitleFromHref('https://en.wikipedia.org/wiki/Blue_whale')
+    // Full absolute URL
+    expect(fullUrl).toBe('Blue_whale')
+
+    const encodedUrl = MediaWiki.extractPageTitleFromHref('/wiki/Blue%20whale')
```
benoit74: This is not a realistic href, white spaces do not exists AFAIK in Mediawikis, they are replaced by underscores. But other characters are indeed escaped, you can test for `/wiki/Radio_Rom%C3%A2nia_Actualit%C4%83%C8%9Bi` for instance (found on Wikipedia EN).

Changelog
```diff
@@ -10,6 +10,7 @@ Unreleased:
 * FIX: Handle categorylockdown error with placeholder instead of crashing (@triemerge #2150)
 * FIX: Improve mwUrl and adminEmail input validation and expanding test coverage for input sanitation (@RishabhThakur-19 #2657)
 * FIX: Recursively resolve CSS @import statements (@triemerge #2649)
+* FIX: Improve test coverage for MWAPI, Redis utilities, and edge-case handling (@RishabhThakur-19 #2663)
```
benoit74: I don't think this PR is testing MWAPI anymore, and `edge-case handling` is too precise and blury at the same, testing edge cases is de-facto the purpose of testing. If you really don't know how to describe other changes, just saying `and others` is more concise and sufficient. But I feel like you can propose a better changelog

RishabhThakur-19:
sorry for not responding early ,
I have tried to solve the requested changes from my side , if still there is any change you want , please do let me know

PR #2668, opened 2026-03-21
Remove unused mocha from production dependencies

nishantharkut:
mocha (v11.7.5) is listed under the dependencies and not the devDependencies but the project uses jest for all the testing work. There are zero mocha imports anywhere in the codebase, no config files related to it and no mocha style test files.

so this means that every npm install mwoffliner downloads mocha and its entire dependency without any use.

Files changed : 
- package.json
- package-lock.json (ran npm i to update it)


benoit74 (changes requested):
Thank you. Please add an entry in CHANGELOG file

nishantharkut:
> Thank you. Please add an entry in CHANGELOG file

I have added the entry in the Changelog in the latest commit. Thank you for the review.

nishantharkut:
Resolved the merge conflicts.

PR #2669, opened 2026-03-21
replace loose equality with strict equality

nishantharkut:
there is a strict equality used in the rest of the codebase except the modified files. 
Hence, All 11 remaining "**==**" comparisons are replaced with the "**===**". 
Every case compares the same type operands so the behaviour is not changed at all.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,6 @@
 Unreleased:
 * FIX: Preserve inline data: URL images instead of rewriting them (@nishantharkut #2670)
+* FIX: Replace loose equality with strict equality (@nishantharkut #2669)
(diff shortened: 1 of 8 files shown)


Markus-Rost (approved):

nishantharkut:
Added the entry in the Changelog in the latest commit. Thank you

nishantharkut:
Resolved the merge conflicts.

benoit74:
> Resolved the merge conflicts.

Unless explicitly requested for because we are afraid of a potential real conflict, there is no need to do that during review, especially because we do not want merge from feature branch to main by convention.

Code looks ok to me, so I rebased on top of main and I'll wait for CI to complete before merging

Issue #2653, opened 2026-03-11
Allow to set User-Agent


benoit74:
Currently the scraper sets its own UA.

It looks like Anubis in its default config allows `curl/xx` UA to bypass Anubis ... and most folks are too lazy to dive into config and configure Anubis differently (see https://github.com/openzim/zim-requests/issues/500#issuecomment-4017284607)

We probably have other scenario where setting the UA would help.

We should at least allow to customize UA for all requests. Maybe more, to be investigated (how does setting the UA to "curl/xxx" allows a better ZIM of postmarket OS)

kelson42:
See that feature as an opportunity to clearly show to the user what is the default UA.

KunalSiyag:
We can add a support for a --userAgent flag to allow overriding the default User-Agent used by mwoffliner and show the UA used in the logs. I would like to work on this.

benoit74:
Fixed by #2673

PR #2673, opened 2026-03-23
 Allow custom User-Agent header via --userAgent flag

KunalSiyag:
#2653

<hr/>

Adds a --userAgent CLI parameter to allow overriding the HTTP User-Agent string used for all requests.

When omitted, the existing default userAgent is preserved. The active UA is now always logged at startup so users can see exactly what is being sent.

image

Changes

src/parameterList.ts : Added userAgent parameter description

src/mwoffliner.lib.ts : Wired the new flag with fallback to the default UA; added startup log line showing the active User-Agent

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -106,6 +106,7 @@ async function execute(argv: any) {
     forceSkin,
     langVariant,
     customCss,
+    userAgent: _userAgent,
   } = argv
 
   let { articleList, articleListToIgnore } = argv
@@ -180,9 +181,13 @@ async function execute(argv: any) {
   const addModules = _addModules ? String(_addModules).split(',') : []
   const trustedJs = javaScript === 'none' ? null : javaScript === 'trusted' ? config.output.mw.js_trusted.concat(addModules) : []
 
+  /* HTTP user-agent string */
+  const uaString = `${_userAgent || config.userAgent} (${adminEmail})`
+  logger.log(`Using User-Agent: ${uaString}`)
+
   /* Download helpers; TODO: Merge with something else / expand this. */
   Downloader.init = {
-    uaString: `${config.userAgent} (${adminEmail})`,
+    uaString,
     speed,
     reqTimeout: requestTimeout * 1000 || config.defaults.requestTimeout,
     optimisationCacheUrl,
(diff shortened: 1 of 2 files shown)


src/mwoffliner.lib.ts
```diff
@@ -180,9 +181,13 @@ async function execute(argv: any) {
   const addModules = _addModules ? String(_addModules).split(',') : []
   const trustedJs = javaScript === 'none' ? null : javaScript === 'trusted' ? config.output.mw.js_trusted.concat(addModules) : []
 
+  /* HTTP user-agent string */
+  const uaString = _userAgent || `${config.userAgent} (${adminEmail})`
+  logger.log(`Using User-Agent: ${uaString}`)
```
Markus-Rost, line 186: The user agent is the only place the admin email is used. Considering the admin email is a required parameter, we should always append it to the user agent including custom user agents.

KunalSiyag: Will work on appending it to the custom agents too.

benoit74: @Markus-Rost wouldn't it make more sense to make either adminEmail or custom user agent mandatory? I find it quite deceptive to let user specify the user-agent he wanna use only to modify it afterwards. And I'm quite sure we will have cases where we need complete control on the user-agent to not be blocked by any form of WAF

KunalSiyag: Maybe we could do something like:

When `--userAgent `is provided, use it as-is, no modification.` --adminEmail` becomes optional.
When `--userAgent` is not provided -> keep the current default MWOffliner/<version> (<adminEmail>). `--adminEmail `stays required.

benoit74: Yes, this is what I feel makes more sense to me. But I would like to have @Markus-Rost feedback on this idea as well

Markus-Rost: @benoit74 adminEmail is provided as a way for the server admins to contact the mwoffliner user if there are any issues. Keeping it required and part of even custom UAs ensures that is always the case.

While I'm aware of some anti-bot WAF settings which don't allow for an IP/UA whitelist, I think the small possibility of remaining to get blocked by them is preferably over the option of allowing mwoffliner to run with a fully generic UA. ([GenAI is already doing that and causing excessive headaches for server admins everywhere](https://weirdgloop.org/blog/clankers))

benoit74: OK, let's start with code as-is, and we will change if we realize we have good reasons to do so

PR #2674, opened 2026-03-23
Deduplicate queued file paths with RedisSet

triemerge:
Follow-up to discussion in [#2656 (comment)](https://github.com/openzim/mwoffliner/pull/2656#discussion_r2958501827).

`queuedFilePathsSet` now deduplicates file paths across repeated `queuePendingFilesToDownload` polls using a Redis-backed `RedisSet`, keeping large dedup state out of Node.js memory.

- Adds `RedisSet` abstraction alongside `RedisKvs` and `RedisQueue`
- Registers `queuedFilePathsSet` in `RedisStore` with full lifecycle handling (`checkForExistingStores`, `close`)
- Uses `SADD` return value as atomic enqueue gate
- First call flushes the file store; subsequent calls delete only the processed batch
- Fixes fire-and-forget `RedisStore.close` by awaiting it in `src/mwoffliner.lib.ts`
- Makes stale Redis cleanup deterministic (await per-key delete) in `src/RedisStore.ts`
- Ensures `queuedFilePathsSet` is flushed even when download processing fails in `src/util/saveArticles.ts`
- Fixes host completion accounting logic in `src/util/saveArticles.ts`
- Adds polling scan throttling + first-scan guarantee in `src/util/saveArticles.ts`
- Avoids retrying non-transport processing errors in `src/util/saveArticles.ts`
- Fixes async setup race in `RedisSet` unit test in `test/unit/util/RedisSet.test.ts`

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -17,6 +17,7 @@ import { fileDownloadMutex, zimCreatorMutex } from '../mutex.js'
 import { truncateUtf8Bytes } from './misc.js'
 import { isMainPage } from './articles.js'
 import RedisQueue from './RedisQueue.js'
+import { processStylesheetContent } from './dump.js'
 
 // Maximum delay between file download attempts on a given host
 // Default upload.wikimedia.org Retry-After value is 11 seconds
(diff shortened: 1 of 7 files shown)


src/util/saveArticles.ts
```diff
@@ -34,38 +35,68 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
...
-          downloadFailure: 0,
-          downloadsComplete: false,
+
+  const queuePendingFilesToDownload = async (isFirstCall = false) => {
+    const queuedPathsBatch: string[] = []
+    await fileStore.iterateItems(1, async (filesToDownload: KVS<FileDetail>) => {
+      for (const [path, { url, mult, width, kind }] of Object.entries(filesToDownload)) {
+        const pathWasQueuedNow = await RedisStore.queuedFilePathsSet.add(path)
+        if (!pathWasQueuedNow) {
+          continue
```
Markus-Rost, line 53: Should probably show a warning here, ideally including the url and width of both entries.

triemerge: > Should probably show a warning here

Added a warning on duplicate paths with the duplicate URL and width.

> ideally including the url and width of both entries.

Including URL/width for the original entry is not reliably available in the current flow, since that metadata is not retained for duplicate checks. Retaining it would may require a separate Redis hash. Happy to add that if you think it's worth it, but wanted to check before adding the complexity.

triemerge:
> I'm a bit worried about the throttling of file store scans. I don't get what allows us to ensure we do not end up in a situation where all files has been downloaded from `hosts` perspective so we stop downloading files, but we still have few files very recently added to the file store but not yet added to the `hosts`.

You're right about the throttling, there is a small window where hosts can mark as complete before late-added files get polled. I can add a final poll after the completion check to close that gap.

> With some step-back, I also feel like we are really over-complicating the whole thing. The fact that we have both `filesToDownloadXPath` and `filesQueues`, and that we moved files from the first to the second was a convenience. And we transferred items from the first to the second to reduce memory usage. But with this PR we now add back paths to an set ... I feel like this does not make real sense and is now really over-complicating everything.

On the complexity, I think the current approach is the right trade-off. The three stores serve different purposes: `filesToDownloadXPath` for resolution upgrades, `queuedFilePathsSet` as a path-only dedup gate (lives in Redis, no Node.js memory), and `filesQueues` for per-host download scheduling.

> One solution I imagine is that whenever we want to add a file to process, we simply immediately add it to both the `filesToDownloadXPath` and the `filesQueues`. And we never clear the `filesToDownloadXPath` during processing. Would this work? To simplify logic, it is probably worth to move this to a dedicated class which will provide few primitives like "addFileToProcess", "getNextFileToDownload", etc ...

Adding directly to both `filesToDownloadXPath` and `filesQueues` at discovery time would be simpler, but `saveArticle` and `processStylesheetContent` don't know about host queues today. The dedicated class with `addFileToProcess` / `getNextFileToDownload` primitives would solve that coupling and make the API cleaner, but it's a bigger change that touches how all file discovery sites interact with the download pipeline. I think the safest path is to land this with the race fix, and tackle the class refactor as a follow-up with its own review. 

WDYT?

benoit74:
Most of this PR changes will be reverted by the change to the other approach (no need for RedisSet anymore, etc ...).

Please open:
- one PR with the changes not linked to deduplication at all (avoid retrying non-transport processing errors in `src/util/saveArticles.ts`, ...) ; maybe on PR per issue would even help review, I don't feel like these are really interconnected topics)
- one PR solving the duplicate queued file paths with the approach I proposed

With this, we will be in a better shape to make an informed decision. Sorry for the additional work, but this is a sensible part of the codebase I do not want to break.

triemerge:
> Please open:
> 
> * one PR with the changes not linked to deduplication at all (avoid retrying non-transport processing errors in `src/util/saveArticles.ts`, ...) ; maybe on PR per issue would even help review, I don't feel like these are really interconnected topics)
> * one PR solving the duplicate queued file paths with the approach I proposed
> 

Thanks for the clear direction. Makes sense, I'll close this PR and split it up as suggested. Will open the unrelated fixes as separate PRs first, then a fresh one for the dedup approach you proposed.

PR #2675, opened 2026-03-23
Detect continuation cycles in article pagination

### LINKED ISSUE #2532: Detect infinite `continue` loops
In https://github.com/openzim/zim-requests/issues/1260#issuecomment-3324636008, we've faced an infinite loop while following continue parameters.

I don't know if this is worth it, but some kind of logic detecting that we are in such an infinite loop would have greatly helped.

Technically speaking, one could hash URLs retrieved so far, and check that new URL to fetch is not among the list of hash previously visited for the given query (batch of articles typically here). This is pretty easy to implement.

But I'm not sure we have a high probability to face such a bug again. And there are many scenario where we might but not in an indefinite loop but something way too big, just offsetting by few `continue` value at a time.

Maybe simply putting an upper bound on the number of URLs to visit for a single query would be sufficient. Something relatively big like 1000 would give plenty room for normal queries to complete all while avoiding infinite loop to continue "forever".

Or maybe this is just accidental complexity for a bug we've very seldom encountered, and not worth to implement.

triemerge:
Fix #2532

`getArticleDetailsIds` and `getArticleDetailsNS` follow MediaWiki `continue` parameters in unbounded `while(true)` loops. `getArticlesByNS` follows the returned `gapContinue` token in an unbounded `do...while` loop. A malformed token (as seen in [openzim/zim-requests#1260 (comment)](https://github.com/openzim/zim-requests/issues/1260#issuecomment-3324636008)) can create a cycle that hangs Zimfarm indefinitely. Both methods now track seen continuation states in a per-call `Set` and throw if a cycle is detected.

- `Set` is scoped to each invocation so memory stays bounded
- Throws rather than warns since a detected cycle means the API response cannot be trusted

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -374,6 +374,7 @@ class Downloader {
   public async getArticleDetailsIds(articleIds: string[], shouldGetThumbnail = false): Promise<QueryMwRet> {
     let continuation: ContinueOpts
     let finalProcessedResp: QueryMwRet
+    const visitedUrls = new Set<string>()
 
     while (true) {
       const queryOpts: KVS<any> = {
@@ -390,6 +391,10 @@ class Downloader {
       }
 
       const reqUrl = this.apiUrlDirector.buildQueryURL(queryOpts)
+      if (visitedUrls.has(reqUrl)) {
+        throw new Error(`Detected continuation cycle while fetching article details by IDs. ` + `visitedUrls=[\n${[...visitedUrls].join('\n')}\n]`)
+      }
+      visitedUrls.add(reqUrl)
 
       const resp = await this.getJSON<MwApiResponse>(reqUrl)
 
@@ -415,6 +420,7 @@ class Downloader {
     let queryContinuation: QueryContinueOpts
     let finalProcessedResp: QueryMwRet
     let gCont: string = null
+    const visitedUrls = new Set<string>()
 
     while (true) {
       const queryOpts: KVS<any> = {
@@ -442,6 +448,10 @@ class Downloader {
       }
 
       const reqUrl = this.apiUrlDirector.buildQueryURL(queryOpts)
+      if (visitedUrls.has(reqUrl)) {
+        throw new Error(`Detected continuation cycle while fetching article details in namespace ${ns}. ` + `visitedUrls=[\n${[...visitedUrls].join('\n')}\n]`)
+      }
+      visitedUrls.add(reqUrl)
 
       const resp = await this.getJSON<MwApiResponse>(reqUrl)
       Downloader.handleMWWarningsAndErrors(resp)
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -101,6 +101,7 @@ export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], allow
   return new Promise(async (resolve, reject) => {
     let totalArticles = 0
     let chunk: { articleDetails: QueryMwRet; gapContinue: string }
+    const seenGapContinueValues: string[] = []
 
     const { articleDetailXId, redirectsXId } = RedisStore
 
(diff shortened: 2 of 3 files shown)


triemerge:
> The `chunk.gapContinue` value in `getArticlesByNS` at `mw-api.ts` needs this same treatment as well.

Thanks for pointing that out, added the same cycle detection to `chunk.gapContinue` in `getArticlesByNS`.

src/util/mw-api.ts
```diff
@@ -129,6 +130,13 @@ export function getArticlesByNS(ns: number, articleIdsToIgnore?: string[], allow
         curStage = 0
         chunk = await Downloader.getArticleDetailsNS(ns, chunk && chunk.gapContinue)
 
+        if (chunk.gapContinue) {
+          if (seenGapContinueStates.has(chunk.gapContinue)) {
+            throw new Error(`Detected continuation cycle while fetching namespace articles. gapContinue=${chunk.gapContinue}`)
```
benoit74: This is so rare that it is worth to output the NS and the whole list of `gapContinue` we've seen which ultimately lead to a cycle.

src/Downloader.ts
```diff
@@ -374,8 +374,15 @@ class Downloader {
   public async getArticleDetailsIds(articleIds: string[], shouldGetThumbnail = false): Promise<QueryMwRet> {
     let continuation: ContinueOpts
     let finalProcessedResp: QueryMwRet
+    const seenContinuationStates = new Set<string>()
 
     while (true) {
+      const continuationState = JSON.stringify(continuation || {})
+      if (seenContinuationStates.has(continuationState)) {
+        throw new Error(`Detected continuation cycle while fetching article details by IDs. continuation=${continuationState}`)
```
benoit74: Same remark, but this time it would be even better to output the list of URLs we have visited (with all parameters).

src/Downloader.ts
```diff
@@ -415,8 +422,18 @@ class Downloader {
...
     let gCont: string = null
+    const seenContinuationStates = new Set<string>()
 
     while (true) {
+      const continuationState = JSON.stringify({
+        gapcontinue,
+        queryContinuation: queryContinuation || {},
+      })
+      if (seenContinuationStates.has(continuationState)) {
+        throw new Error(`Detected continuation cycle while fetching article details by namespace. continuation=${continuationState}`)
```
benoit74: Same remark, but this time it would be even better to output the list of URLs we have visited (with all parameters).

triemerge:
> Small remarks, let me know if it does not makes sense to you

Makes total sense, having the full URL sequence and namespaces in the error will definitely help with debugging these loops. I'll get those added.

src/util/mw-api.ts
triemerge: The error now includes the `ns` number and the full array of all seen `gapContinue` values, which should make reconstructing the loop much easier.

src/Downloader.ts
triemerge: Added a `visitedUrls` array to track every request URL right before it's sent. The error now dumps the full list including all query parameters if a cycle is hit.

src/Downloader.ts
triemerge: Done here as well, using the same `visitedUrls` tracking

triemerge:
> Is there an added value to have both `seenContinuationStates` and `visitedUrls` variables? Would it be possible to only have `visitedUrls` as a set and use that for cycle detection? I feel like it could make the code easier to read

Good point. I updated both functions to rely purely on a visitedUrls Set for cycle detection. It definitely makes the code easier to read, also it now avoids the JSON.stringify key-ordering issue

benoit74:
@triemerge please fix QA issue

triemerge:
> @triemerge please fix QA issue

My bad, formatting is now fixed.

PR #2676, opened 2026-03-24
feat: add --mainPageLongDescription flag to populate ZIM LongDescription

### LINKED ISSUE #2384: Use `<meta name="description" content="">` DOM node to fill ZIM `LongDescription` Metadata

Aunirbhan:
## Problem
ZIM `LongDescription` metadata was never populated by mwoffliner. Closes #2384 

## Solution

Added `--mainPageLongDescription` flag (off by default):

1. Fetches the Wikidata description via `action=query&prop=description` — single API call that respects `maxlag`
2. HTML fallback using `<meta name="description">` for non-Wikimedia wikis
3. `--customZimLongDescription` always takes priority

Uses the API rather than HTML scraping as the primary approach because `<meta name="description">` is absent from all standard MediaWiki installs (tested: Wikipedia, Wiktionary, mediawiki.org, Arch Wiki, OSM Wiki). The API returns structured Wikidata descriptions reliably on all Wikimedia wikis. I used a flag to prevent downstream tooling from being affected since `LongDescription` was always empty before. 

## Changes

| File | Change |
|---|---|
| `src/parameterList.ts` | Registers `--mainPageLongDescription` flag |
| `src/mwoffliner.lib.ts` | Fetches description via API with HTML fallback; updates metadata line |

## Testing

- 31/33 test suites pass
- 2 pre-existing failures (`downloader.test.ts`, `webpAndRedirection.test.ts`) — reproduce identically on unmodified upstream

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,7 @@
 Unreleased:
+* UPDATE: Replace iFrame with a placeholder card instead of deleting them (@kunalsiyag #2481)
 * NEW: Scrape only wikitext contentmodel articles by default (@triemerge #2445)
+* FIX: In nopic mode, fixed layout issues by preserving image layout using SVG placeholders (@kunalsiyag #1004)
 * NEW: Add support for ResourceLoader based JavaScript (@Markus-Rost #2310)
 * CHANGED: Keep only ActionParse renderer (@benoit74 #2210)
 * CHANGED: Keep inline scripts if all JS is allowed (@Markus-Rost #2555)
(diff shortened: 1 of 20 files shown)


benoit74:
@Aunirbhan I did not looked into your code at all, because we already have a pending PR for that issue : https://github.com/openzim/mwoffliner/pull/2641 ; if you feel like you have something worth to add / change in this PR, then please contribute to the PR by indicating what could be done better / differently. Or if you feel like this other PR is simply wrong and should be done differently, then discuss about it in the PR or the issue before pushing another PR with a different approach which could help us weight alternatives and decide. I can't spend time reviewing duplicate work, and it is against the spirit of open source collaboration.

benoit74:
Closing in favor of #2641

PR #2681, opened 2026-03-27
Fix fire-and-forget async patterns and file download accounting bugs

triemerge:
- Replaced all `forEach(async` callbacks with `await Promise.all(items.map(async ...))` to properly await async operations
- Added missing `await` on `RedisStore.close()`, `filesToDownload.flush()`, and `client.del()`
- Fixed `queue.flush` referenced as a property instead of called as `queue.flush()`
- Fixed `completedHosts` reduce accumulator reset bug
- Added `filesTotal > 0` guard to prevent division-by-zero in download progress accounting

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,4 +1,5 @@
 Unreleased:
+* FIX: Fix fire-and-forget async patterns and file download accounting bugs (@triemerge #2681)
 * FIX: Preserve inline data: URL images instead of rewriting them (@nishantharkut #2670)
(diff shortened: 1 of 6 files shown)


triemerge:
> Please add a changelog entry so that it is more obvious in which version we've fixed these issues.

Done, added the changelog entry (also the formatting issue has now been fixed).

benoit74:
I've added PR number in changelog (we do that when we do not have an issue for the PR)

Issue #2637, opened 2026-03-04
Some `maxi` ZIM creation continue to freeze


benoit74:
- https://farm.openzim.org/pipeline/40ae9049-960a-45c3-b659-3883bd74e80d
- another one I forgot to write down ...

RishabhThakur-19:
I think the problem is that the build is very slow  because large  amount of media is being downloaded , which results in many `HTTP 429` ,and the current retry mechanism in mwoffliner is using backoff per-request for example if 1000 images are getting downloaded  then 1000 http 429 that means we have 1000 exponential backoff running independently ( each retrying up to 6 min) , each  retries on its scheduled time thus resulting  in  a loop due large amount of media   

<img width="1159" height="313" alt="Image" src="https://github.com/user-attachments/assets/ff64dd51-a899-4456-b97e-2e1c1a742d72" />

I am still analyzing the scraper logs for better understanding .
i will share more detailed findings shortly

RishabhThakur-19:
<img width="1631" height="734" alt="Image" src="https://github.com/user-attachments/assets/e3195482-1f21-408c-ae10-96871a2ef4e4" />  

my bad  , this is not a exponential backoff problem beacuse the files were skipped immediately rather than retried with exponential backoff. There are no repeated attempts for the same URLs, no increasing delay patterns, and no evidence of long retry cycles

RishabhThakur-19:
`[2026-02-28 03:03:17,383: INFO] starting zimfarm task-worker for 40ae9049-960a-45c3-b659-3883bd74e80d.`
`[2026-02-28 03:03:19,756: INFO] Starting scraper. Expects files at: /data/volume1/zimfarm/data/40ae9049-960a-45c3-b659-3883bd74e80d `
`[2026-02-28 14:46:59,706: INFO] ZIM file created: wikipedia_hu_all_mini_2026-02.zim, 768.92 MiB`
`[2026-02-28 14:49:00,408: INFO] Updating file-status=uploaded for wikipedia_hu_all_mini_2026-02.zim`
`2026-02-28 22:35:49,522: INFO] ZIM file created: wikipedia_hu_all_nopic_2026-02.zim, 2.96 GiB`
`2026-02-28 22:35:50,803: INFO] Starting zim checker for wikipedia_hu_all_nopic_2026-02.zim`
[...]

PR #2683, opened 2026-03-27
fix(downloader): correctly increment completed-hosts count in getNextFileToDownload

Pranav140:
### **Description**
This PR fixes a logic bug in the `reduce` accumulator inside [getNextFileToDownload](cci:1://file:///f:/mwoffliner/mwoffliner/src/util/saveArticles.ts:69:2-132:3) (within [src/util/saveArticles.ts](cci:7://file:///f:/mwoffliner/mwoffliner/src/util/saveArticles.ts:0:0-0:0)).

Previously, the accumulator was reset to `0` whenever a single non-completed host was encountered in the `hosts` array. This caused the early-exit check (`completedHosts === hostValues.length`) to almost never be true as long as there were multiple hosts and not all were already finished. As a result, workers would often continue polling and spinning even after all downloads were functionally complete, only exiting when the 1-hour safety-net timeout was reached.

The fix ensures that the accumulator (`buf`) is preserved when a host is not completed, rather than resetting the total count to zero.

### **Impact**
- **Fixes unnecessary polling:** Workers will now exit the download loop immediately once all hosts are marked as complete.
- **Addresses Pipeline Freezes:** Directly linked to issues like #2637, where `maxi` ZIM creations appear to "freeze" for long periods despite finishing most tasks.
- **Efficiency:** Reduces idle CPU usage and ensures cleaner termination of the scraping process.

### **Changes**
- **[src/util/saveArticles.ts](cci:7://file:///f:/mwoffliner/mwoffliner/src/util/saveArticles.ts:0:0-0:0)**: Modified the `reduce` callback to return `buf` instead of `0` when `host.downloadsComplete` is false.
- **[test/unit/downloadFiles.test.ts](cci:7://file:///f:/mwoffliner/mwoffliner/test/unit/downloadFiles.test.ts:0:0-0:0)**: Added a new suite of regression tests that isolate and verify the count logic for mixed host states (complete/incomplete).

### **Testing**
I have added a new unit test file specifically for this logic. Since the fix is in a private closure, the tests verify the identical `reduce` logic in isolation to ensure deterministic results.

**To run the tests:**
`npm run test:unit test/unit/downloadFiles.test.ts`

### **Checklist**
- [x] Correctly identifies the bug in the `reduce` accumulator.
- [x] Includes a regression test to prevent future occurrences.
- [x] Touches only the necessary lines to keep the PR focused and surgical.
- [x] Addresses potential causes for long-running "freezes" reported in existing issues.

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -97,7 +97,7 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
       // check if all donwloads have completed and exit
       const hostValues = Array.from(hosts.values())
       const completedHosts = hostValues.reduce((buf, host) => {
-        return host.downloadsComplete ? buf + 1 : 0
+        return host.downloadsComplete ? buf + 1 : buf
       }, 0)
       if (completedHosts === hostValues.length) {
         return null
--- a/test/unit/downloadFiles.test.ts
+++ b/test/unit/downloadFiles.test.ts
@@ -1,0 +1,38 @@
+/**
+ * Regression tests for the reduce-accumulator bug in getNextFileToDownload()
+ * (src/util/saveArticles.ts).
+ *
+ * The old code:  return host.downloadsComplete ? buf + 1 : 0
+ * Reset the accumulator to 0 on any non-complete host, making the early-exit
+ * check `completedHosts === hostValues.length` almost never true.
+ *
+ * The fix:       return host.downloadsComplete ? buf + 1 : buf
+ */
+describe('downloadFiles: completed-hosts reduce logic', () => {
+  // Mirrors the fixed logic from saveArticles.ts
+  const countCompleted = (hosts: { downloadsComplete: boolean }[]) =>
+    hosts.reduce((buf, host) => (host.downloadsComplete ? buf + 1 : buf), 0)
+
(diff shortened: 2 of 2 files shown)


benoit74:
Thank you but this is already fixed in https://github.com/openzim/mwoffliner/pull/2681 and I'm kinda concerned this could be an attempt to overtake @triemerge work linked to the fact I've asked him to split his PR ... 

Tests should tests real codebase, not code in isolation. If we can't ... then we can't

> Previously, the accumulator was reset to 0 whenever a single non-completed host was encountered in the hosts array. This caused the early-exit check (completedHosts === hostValues.length) to almost never be true as long as there were multiple hosts and not all were already finished. As a result, workers would often continue polling and spinning even after all downloads were functionally complete, only exiting when the 1-hour safety-net timeout was reached.

Goal is to know when all hosts have completed. So to me it is true that there is a bug in the computed of number of completed hosts when there is still one host running. But once all hosts have completed, the `completedHosts` value is correct and the loop exits, so this bug had zero consequence. We never saw the 1h exit condition trigger in production.

For all reasons above, I'm going to close the PR for the time being. If you feel like you know how to enhance the codebase with the same tests but exercising the real codebase (and not at the expense of too complex code), please open another PR once #2681 is merged

PR #2685, opened 2026-03-27
Avoid retrying non-transport processing errors

triemerge:
`workerDownloadFile` previously retried most failures, including deterministic processing errors. It now generally treats processing failures such as `zimCreator.addItem` errors as non-transport failures and skips re-queueing.

This reduces unnecessary retries and queue churn while preserving retries for recognized transport/network errors.


benoit74:
Please add a changelog entry as well, so that we do not forget we've made this change in the version 2

triemerge:
> Please add a changelog entry as well, so that we do not forget we've made this change in the version 2

Done, added the changelog entry.

benoit74:
I've added PR number in changelog, we do that when we do not have an issue matching the PR

PR #2686, opened 2026-03-28
fix: gradually recover requestInterval after successful downloads

### LINKED ISSUE #2572: Very long retry intervals

EDWARD-012:
## Problem

Fixes #2572

When a server returns HTTP 429/503/524, `requestInterval` is multiplied by `1.2` (capped at `MAXIMUM_FILE_DOWNLOAD_DELAY` = 20 seconds). However, after the server recovers and downloads succeed again, `requestInterval` **never decreases** — the scraper stays permanently throttled at the maximum interval for that host, making it appear frozen.

## Fix

On each successful download, gradually reduce `requestInterval` by multiplying by `0.9`, with a floor of `30ms` (the initial value):
```ts
hostData.requestInterval = Math.max(30, hostData.requestInterval * 0.9)
// gradually recover interval on success
```

This mirrors the existing slow-down logic symmetrically — the interval increases on failure and recovers on success.

## Changes

- `src/util/saveArticles.ts`: Add interval recovery on successful download + update comment

### DIFF
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -162,6 +162,7 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
           await zimCreatorMutex.runExclusive(() => zimCreator.addItem(item))
           dump.status.files.success += 1
           hostData.downloadSuccess += 1
+          hostData.requestInterval = Math.max(30, hostData.requestInterval * 0.9) // gradually recover interval on success
         } else {
           throw new Error(`Bad response received: ${resp}`)
         }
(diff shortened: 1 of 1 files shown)


benoit74:
The scraper does not appear frozen, it pace its requests at the interval found to be optimal to not have 429. 

Since we increase intervals by only 20% at each failed requests, we have high chances to find a "close to optimal" value.

And gradually recovering with a naive approach like in this PR has been experimented with but it proved to have detrimental side effects where the server finally issues more 429 because it feels like the scraper is too aggressive.

Only problem occurs when the server reply with 429 while it has not speed issues, which sets the interval to its maximum of 20s. But even in this case, recovering on success does not really help.

Unless you have more sound arguments than "it should better work like this", I will have to close this PR.

PR #2687, opened 2026-03-30
Restrict the list of available formats + add more meaningful labels

benoit74:
For some reasons, we've originally allowed more flavours / formats than what we really used in production

I've check prod DB and these are the sole values we use.

@Popolechien do I miss something or do you confirm these are the only formats we wanna use on Zimfarm?

@elfkuzco do you see any problem to have such labels?

Current:
image

After:
image

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -98,15 +98,10 @@
       "description": "Which flavours to build, as `<flavour>:<custom-suffix>`. Empty option is full without suffix.",
       "alias": "format",
       "choices": [
-        { "title": "NODET_NOPIC_MINI", "value": "nodet,nopic:mini" },
-        { "title": "NODET_MINI", "value": "nodet:mini" },
-        { "title": "NOPIC_NOPIC", "value": "nopic:nopic" },
-        { "title": "NOVID_MAXI", "value": "novid:maxi" },
-        { "title": "EMPTY", "value": "" },
-        { "title": "NODET", "value": "nodet" },
-        { "title": "NOPIC", "value": "nopic" },
-        { "title": "NOVID", "value": "novid" },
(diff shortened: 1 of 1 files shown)


benoit74:
The full list after this change:

image

Popolechien:
I believe some wikis (thinking of wikimed in particular) do have a significant amount of videos, so `novid maxi` is not the only "big" option we should have.

benoit74:
which other option do we have then?

Popolechien:
Ah my bad, for some reason I thought we had a video-compatible `maxi` version. I seem to recall we had a couple of wikis with youtube videos embedded, but I wonder if I'm hallucinating that ಠ_ಠ

benoit74:
> Ah my bad, for some reason I thought we had a video-compatible maxi version. I seem to recall we had a couple of wikis with youtube videos embedded, but I wonder if I'm hallucinating that ಠ_ಠ

I don't think this ever existed, and at least I would prefer `maxi` is ... `maxi` ಠ_ಠ

Popolechien:
Ah yes - this guy here : https://github.com/openzim/zim-requests/issues/1112#issuecomment-4135362615

Turns out that when no option is selected, MWoffliner does capture videos

elfkuzco:
There shouldn't be any issues with the labels since they're modeled as enums with the human readable text as the values. Just that the enum names should be valid Python identifiers

elfkuzco:
Turns out this is valid in Python 3.13. Thought those titles being invalid identifiers would break but it seems to work.

benoit74:
> Ah yes - this guy here : https://github.com/openzim/zim-requests/issues/1112#issuecomment-4135362615
>
> Turns out that when no option is selected, MWoffliner does capture videos

Yes, this is the "All" option above, which is indeed the default

PR #2689, opened 2026-03-30
Refactor file download pipeline to inline host queueing

triemerge:
Follow-up to discussion in [#2674 (comment)](https://github.com/openzim/mwoffliner/pull/2674#pullrequestreview-4019512661).

Extracts the `downloadFiles` block from `saveArticles.ts` into a dedicated `FileManager` class. 

File discovery (CSS images, thumbnails, article media) now pushes files inline. They are deduplicated and queued during discovery, instead of batching everything at the end of the dump.

* Removes `Downloader.cssDependenceUrls` cache. Redis is now the sole source of truth for CSS dependencies.
* Flushes per-host Redis queues immediately upon creation to ensure clean per-host queue state for the dump.
* Adds `addManyFilesToProcess` primitive for batched article media pushes.

### DIFF
--- a/src/util/FileManager.ts
+++ b/src/util/FileManager.ts
@@ -1,0 +1,281 @@
+import * as logger from '../Logger.js'
+import Downloader from '../Downloader.js'
+import RedisStore from '../RedisStore.js'
+import { Creator, StringItem } from '@openzim/libzim'
+import pmap from 'p-map'
+import { Dump } from '../Dump.js'
+import { parseRetryAfterHeader } from './misc.js'
+import { FILES_DOWNLOAD_FAILURE_MINIMUM_FOR_CHECK, FILES_DOWNLOAD_FAILURE_TRESHOLD_PER_TEN_THOUSAND, MAX_FILE_DOWNLOAD_RETRIES } from './const.js'
+import urlHelper from './url.helper.js'
+import { fileDownloadMutex, zimCreatorMutex } from '../mutex.js'
+import RedisQueue from './RedisQueue.js'
+
+const MAXIMUM_FILE_DOWNLOAD_DELAY = 20000
+
+interface HostData {
+  queue: RedisQueue<FileToDownload>
+  lastRequestDate?: number
+  requestInterval: number
+  notBeforeDate?: number
+  downloadSuccess: number
+  downloadFailure: number
+  downloadsComplete: boolean
+}
+
+class FileManager {
+  private static instance: FileManager
+  public static getInstance() {
+    if (!FileManager.instance) {
+      FileManager.instance = new FileManager()
+    }
+    return FileManager.instance
+  }
+
+  private hosts: Map<string, HostData> = new Map()
+
+  private get filesToDownloadXPath(): RKVS<FileDetail> {
+    return RedisStore.filesToDownloadXPath
+  }
+
+  private constructor() {}
+
+  /**
+   * Reset internal state for a new dump cycle.
+   */
+  public reset(): void {
+    this.hosts = new Map()
+  }
+
+  /**
+   * Add a single file to download. Checks dedup via filesToDownloadXPath.
+   * On first add: writes to both filesToDownloadXPath and host queue.
+   * On resolution upgrade: updates filesToDownloadXPath only (queue already has the path).
+   *
+   * This method is typically called multiple times when the file is an image and the scraper detects
+   * a second use of an image already seen but with a higher resolution. Could be that we call this
+   * method twice with different urls / data for other usage in the future.
+   */
+  public async addFileToProcess(path: string, detail: FileDetail): Promise<void> {
+    const existing = await this.filesToDownloadXPath.get(path)
+    if (existing) {
+      const isHigherRes = existing.width < (detail.width || 10e6) || existing.mult < (detail.mult || 1)
+      if (!isHigherRes) return
+      // Resolution upgrade: update store only, queue already has this path
+      await this.filesToDownloadXPath.set(path, detail)
+      return
+    }
+    // New file: add to both store and host queue
+    await this.filesToDownloadXPath.set(path, detail)
+    await this.pushToHostQueue(path, detail)
+  }
+
+  /**
+   * Batch version of addFileToProcess. Uses getMany for efficiency.
+   */
+  public async addManyFilesToProcess(files: KVS<FileDetail>): Promise<void> {
+    const paths = Object.keys(files)
+    if (!paths.length) return
+
+    const existingVals = await this.filesToDownloadXPath.getMany(paths)
+    const toSet: KVS<FileDetail> = {}
+
+    for (const [path, detail] of Object.entries(files)) {
+      const existing = existingVals[path]
+      if (existing) {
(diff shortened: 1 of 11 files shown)


src/util/FileDownloadManager.ts
```diff
@@ -0,0 +1,258 @@
...
+  queue: RedisQueue<FileToDownload>
+  lastRequestDate?: number
+  requestInterval: number
+  notBeforeDate?: number
+  downloadSuccess: number
+  downloadFailure: number
+  downloadsComplete: boolean
+}
+
+export default class FileDownloadManager {
```
benoit74: This should have a singleton pattern with `getInstance()` and you export this as default (see Downloader class for instance).

I propose to rename it `FileManager` for conciseness, unless it collides with something else I've oversight

src/util/FileDownloadManager.ts
```diff
@@ -0,0 +1,258 @@
...
+  private filesToDownloadXPath: RKVS<FileDetail>
+  private hosts: Map<string, HostData> = new Map()
+
+  constructor(filesToDownloadXPath: RKVS<FileDetail>) {
```
benoit74: Once it is a singleton and moved out of `RedisStore`, I don't think we need to pass the `filesToDownloadXPath`, we can directly fethc it from `RedisStore.filesToDownloadXPath`

src/util/saveArticles.ts
benoit74: I would recommend to rename this something like`articleFiles` so that it is way more obvious it is a "local" KVS, not the Redis one. It took me ages to understand that.

src/util/FileDownloadManager.ts
```diff
@@ -0,0 +1,258 @@
...
+
+    for (const [path, detail] of Object.entries(files)) {
+      const existing = existingVals[path]
+      if (existing) {
+        const isHigherRes = existing.width < (detail.width || 10e6) || existing.mult < (detail.mult || 1)
+        if (!isHigherRes) continue
+        toSet[path] = detail // upgrade only, no queue push
+      } else {
+        toSet[path] = detail
+        newPaths.push({ path, detail }) // new file, needs queue push
```
benoit74: why not doing the `this.pushToHostQueue` directly here? Would save the `newPaths` variable + the additional for loop later

src/util/FileManager.ts
```diff
@@ -0,0 +1,258 @@
...
+          (dump.status.files.fail * 10000) / filesTotal > FILES_DOWNLOAD_FAILURE_TRESHOLD_PER_TEN_THOUSAND
+        ) {
+          throw new Error(`Too many files failed to download: [${dump.status.files.fail}/${filesTotal}]`)
+        }
+      }
+
+      fileToDownload.downloadAttempts += 1
+
+      // Read latest details from filesToDownloadXPath (resolution may have been upgraded since queuing)
+      const latestDetail = await filesToDownloadXPath.get(fileToDownload.path)
```
benoit74, line 197: Since we need to always fetch latest details from the RKVS, why do we store details in the host queues? It was convenient with previous implementation, but I feel like we could maybe simple store the path and download attempts in the host queues, and get what needs to be stored at this path from the RKVS. Would save duplicating info and "weird" latest logic

benoit74:
Nice, please ping me once it is ready for review

triemerge:
> Nice, please ping me once it is ready for review

@benoit74 ready for review.

benoit74 (changes requested):

src/util/FileDownloadManager.ts
triemerge: Done, renamed to `FileManager` and added `getInstance()` matching the pattern in Downloader.

src/util/FileDownloadManager.ts
triemerge: Done, removed the constructor parameter, `filesToDownloadXPath` is now reading directly from `RedisStore.filesToDownloadXPath`.

src/util/saveArticles.ts
triemerge: Done, renamed to `articleFiles`.

src/util/FileDownloadManager.ts
triemerge: Cleaned up, `pushToHostQueue` is now called inline in the `else` branch, `newPaths` variable and second loop are both gone.

benoit74:
Let me know when it is ready for a new review round

src/util/FileManager.ts
triemerge: Thanks for raising this, `FileToDownload` is now just `{ path, downloadAttempts }`. All metadata lives exclusively in `filesToDownloadXPath`. Workers fetch latest details from the RKVS before downloading, so the "latest" fallback logic is gone too.

triemerge:
> Let me know when it is ready for a new review round

@benoit74 ready for review

src/util/dump.ts
```diff
@@ -64,9 +63,8 @@ export async function processStylesheetContent(cssUrl: string, linkMedia: string
         rewrittenCss = rewrittenCss.replace(url, relativePath.replace(/'/g, '%27').replace(/\(/g, '%28').replace(/\)/g, '%29'))
 
         /* Download CSS dependency, but avoid duplicate calls */
```
benoit74: I don't think we avoid duplicate calls anymore, but this is fine. We can probably simply get rid of this comment IMHO, doesn't add much value but confusion.

I also don't think the `if (filepath)` is really needed anymore (why would this be "False")?

src/util/FileManager.ts
```diff
@@ -0,0 +1,277 @@
...
+   * Reset internal state for a new dump cycle.
+   */
+  public reset(): void {
+    this.hosts = new Map()
+  }
+
+  /**
+   * Add a single file to download. Checks dedup via filesToDownloadXPath.
+   * On first add: writes to both filesToDownloadXPath and host queue.
+   * On resolution upgrade: updates filesToDownloadXPath only (queue already has the path).
```
benoit74, line 52: While I do get "now" what "resolution upgrade" means, I would prefer to clearly mention that this method is typically called multiple times when the file is an image and scraper detected a second use of an image already seen but with a higher resolution. Could be that we call this method twice with different urls / data for other usage in the future

benoit74 (approved):
Nice, two minor comments to consider and I think we are good to merge. I like this change, it is a important simplification of the codebase.

benoit74 (changes requested):
Sorry missclick we need these last changes before merging 😉

src/util/dump.ts
triemerge: Done, removed the stale comment in `dump.ts` and the `if (filepath)` guard since it is not really needed anymore.

src/util/FileManager.ts
triemerge: Updated the comment to explicitly mention the image resolution upgrade scenario, and noted that other use cases with different urls / data could arise in the future

triemerge:
@benoit74 Just pinging you on this since the PR has been open for a while. Please let me know if anything else is needed.

benoit74 (approved):
LGTM

Issue #2072, opened 2024-07-27
mwoffliner should snapshot flaggedRevised "stable" version of articles


kelson42:
If possible, otherwise take the most recent revision.

From https://sourceforge.net/p/kiwix/feature-requests/909/

KunalSiyag:
I would like to work on this issue. Since this is closely related to #1587 (adding the option to scrape FlaggedRevs), I plan to implement both in a single PR to ensure the configuration and the logic are aligned. 
My workflow will be to 
- update the existing MediaWiki Action API calls to include prop=flagged. 
- prioritize stable_revid if available; otherwise, fall back to lastrevid to ensure no articles are skipped. 
- Update the downloader to request the specific revision via the REST API endpoint (/revision/{revid}/html) when a stable ID is identified.

That's the plan for now. Tried a POC on my browser only, will see how that holds out in my system when integrated.

<img width="439" height="237" alt="Image" src="https://github.com/user-attachments/assets/13cbc75b-6580-461a-a4de-fa33f30ebf20" />

<img width="964" height="639" alt="Image" src="https://github.com/user-attachments/assets/5965598d-3627-4a98-8203-02ed431a7118" />

benoit74:
#2698 added a CLI option to make this behavior available optionally. I don't think turning this on systematically is a good idea at this stage given we have no clue how FlaggedRevs is really used when enabled (could be enabled but abandoned, causing only too old revs to be retrieved).

I propose to not move this issue forward at this stage (keep behavior optional)

PR #2698, opened 2026-04-07
Add CLI option to optionally prefer stable revisions (only FlaggedRevs-enabled wikis supported)

### LINKED ISSUE #1587: Option to scrape FlaggedRevs revision of each article (if specified)
This would allow to have a better garanty to avoid vandalism.

See https://phabricator.wikimedia.org/T19189

KunalSiyag:
This PR adds automatic detection of the FlaggedRevs extension and, when available, prefers stable (reviewed) revisions instead of the latest ones. The goal is to reduce vandalism and improve the quality of generated ZIM files.

Closes #1587, relates to #2072 

## Changes
- **`MediaWiki.ts`**: Added `hasFlaggedRevs()` method that probes `prop=info|flagged` to detect the extension (cached, similar to `hasCoordinates()`)
- **`Downloader.ts`**: Appends `|flagged` to article query `prop` when FlaggedRevs is active
- **`mw-api.ts`**: Extracts `val.flagged.stable_revid` → `articleDetail.stableRevisionId` in `mwRetToArticleDetail()`
- **`types.d.ts`**: Added `flagged` , `stableRevisionId`, `oldid`
- **`action-parse.director.ts`**: 
   - `buildArticleURL()` uses `oldid=<stable_revid>` instead of `page=<title>` when `oldid` is provided
  - Handle missing `redirects` field in API response when using `oldid=` (not returned for specific revision requests)
- **`saveArticles.ts`**: Both `saveArticles()` and `getAllArticlesToKeep()` pass `stableRevisionId` as `oldid` when `useLatestRevision` is not set
- **`parameterList.ts`**: Added `--useLatestRevision` CLI flag description
- **`Dump.ts`**: Added `useLatestRevision` to dump opts
- **`mwoffliner.lib.ts`**: Wires detection at boot, passes flag through, adds `flaggedrevs:stable` ZIM tag

### Unit Tests (new)
- `mwRetToArticleDetail`: extracts `stableRevisionId`, handles absence, handles mixed data
- `buildArticleURL`: uses `oldid` instead of `page`, preserves existing behavior, combines `oldid` + `sectionId`

### Manual Docker Test
- ✅ `de.wikipedia.org` (FlaggedRevs) -> stable revision used, `oldid=` in parse URL
- ✅ `de.wikipedia.org` + `--useLatestRevision` -> latest revision used, `page=` in parse URL
- ✅ `en.wikipedia.org` (no FlaggedRevs) -> graceful fallback, no behavior change

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -127,6 +127,7 @@ class MediaWiki {
   #hasActionParseApi: boolean | null
   #hasCoordinates: boolean | null
   #hasModuleApi: boolean | null
+  #hasFlaggedRevs: boolean | null
 
   set username(value: string) {
     this.#username = value
@@ -221,6 +222,7 @@ class MediaWiki {
     this.#hasActionParseApi = null
     this.#hasCoordinates = null
     this.#hasModuleApi = null
+    this.#hasFlaggedRevs = null
     this.metaData = null
   }
 
@@ -267,6 +269,25 @@ class MediaWiki {
       logger.log('Checked for Module API at', checkUrl, '-- result is: ', this.#hasModuleApi)
     }
     return this.#hasModuleApi
+  }
+
+  public async hasFlaggedRevs(): Promise<boolean> {
+    if (this.#hasFlaggedRevs === null) {
+      const reqOpts = {
+        ...this.queryOpts,
+        prop: 'info|flagged',
+        titles: this.apiCheckArticleId,
+      }
+      const resp = await Downloader.getJSON<MwApiResponse>(this.#apiUrlDirector.buildQueryURL(reqOpts))
+      const isFlaggedWarning = JSON.stringify(resp?.warnings?.query ?? '').includes('flagged')
+      if (isFlaggedWarning) {
+        logger.log('FlaggedRevs not available on this wiki')
+        return (this.#hasFlaggedRevs = false)
+      }
+      logger.log('FlaggedRevs available on this wiki')
+      return (this.#hasFlaggedRevs = true)
+    }
+    return this.#hasFlaggedRevs
   }
 
   private setModuleURL() {
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -103,6 +103,7 @@ async function execute(argv: any) {
     langVariant,
     customCss,
     userAgent: _userAgent,
+    stableRevision,
   } = argv
 
   let { articleList, articleListToIgnore } = argv
@@ -258,6 +259,10 @@ async function execute(argv: any) {
   await MediaWiki.hasActionParseApi()
   await MediaWiki.hasModuleApi()
 
+  if (stableRevision && !(await MediaWiki.hasFlaggedRevs())) {
+    throw new Error('--stableRevision was specified but this wiki does not support stable revisions (FlaggedRevs extension not found)')
+  }
+
   await RenderingContext.createRenderers(forceRender)
 
   await RedisStore.connect()
@@ -378,6 +383,7 @@ async function execute(argv: any) {
           keepEmptySections,
           tags: customZimTags,
           filenameDate,
+          stableRevision,
         },
         { ...mwMetaData, mainPage },
         customProcessor,
@@ -393,7 +399,6 @@ async function execute(argv: any) {
       } catch {
         shouldSkip = true
       }
-
       if (shouldSkip) {
         logger.log('Skipping dump')
       } else {
--- a/src/parameterList.ts
+++ b/src/parameterList.ts
@@ -45,6 +45,7 @@ export const parameterDescriptions = {
   insecure: 'Skip HTTPS server authenticity verification step',
   customCss: 'Comma-separated list of CSS URLs to inject into all rendered pages',
   userAgent: 'Custom User-Agent header for all HTTP requests. Defaults to "MWOffliner/<version> (<adminEmail>)"',
+  stableRevision: 'Prefer stable articles revisions when available, based on FlaggedRevs extension (which needs to be active on the wiki).',
 }
 
 // TODO: Add an interface based on the object above
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -33,6 +33,7 @@ type ArticleDetail = PageInfo & {
   coordinates?: string // coordinates.0.lat;coordinates.0.lon
   timestamp?: string // revisions.0.timestamp
   revisionId?: number // revisions.0.revid
+  stableRevisionId?: number // flagged.stable_revid (FlaggedRevs)
   internalThumbnailUrl?: string // internalThumbnailUrl
   nextArticleId?: string
   prevArticleId?: string
@@ -120,6 +121,9 @@ interface QueryRet {
   revisions?: QueryRevisionsRet
   coordinates?: QueryCoordinatesRet
   redirects?: QueryRedirectsRet
+  flagged?: {
+    stable_revid?: number
+  }
   pagelanguagehtmlcode?: string
   pagelanguagedir?: TextDirection
   contentmodel: string
@@ -237,6 +241,7 @@ interface QueryContinueOpts {
 interface RendererArticleOpts {
   sectionId?: string
   langVar?: string
+  oldid?: number
 }
 
 interface PageInfo {
--- a/src/util/saveArticles.ts
+++ b/src/util/saveArticles.ts
@@ -16,14 +16,19 @@ import FileManager from './FileManager.js'
 import { truncateUtf8Bytes } from './misc.js'
 import { isMainPage } from './articles.js'
 
+function getArticleRenderUrl(articleId: string, articleDetail: ArticleDetail, dump: Dump): string {
+  const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
+  const oldid = dump.opts.stableRevision && articleDetail.stableRevisionId !== undefined ? articleDetail.stableRevisionId : undefined
+  return Downloader.getArticleUrl(articleId, { sectionId: leadSectionId, oldid, langVar: dump.langVar })
+}
+
 async function getAllArticlesToKeep(articleDetailXId: RKVS<ArticleDetail>, dump: Dump, articlesRenderer: Renderer) {
   await articleDetailXId.iterateItems(Downloader.workers, async (articleKeyValuePairs) => {
     for (const [articleId, articleDetail] of Object.entries(articleKeyValuePairs)) {
       let rets: any
       try {
         const mainPage = isMainPage(articleId)
-        const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
-        const articleUrl = Downloader.getArticleUrl(articleId, { sectionId: leadSectionId, langVar: dump.langVar })
+        const articleUrl = getArticleRenderUrl(articleId, articleDetail, dump)
 
         rets = await Downloader.getArticle(articleId, articleDetailXId, articlesRenderer, articleUrl, dump, articleDetail)
         await new Promise((resolve) => setTimeout(resolve, Downloader.articleRequestInterval))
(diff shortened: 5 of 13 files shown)


KunalSiyag:
I’m still a bit conflicted on whether we should default to stable revisions (for higher-quality ZIMs) or the latest revisions (for the most up-to-date content).

For now, I’ve leaned towards prioritizing quality by default and added a `--useLatestRevision` flag for cases where freshness is more important.

VikramAditya33:
Hey @KunalSiyag I just feel like that the code you've written from line 246 to 251 and 425 to 430 in `src/util/saveArticles.ts`, these two code blocks are identical duplicates, they compute the same `preferStable` and `renderOldid` values, log the same message if applicable, and generate the same `articleUrl` using `Downloader.getArticleUrl`. I feel like you can refactor these into one and extract the logic into a reusable function to eliminate duplication while preserving the original behavior.

KunalSiyag:
> Hey @KunalSiyag I just feel like that the code you've written from line 246 to 251 and 425 to 430 in `src/util/saveArticles.ts`, these two code blocks are identical duplicates, they compute the same `preferStable` and `renderOldid` values, log the same message if applicable, and generate the same `articleUrl` using `Downloader.getArticleUrl`. I feel like you can refactor these into one and extract the logic into a reusable function to eliminate duplication while preserving the original behavior.

Good catch. I will refactor it into a single function and call upon it.

kelson42:
> I’m still a bit conflicted on whether we should default to stable revisions (for higher-quality ZIMs) or the latest revisions (for the most up-to-date content).

This new feature requires a dedicated cmd line argument `--flagged-revs`

KunalSiyag:
> > I’m still a bit conflicted on whether we should default to stable revisions (for higher-quality ZIMs) or the latest revisions (for the most up-to-date content).
> 
> This new feature requires a dedicated cmd line argument `--flagged-revs`

Currently we have a ` --useLatestRevision` but

Should we do something like:

--revision-mode=stable (Current)
--revision-mode=latest
--revision-mode=hybrid (Fetch the stable, if too old -> fetch the latest and if no stable version found then fetch latest and many more rules)


Commenting from mobile (formatting might be misplaced 😓)

KunalSiyag:
Refactored the duplication into a single function, Moved `flaggedrevs:stable` tag into Dump for dedup safety.

Added a more robust logging instead of just printing "Stable article found".
- If stable ≠ latest → we see at default verbosity: Using stable revision [X] instead of latest [Y] for article [Berlin]
- If stable = latest → visible with `--verbose=info` Stable revision matches latest [X] for article [Berlin]
- With `--useLatestRevision` → no log (stable is bypassed entirely)


image

src/MediaWiki.ts
```diff
@@ -269,6 +271,27 @@ class MediaWiki {
     return this.#hasModuleApi
   }
 
+  public async hasFlaggedRevs(): Promise<boolean> {
```
Markus-Rost, line 276: A wiki might have FlaggedRevs installed, but disabled (like en.wikipedia.org). This function would return true in that case, however the resulting ZIM would be no different from a normal one except for the addition of the `flaggedrevs:stable` tag. Do we always want to add the tag or only if we actually find an article with a non-stable revision?

KunalSiyag: Good point. It is new knowledge to me; I wasn’t aware that some wikis can have FlaggedRevs installed but not actively use it.

In that case, I think we should only add the `flaggedrevs:stable` tag if we actually end up using a stable (reviewed) revision for at least one article. 

Or a deeper way would be to include the percentage of stable articles used and upon crossing a certain threshold, include the tag.

What do you think?

kelson42:
> > > I’m still a bit conflicted on whether we should default to stable revisions (for higher-quality ZIMs) or the latest revisions (for the most up-to-date content).
> > 
> > 
> > This new feature requires a dedicated cmd line argument `--flagged-revs`
> 
> Currently we have a ` --useLatestRevision` but
> 
> Should we do something like:
> 
> --revision-mode=stable (Current) --revision-mode=latest --revision-mode=hybrid (Fetch the stable, if too old -> fetch the latest and if no stable version found then fetch latest and many more rules)
> 
> Commenting from mobile (formatting might be misplaced 😓)

I'm not in favour of adding `--useLatestRevision` or having any complex revision logic like you have done. This is IMHO just uselessly complicated. If nothing is specified, then it will continue to work like i has done for the last 15 years. Please just make a `--flaggedRevs` which relies on the Flagged revisions (if they exists). Please ensure at start that MWoffliner complains if the user specifies `--flaggedRevs` but the MediaWiki instance does not have the appropriate extension.

KunalSiyag:
I will revise the modifications, the core working remains same, but with an addition of the `--flaggedrev` flag that users can select to fetch the stable version of the articles when available. Will also look into the e2e tests that are failing because of this.

benoit74:
I'm really not sure `--flaggedRevs` is the best parameter name because it does not indicate we will use the revision flagged `stable` whenever possible. 

Seeing the disclaimer at https://www.mediawiki.org/wiki/Extension:FlaggedRevs, I also doubt that using "flagged revs" in parameter name is adequate.

image

I would propose to use a more generic parameter like `--stableRevision` which can be adapted to support both `FlaggedRevs` and `Approved Revs` extensions and any other one deemed of interest.

@kelson42 WDYT?

kelson42:
> I would propose to use a more generic parameter like --stableRevision which can be adapted to support both FlaggedRevs and Approved Revs extensions and any other one deemed of interest.

@benoit74 yes.

KunalSiyag:
The old logic remains the same, only the new `--stableRevision` parameter is added as per the discussion. 

Also even though it is a generic name for both the FlaggedRev as well as the ApprovedRev, I believe implementing the latter should require a separate PR, since that one requires different approach (did some search with AI as well as the MediaWiki Doc; it works with a single `<page id>` at a time, 10,000 API calls for 10k articles, reading the docs to see how much latency can it induce even when we have with concurrent requests).

src/util/saveArticles.ts
```diff
@@ -236,14 +236,30 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
   )
 }
 
+/**
+ * Build the article URL, using the stable (FlaggedRevs) revision when --stableRevision
+ * is set and a stable revision is available for the article.
+ */
+export function getArticleUrlForRevision(articleId: string, articleDetail: ArticleDetail, dump: Dump): string {
```
benoit74, line 243: Function name and comment are misleading. 

This function does way more than adapting to the fact that `--stableRevision` is set or not. It also adapts to `nodet`. Please choose a better name so that we are not misled + we do not need the comment because it is mostly self-explanatory.

src/util/saveArticles.ts
```diff
@@ -236,14 +236,30 @@ export async function downloadFiles(fileStore: RKVS<FileDetail>, zimCreator: Cre
...
+  const leadSectionId = dump.nodet && !articleDetail.contentmodel && !articleDetail.missing ? config.filters.leadSectionId : ''
+  const shouldUseStable = dump.opts.stableRevision && articleDetail.stableRevisionId !== undefined
```
benoit74, line 245: I think all the following lines are overly complicated for little added value. 

First I'm not sure at all we want to add a dedicated ZIM Tag. This has been discussed nowhere and will not be used anywhere ATM, so please do not add this (or point to the discussion I've missed). 

And I don't think logs are useful (too many logs produced, we cannot log every decision the scraper makes).

src/MediaWiki.ts
```diff
@@ -221,6 +223,8 @@ class MediaWiki {
     this.#hasActionParseApi = null
     this.#hasCoordinates = null
     this.#hasModuleApi = null
+    this.#hasFlaggedRevs = null
```
benoit74, line 226: Why would we need two properties? I don't get which one is used when.

src/MediaWiki.ts
benoit74: As pointed in my other comment, I don't think we should add a tag at all

src/MediaWiki.ts
```diff
@@ -269,6 +273,27 @@ class MediaWiki {
...
+    if (this.#hasFlaggedRevs === null) {
+      const reqOpts = {
+        ...this.queryOpts,
+        prop: 'info|flagged',
+        titles: this.apiCheckArticleId,
+      }
+      const resp = await Downloader.getJSON<MwApiResponse>(this.#apiUrlDirector.buildQueryURL(reqOpts))
+      const isFlaggedWarning = JSON.stringify(resp?.warnings?.query ?? '').includes('flagged')
+      const hasError = !!resp?.error
+      if (isFlaggedWarning || hasError) {
```
benoit74, line 286: Please use same log "format" + code "structure" than other checks, e.g. we do not need an if/else block

src/mwoffliner.lib.ts
```diff
@@ -255,6 +256,15 @@ async function execute(argv: any) {
   await MediaWiki.hasActionParseApi()
   await MediaWiki.hasModuleApi()
 
+  // Only probe for FlaggedRevs when --stableRevision is set
+  if (stableRevision) {
+    const hasFlagged = await MediaWiki.hasFlaggedRevs()
```
benoit74, line 261: I don't think we need to store result in a variable since it is already stored in `Mediawiki.#hasFlaggedRevs`

src/parameterList.ts
```diff
@@ -45,6 +45,7 @@ export const parameterDescriptions = {
   insecure: 'Skip HTTPS server authenticity verification step',
   customCss: 'Comma-separated list of CSS URLs to inject into all rendered pages',
   userAgent: 'Custom User-Agent header for all HTTP requests. Defaults to "MWOffliner/<version> (<adminEmail>)"',
+  stableRevision: 'Use stable (reviewed) revisions when available via FlaggedRevs. Errors if the target wiki does not support stable revisions.',
```
benoit74, line 48: This is confusing as well. Something along the lines of `Prefer stable articles revisions when available, based on FlaggedRevs extension (which needs to be active on the wiki).` (feel free to enhance further)

src/types.d.ts
```diff
@@ -120,6 +121,11 @@ interface QueryRet {
   revisions?: QueryRevisionsRet
   coordinates?: QueryCoordinatesRet
   redirects?: QueryRedirectsRet
+  flagged?: {
+    stable_revid?: number
+    level?: number
```
benoit74, line 126: Do we need to capture `level` and `level_text` since we don't use them AFAIK?

benoit74 (changes requested):

src/util/saveArticles.ts
KunalSiyag: I will simply them, the tag was an addition from my side, since I assumed that snapshot would mean some kind of tag to represent those articles and logs were for me to verify whether everything was working. I will make it clean.

src/MediaWiki.ts
KunalSiyag: You are right, Only 1 should remain. (Will keep `#hasFlaggedRevs` cached via `hasFlaggedRevs()`)

src/MediaWiki.ts
KunalSiyag: Will mirror it like other logs.

src/mwoffliner.lib.ts
KunalSiyag: Will include the call directly in the check.

src/parameterList.ts
KunalSiyag: Will change it to -> "Prefer stable articles revisions when available, based on FlaggedRevs extension (which needs to be active on the wiki)". This should work ig.

src/types.d.ts
KunalSiyag: Only the `stable_revid` is used. This was an error on my part, didn't check enough. Will remove them.

KunalSiyag:
I have made the changes, will run the tests and on successful compilation, push the branch. Thank you for the suggestions. I missed some important aspects, which I wouldn't have found otherwise.

KunalSiyag:
Update:
Running into local errors, likely due to a version/ABI mismatch after adding the Kiwix PPA.

`symbol lookup error: zimdump: undefined symbol: _ZN3zim5EntryC1ESt10shared_ptrINS_8FileImplEEj`

Seems like zim-tools and libzim are out of sync on my system.
I’ll fix the environment and rerun the tests before pushing the changes.

benoit74:
Rebased, hopefully will merge this soon

benoit74 (approved):
LGTM

Issue #1891, opened 2023-08-25
dump.isMainPage() method does not return true when retrieve single main page from CLI


VadimKovalenkoSNF:
mwoffliner doens't trigger events related to main page while trying to retrieve it directly

## Steps to reproduce
Run:
`npm start -- --mwUrl=https://en.wikivoyage.org --adminEmail=mail@mail.com --articleList=Main_Page`

## Expected result:
`dump.isMainPage(articleId)` to return `true`

## Actual result:
`dump.isMainPage(articleId)` returns `false`

## Solution
While deeper investigation I found that `dump.mwMetaData.mainPage` is set to an empty string on the stage of sanitizing the main page while the execution process is in `src/mwoffliner.lib.ts`. The sanitizing process should probably be refactored to handle a single main page in the article list. Also, the issue mentioned in https://github.com/openzim/mwoffliner/issues/1872 is probably a symptom of the wrong main page calculation.

VadimKovalenkoSNF:
Upd: this problem might be mitigated by explicit setting of the main page title in the CLI:
`npm start -- --mwUrl=https://en.wikivoyage.org --adminEmail=mail@mail.com --articleList=Main_Page --customMainPage=Main_Page`

VadimKovalenkoSNF:
>Also, the issue mentioned in https://github.com/openzim/mwoffliner/issues/1872 is probably a symptom of the wrong main page calculation.

Upd 2: This is not related to a problem with CLI options but rather because `injectHeader` is not included in the Wikimedia Desktop API renderer.

KunalSiyag:
I would like to look into this.

benoit74:
@KunalSiyag please first double check this issue is still valid in `main` branch

benoit74:
And especially what the consequences are.

KunalSiyag:
I ran `npm start -- --mwUrl=https://en.wikivoyage.org --adminEmail=mail@mail.com --articleList=Main_Page` and saw that the zim succeeds but the generated ZIM main page is index, not Main_Page and with ` --customMainPage=Main_Page`, the main page is Main_Page. 
I also ran an instrumentation script and:
- `dump.mwMetaData.mainPage was ""`
- `MediaWiki.metaData.mainPage was "Main_Page"`
- `isMainPage('Main_Page') returned true`

[...]

PR #2703, opened 2026-04-10
Create multiple files for language variants

### LINKED ISSUE #2691: Create multiple files for language variants, similar to different formats
The language variant argument should accept a list, similar to the format, and create a different ZIM file for each of them one after the other. 

Wikis with language variants, like Chinese, often want files in multiple variants. Treating them similar to the format currently would make creating them a lot easier.

Additionally this would ensure we don't overwhelm a wiki by doing scrapes for multiple variants at the same time (like in https://github.com/openzim/zim-requests/issues/755#issuecomment-4179612839, re @MESLewis).

Markus-Rost:
Fix #2691

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -261,9 +261,9 @@
       "required": false
     },
     "langVariant": {
-      "type": "string",
-      "title": "Language Variant",
-      "description": "Use a specific language variant, only for wikis supporting language conversion",
+      "type": "list-of-string",
+      "title": "Language Variants",
+      "description": "Use specific language variants, only for wikis supporting language conversion. Each variant will cause a new ZIM file to be created.",
       "required": false
     },
     "javaScript": {
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -247,9 +247,9 @@
       "required": false
     },
     "langVariant": {
-      "type": "string",
-      "title": "Language Variant",
-      "description": "Use a specific language variant, only for wikis supporting language conversion",
+      "type": "list-of-string",
+      "title": "Language Variants",
+      "description": "Use specific language variants, only for wikis supporting language conversion. Each variant will cause a new ZIM file to be created.",
       "required": false
     },
     "javaScript": {
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -65,10 +65,6 @@ const __dirname = path.dirname(__filename)
 const packageJSON = JSON.parse(readFileSync(path.join(__dirname, '../package.json'), 'utf8'))
 
 async function execute(argv: any) {
-  /* ********************************* */
-  /* CUSTOM VARIABLE SECTION ********* */
-  /* ********************************* */
-
   const {
     speed: _speed,
     adminEmail,
@@ -110,6 +106,17 @@ async function execute(argv: any) {
   } = argv
 
   let { articleList, articleListToIgnore } = argv
+
+  /* *********************************** */
+  /*       SYSTEM VARIABLE SECTION       */
+  /* *********************************** */
+
+  const dumpFormats = getDumps(format)
+  const langVariants: string[] = getDumps(langVariant)
+
+  /* ********************************* */
+  /* CUSTOM VARIABLE SECTION ********* */
+  /* ********************************* */
 
   logger.setVerboseLevel(verbose ? verbose : 'log') // Default log level is 'log'
 
@@ -222,7 +229,7 @@ async function execute(argv: any) {
   /* Get MediaWiki Info */
   let mwMetaData
   try {
-    mwMetaData = await MediaWiki.getMwMetaData({ addNamespaces, mwModulePath, forceSkin, langVariant })
+    mwMetaData = await MediaWiki.getMwMetaData({ addNamespaces, mwModulePath, forceSkin, langVariants })
   } catch (err) {
     logger.error('FATAL - Failed to get MediaWiki Metadata')
     throw err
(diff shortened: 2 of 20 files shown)


triemerge:
Shouldn't we update the `langVariant` description in `parameterList.ts` to reflect the new list support? Also, the `#1891` changelog entry appears to be duplicated

Markus-Rost:
> Shouldn't we update the `langVariant` description in `parameterList.ts` to reflect the new list support? Also, the `#1891` changelog entry appears to be duplicated

You are correct, updated/fixed.

src/mwoffliner.lib.ts
```diff
@@ -356,49 +357,52 @@ async function execute(argv: any) {
...
+          minifyHtml,
+          keepEmptySections,
+          tags: customZimTags,
+          filenameDate,
+        },
+        { ...mwMetaData, mainPage },
+        customProcessor,
+      )
+      dumps.push(dump)
+      logger.log('Doing dump')
```
benoit74: I know this was here before this PR, but I now realize it would be way better to add lang variant + dump format to this log message, and maybe some "emphasis" (stars, ...) since this is a very important log separating huge portions of logs

src/parameterList.ts
```diff
@@ -41,7 +41,7 @@ export const parameterDescriptions = {
   optimisationCacheUrl: 'Object Storage URL (including credentials and bucket name) to cache optimised media files',
   forceRender: 'Force the usage of a specific API end-point/render, automatically chosen otherwise. Accepted values: [ ActionParse ]',
   forceSkin: 'Force the usage of a specific skin, automatically chosen otherwise.',
-  langVariant: 'Use a specific language variant, only for wikis supporting language conversion.',
+  langVariant: 'Use a specific language variant, only for wikis supporting language conversion. Each --langVariant argument will cause a new local file to be created.',
```
benoit74: I feel like `a new local file` is a bit confusing, why not simply saying `a new ZIM`?

offliner-definition.json
```diff
@@ -261,9 +255,9 @@
       "required": false
     },
     "langVariant": {
-      "type": "string",
-      "title": "Language Variant",
-      "description": "Use a specific language variant, only for wikis supporting language conversion",
+      "type": "list-of-string",
+      "title": "Language Variants",
+      "description": "Use specific language variants, only for wikis supporting language conversion. Each variant will cause a new local file to be created.",
```
benoit74: same remark

benoit74 (changes requested):
Thank you, looks pretty ok to me.

Only problem is that it looks like the `computeFilenameRadical` logic does not match our [openZIM convention](https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention#ZIM_Name_Metadata) where name should be composed of only 3 parts, while here it looks like it would give 4.

Our solution to solve this is to replace the `lang` part by the `variant`. This is supposed to be sufficient to not have colliding ZIM names, unless I miss something.

Shall we add one more test to `test/unit/dump.test.ts` to check that behavior when a variant is passed?

What should happen when `filenamePrefix` is passed is a bit concerning. If we keep this "as-is", then we probably have collisions when multiple variants are requested. If we append the variant to the prefix, then we probably break the openZIM convention. What I would propose is to keep the `filenamePrefix` as-is for this PR, fail the scraper if filenamePrefix and multiple variants are requested, and open a follow-up issue to revisit this `filenamePrefix` argument to follow other scrapers behavior where one can rather specify ZIM name with placeholders which are then replaced by the scraper. E.g. default ZIM name should be `{domain}_{lang_or_variant}_{selection}`. WDYT?

src/parameterList.ts
Markus-Rost: I just copied the test from the `format` parameter, I'll update both.

Markus-Rost:
> Thank you, looks pretty ok to me.
> 
> Only problem is that it looks like the `computeFilenameRadical` logic does not match our [openZIM convention](https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention#ZIM_Name_Metadata) where name should be composed of only 3 parts, while here it looks like it would give 4.
> 
> Our solution to solve this is to replace the `lang` part by the `variant`. This is supposed to be sufficient to not have colliding ZIM names, unless I miss something.
> 
> Shall we add one more test to `test/unit/dump.test.ts` to check that behavior when a variant is passed?
> 
> What should happen when `filenamePrefix` is passed is a bit concerning. If we keep this "as-is", then we probably have collisions when multiple variants are requested. If we append the variant to the prefix, then we probably break the openZIM convention. What I would propose is to keep the `filenamePrefix` as-is for this PR, fail the scraper if filenamePrefix and multiple variants are requested, and open a follow-up issue to revisit this `filenamePrefix` argument to follow other scrapers behavior where one can rather specify ZIM name with placeholders which are then replaced by the scraper. E.g. default ZIM name should be `{domain}_{lang_or_variant}_{selection}`. WDYT?

Looks like we are already not following this convention when multiple flavor types are provided like `nodet,nopic` (instead of `nodet,nopic:mini`) as they are also joined using `_`.

I agree that we should probably switch to using placeholder in the filenamePrefix/ZIM name. For now I've implemented the suggested error when the custom prefix is used together with multiple variants.

src/mwoffliner.lib.ts
Markus-Rost: Updated

offliner-definition.json
Markus-Rost: Updated

benoit74 (approved):
LGTM, thank you! I will open a dedicated issue about placeholders and problem with flavours

PR #2706, opened 2026-04-14
Fix Request-URI Too Long when fetching batches of article details

### LINKED ISSUE #2705: A new case of Request-URI Too Long while fetching batches of articles details
See https://github.com/openzim/zim-requests/issues/1727

I significantly suspect non-latin characters which have to be escaped in the URL are not "counted" appropriately.

Interestingly, the `all` recipe is working well, so we are probably just "unlucky" with `top` and "unlucky" ordering of articles.

raj921:
## Summary

- **Use `URLSearchParams` for URL size estimation** to match the encoding actually used by `buildQueryURL`, fixing a mismatch where `encodeURIComponent` underestimates the size for characters like `(`, `)`, `!`, `~`, `'` (1 byte vs 3 bytes with `URLSearchParams`)
- **Reduce max titles param size from 7900 to 7400** to account for the ~500-600 bytes of URL overhead (base URL, other query params, continuation params) that was previously ignored
- **Add `articleIdsBatch.length > 1` guard** to prevent popping all articles, and recompute size on each pop instead of buggy incremental subtraction

Fixes #2705

## Test plan

- [ ] Verify scraping a non-Latin wiki (e.g. Telugu Wikipedia with `top` recipe) no longer triggers 414 Request-URI Too Long
- [ ] Verify Latin wiki scraping still works correctly with unchanged batch sizes
- [ ] Verify the `all` recipe continues to work as before

### DIFF
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -11,7 +11,10 @@ export async function getArticlesByIds(articleIds: string[], log = true): Promis
   let from = 0
   let numThumbnails = 0
   const MAX_BATCH_SIZE = 50
-  const MAX_URL_SIZE = 7900 // in bytes, approx.
+  // HTTP servers typically limit URLs to ~8000 bytes. The full URL includes the
+  // base URL, other query params, and continuation params (~500-600 bytes overhead).
+  // This limit applies to the "titles=<encoded>" portion of the query string.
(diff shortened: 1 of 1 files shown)


benoit74:
Closing in favor of #2711

Issue #2707, opened 2026-04-14
was `--customFlavour` option removed?


oneorseveralcats:
I plan on using the english wikipedia mini top 50,000 as a base for creating a zim file that contains only the introduction (no title, initial heading or infobox) that can be converted to stardict (via pyglossary) for an encyclopedia on my e-reader. 

```bash
podman run -v "$(pwd)"/out:/out -ti ghcr.io/openzim/mwoffliner mwoffliner \
  --mwUrl=https://en.wikipedia.org/ \
  --adminEmail=contact@kiwix.org \
  --articleList=http://download.openzim.org/wp1/enwiki/tops/100.tsv \
  --customFlavour customFlavour \
  --customZimTitle='Best of Wikipedia' \
  --customZimDescription='A selection of the best 50,000 Wikipedia articles' \
  --customZimFavicon=https://drive.farm.openzim.org/blobs/9a529a27-47d6-4cae-b7cb-d0575d4af56b.png \
  --publisher=openZIM \
  --filenamePrefix=wikipedia_en_top \
  --format=nodet,nopic:mini \
  --webp \
  --forceRender=ActionParse
```

(testing using the top 100 articles instead of 50,000. also podman can successfully create the zim file without customFlavour specified, so don't think it's the issue)

The above command returns `Unknown argument: customFlavour` and it's not listed in the `--help` so I assume it's been removed?

kelson42:
Not hat I know... or remember... but we never have use it in production.

oneorseveralcats:
Seems like the only thing that used the option was wiktionary_fr and the build for it has been failing for like 4 years: https://farm.openzim.org/recipes/wiktionary_fr_app. the failure 4 years ago was trying to read a null, but the one from ~~2 years ago~~ a year ago was the option not found one i've been getting.

I believe this was where it was removed from the help dialogue: https://github.com/openzim/mwoffliner/pull/1996/changes#diff-4b38193f0ea26c686c757973359e1cd549a6a2ab1fe631da5b188bb26e6c56edL41

it's also still listed as an argument: https://github.com/openzim/mwoffliner/blob/2ae1e96ec81537f737f5f7e41a908b032efd3196/src/mwoffliner.lib.ts#L104

benoit74:
I confirm it has been removed by #1996 (by mistake? I don't recall)

@kelson42 how do you want to fix this?

kelson42:
[...]

PR #2708, opened 2026-04-15
synchronize offliner-definition.json and parameterList.ts

elfkuzco:
## Rationale
This PR attempts to synchronize the meaning of the flags between the `paramterList.ts` and `offliner-definition.json`. See (openzim/zimfarm#1682) and (https://github.com/openzim/zimfarm/issues/1682#issuecomment-4182630716)


elfkuzco:
@benoit74 , I found that some of the descriptions in the `offliner-definition.json` were better for non-technical users compared to the one in the `parameterList.ts`. Moreover, the one in the JSON file doesn't try to describe options that are enums since those are going to be visible to users as drop down options. However, there are a few discrepancies between the two files worth mentioning.

- the `offliner-definition.json` contains flags `getCategories` and `customFlavour` while the `parameterList.ts` doesn't
- the description of the `flienamePrefix` is better (to me) in the JSON file than in the `parameterList.ts`
- With respect to the `speed` flag, there is a conflicting definition between the two files regarding the default value. `offliner-definition.json` says "Otherwise 1", while `src/parameterList.ts` states "(by default the number of CPU cores). The default value is 1." This could be confusing to a user as to whether it scales automatically with CPU cores or strictly defaults to 1.
- `parameterList.ts` contains `resume` and `redis` flags but JSON spec doesn't

benoit74:
> Moreover, the one in the JSON file doesn't try to describe options that are enums since those are going to be visible to users as drop down options. However, there are a few discrepancies between the two files worth mentioning.

These enums particularity is a normal "discrepancy"

> the offliner-definition.json contains flags getCategories and customFlavour while the parameterList.ts doesn't

We should remove them from `offliner-definition.json` as well for the time being (there is a separate [issue](https://github.com/openzim/mwoffliner/issues/2707) right now about the fact that `--customFlavour` should probably not have been removed, but let's keep things align on the list of parameters, and `parameterList.ts` anyway wins since this is what the scraper is using to accept or refuse a parameter. I.e. it is normal to have more parameters in `parameterList.ts` (more on this a bit later in this comment) but not less

> With respect to the speed flag, there is a conflicting definition between the two files regarding the default value. offliner-definition.json says "Otherwise 1", while src/parameterList.ts states "(by default the number of CPU cores). The default value is 1." This could be confusing to a user as to whether it scales automatically with CPU cores or strictly defaults to 1.

The `parameterList.ts` definition is the correct one, the default is the number of CPU cores

> parameterList.ts contains resume and redis flags but JSON spec doesn't

This is normal, these parameters makes no sense in a "Zimfarm context", like we can use them only when using the scraper directly on a machine.

offliner-definition.json
```diff
@@ -95,13 +95,13 @@
...
       "title": "Filename prefix",
-      "description": "Custome filename up to the formats and date parts."
+      "description": "Custom filename up to the formats and date parts."
     },
     "formats": {
       "type": "list-of-string-enum",
       "required": false,
       "title": "Formats",
-      "description": "Which flavours to build, as `<flavour>:<custom-suffix>`. Empty option is full without suffix.",
+      "description": "Which flavours to build. Empty option is will scrape all article contents, including videos.",
```
benoit74: `is will` => `will`

PR #2718, opened 2026-04-24
placeholder-based zim naming

### LINKED ISSUE #2710: Enhance computation of ZIM name and filename
Current computation of ZIM name and filename is fragile, especially if one wants to match our [ZIM_Naming_Convention](https://wiki.openzim.org/wiki/Content_team/ZIM_Naming_Convention).

As shown in https://github.com/openzim/mwoffliner/pull/2703, we have at least two issues:
- one cannot choose a custom name if multiple variants are asked for
- there is already a problem when multiple flavor types are provided like `nodet,nopic` (instead of nodet,nopic:mini) as they are also joined using `_` (instead of `-` probably)

We should fix the flavor issue and adopt a more "classic" approach like we now have in other scrapers where one can specify custom ZIM name and filename based on placeholders which are populated based on current context (flavour, langOrVariant, domain, ...).

anuragagrawal17:
fix #2710

### DIFF
--- a/src/Dump.ts
+++ b/src/Dump.ts
@@ -12,11 +12,12 @@ interface DumpOpts {
   withoutZimFullTextIndex: boolean
   customZimTags?: string
   customZimLanguage?: string
+  customZimName?: string
   customZimTitle?: string
   customZimDescription?: string
   customZimLongDescription?: string
+  customZimFilename?: string
   mainPage?: string
-  filenamePrefix?: string
   pageList?: string
   resume?: boolean
   minifyHtml: boolean
@@ -75,7 +76,7 @@ export class Dump {
   public computeFlavour() {
     const flavour = []
     if (typeof this.formatFlavour === 'string') {
-      flavour.push(this.formatFlavour)
+      return this.formatFlavour
     } else {
       if (this.nopic) {
         flavour.push('nopic')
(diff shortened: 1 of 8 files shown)


src/Dump.ts
```diff
@@ -87,42 +89,110 @@ export class Dump {
...
+    return flavour.join('-')
+  }
+
+  private computeDomain() {
+    return this.mwMetaData.creator.charAt(0).toLowerCase() + this.mwMetaData.creator.substr(1)
+  }
+
+  private computeLang() {
+    const hostParts = new URL(this.mwMetaData.webUrl).hostname.split('.')
+    let langSuffix = this.mwMetaData.langIso2
```
benoit74: rename from `langSuffix`  to `lang` for clarity

src/Dump.ts
```diff
@@ -87,42 +89,110 @@ export class Dump {
...
+        langSuffix = part
+        break
+      }
+    }
+    return langSuffix
+  }
+
+  private computeSelection() {
+    if (this.opts.articleList) {
+      let filenamePostfix = pathParser
```
benoit74: rename `filenamePostfix` to `selection` for clarity

benoit74 (changes requested):
Thank you, you are on the right track!

Please:
- drop `--filenamePrefix` CLI option ; we are lucky to be preparing a major release so it is exactly the right moment to introduce breaking changes
- drop `flavour_suffix` from potential placeholders, it adds little value but confusion
- rename `computeFilenameRadical` to `computeFilename` and include `.zim` suffix in it
- looks like we can now drop arguments of `computeFilename` (Radical) function since they now look unused 
- all this probably opens the doors to other enhancements / cleanup
- add an entry to changelog; probably worth to specify it is a breaking change; maybe worth to have multiple entries to clarity, I let you decide
- see inline comments

anuragagrawal17:
thank you @benoit74 sir for reviewing my pr, i will get it done by thisweek

benoit74:
@anuragagrawal17 any chance you can complete this PR? Just let me now when / if you have time to work on this. No worries.

kelson42:
@anuragagrawal17 Any feedback? We are preparing release of next milestone.

PR #2721, opened 2026-04-28
Redesign speed to handle both worker count and request pacing

### LINKED ISSUE #2393: Change default speed to something way more conservative
Currently, default speed seems to induce quite a lot of "pressure" on mediawikis.

From someone at minecraft.wiki:

> The way they are scraping it is really bad/resource intensive
Like they are using an action=parse endpoint that is guaranteed to not hit the parser cache with a parallelism of 10
Which basically guarantees they are going to peg 9-10 CPU cores while it's running
Which is just not acceptable for our infra setup, or really anyone's setup besides enwp/fandom

In general, besides some huge Wikimedia wikis, most wikis are pretty small (in the order of 10k articles) and we are definitely not in a hurry to complete them, not speaking about the fact that these could be hosted on very small machines.

I think we should provide more conservative speed setting (no parallelism) so that by default we do not "destroy" the mediawiki instance in front of mwoffliner. And use the speed setting to increase parallelism where required / possible.

Note that https://github.com/openzim/mwoffliner/pull/2392 would probably help as well

triemerge:
Fix #2393

The CPU-count multiplier default was generating too much "pressure" on MediaWiki instances, particularly those not sized for high parallelism. Speed now directly controls behavior: values >= 1 set the worker count, values < 1 increase the inter-request delay for wikis that need gentler scraping.

Changes:
- default speed is 1 (single worker, 100ms inter-request delay)
- float values below 1 slow down further (e.g. 0.1 → 1s delay)
- speed validation moved to `sanitize_speed`

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -50,7 +50,8 @@ imageminOptions.get('webp').set('image/jpeg', {
 
 interface DownloaderOpts {
   uaString: string
-  speed: number
+  workers: number
+  articleRequestInterval?: number
   reqTimeout: number
   optimisationCacheUrl: string
   s3?: S3
@@ -144,7 +145,8 @@ class Downloader {
     }
     return Downloader.instance
   }
-  private _speed: number
+  private _workers: number
+  private _articleRequestInterval: number
   private _webp: boolean = false
   private _requestTimeout: number
   private _basicRequestOptions: AxiosRequestConfig
@@ -164,8 +166,12 @@ class Downloader {
   private articleUrlDirector: URLDirector
   private insecure: boolean = false
 
-  get speed() {
-    return this._speed
+  get workers() {
+    return this._workers
+  }
+
+  get articleRequestInterval() {
+    return this._articleRequestInterval
   }
 
   get webp() {
@@ -190,10 +196,22 @@ class Downloader {
     return this._apiUrlDirector
   }
 
-  set init({ uaString, speed, reqTimeout, optimisationCacheUrl, s3, webp, trustedJs = config.output.mw.js_trusted.slice(), backoffOptions, insecure }: DownloaderOpts) {
+  set init({
+    uaString,
+    workers,
+    articleRequestInterval,
+    reqTimeout,
+    optimisationCacheUrl,
+    s3,
+    webp,
+    trustedJs = config.output.mw.js_trusted.slice(),
+    backoffOptions,
+    insecure,
+  }: DownloaderOpts) {
     this.reset()
     this.uaString = uaString
-    this._speed = speed
+    this._workers = workers
+    this._articleRequestInterval = articleRequestInterval
     this._requestTimeout = reqTimeout
     this.optimisationCacheUrl = optimisationCacheUrl
     this._webp = webp
@@ -301,7 +319,8 @@ class Downloader {
 
   private reset() {
     this.uaString = undefined
-    this._speed = undefined
+    this._workers = undefined
+    this._articleRequestInterval = undefined
     this._requestTimeout = undefined
     this.optimisationCacheUrl = undefined
     this._webp = false
--- a/src/util/const.ts
+++ b/src/util/const.ts
@@ -1,4 +1,4 @@
-export const MAX_CPU_CORES = 8
+export const ARTICLE_REQUEST_INTERVAL = 100 // default pause (ms) between action=parse API requests per worker
 export const IMAGE_THUMB_URL_REGEX =
   /^(.*?\/)(transcoded\/|thumb\/)?([0-9a-fA-F]{1}\/[0-9a-fA-F]{2}\/)?([^/]+\.[A-Za-z0-9]{2,6}\/)?(\d+px[-]+)?([^/]+?\.[A-Za-z0-9]{2,6}(\.[A-Za-z0-9]{2,6})?)(\?[0-9a-fA-F]+|\?utm_.+)?$/
 export const LATEX_IMAGE_URL_REGEX = /^(.*\/math\/render\/svg\/)([A-Za-z0-9]+)$/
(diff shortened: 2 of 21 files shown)


benoit74:
@triemerge thank you, and sorry for the long period without feedback. 

Reading this PR, I wonder if we should not allow float speed values again:
- when `speed` is equal or above 1, it is the number of workers (and must be an integer)
- when `speed` is below 1, it means we have a single worker **and** we use the value to increase the interval between requests: interval = default_interval / speed => speed 0.1 means 1 request every second, ...

I feel like it would make sense to allow to slow down the scraper further then 100ms is still too much for a given website.

WDYT?

triemerge:
Thank you for the suggestion. I like the single knob approach ; simpler than a separate `--delay` parameter.

A few things:
- Is `speed=0.1` intuitive? The parameter is called "speed" but values below 1 mean "go slower". Not sure if this is confusing in practice or if documentation makes it clear enough.
- What should be the minimum? with interval = 100ms / speed, speed=0.01 gives a 10s gap. For a 10k article wiki that is roughly 28 hours of pure delay. I don't think there is a real use case below 0.1 (1 req/s).
- ~~I wonder if this should go in this PR or as a follow-up. Current scope is already a clean breaking change on its own, but I can also do everything in one go if you prefer.~~

WDYT?

benoit74:
> Is speed=0.1 intuitive? 

I'm not sure. It has been used "like this" (with 0.1 value) for years on our side, so our Zimfarm editor(s) are used to it. But I would not say it is intuitive. Probably worth spending a bit of time on parameter description to make it quite obvious with something like "1 is the default, 0.1 make it ten times slower, 10 makes it ten times faster".

> What should be the minimum? with interval = 100ms / speed, speed=0.01 gives a 10s gap. For a 10k article wiki that is roughly 28 hours of pure delay. I don't think there is a real use case below 0.1 (1 req/s).

I would not mind putting a minimum, who knows which requirement we might encounter in the wild.

triemerge:
Updated. Float values below 1 now increase the interval (100ms / speed), no minimum floor. Used your suggested description wording.

src/util/saveArticles.ts
```diff
@@ -167,6 +168,7 @@ export async function saveArticles(zimCreator: Creator, dump: Dump) {
           const articleUrl = Downloader.getArticleUrl(articleId, { sectionId: leadSectionId })
 
           rets = await Downloader.getArticle(articleId, articleDetailXId, RenderingContext.articlesRenderer, articleUrl, dump, articleDetail)
+          await new Promise((resolve) => setTimeout(resolve, Downloader.requestInterval))
```
benoit74: why pausing immediately after `getArticle` returns rather that at the "end" of for loop?

benoit74 (changes requested):
Thank you. See inline comments.

I'm also puzzled by the fact that we have `ARTICLE_REQUEST_INTERVAL` which is then transformed into `request_interval` (which entitles it will be applied to all requests, not only to articles fetching).

Same concern with Downloader.speed which is not the speed but number of workers. Probably worth to rename to Downloader.workers or nb_workers for maintainability. Isn't this setting also used for other operations? Don't we risk to have a conflict between some operations which needs to slow down (because they call a given API) while others could be faster (because they use another API)?

src/util/saveArticles.ts
triemerge: No strong reason. It makes it a bit more obvious the delay is about the action=parse call, but both placements are equivalent here since the loop is serial.

triemerge:
> I'm also puzzled by the fact that we have `ARTICLE_REQUEST_INTERVAL` which is then transformed into `request_interval` (which entitles it will be applied to all requests, not only to articles fetching).
> 
> Same concern with Downloader.speed which is not the speed but number of workers. Probably worth to rename to Downloader.workers or nb_workers for maintainability. 

Renamed

> Isn't this setting also used for other operations? Don't we risk to have a conflict between some operations which needs to slow down (because they call a given API) while others could be faster (because they use another API)?

workers is shared across all operations, yes. But each operation that needs throttling already has its own mechanism. `saveArticles` has the inter-request delay for `action=parse`, `FileManager` has per-host intervals for file downloads. ~~The shared count only controls parallelism, not rate~~. I think we are fine for now, but worth keeping in mind if we ever need different concurrency per operation.

src/Downloader.ts
```diff
@@ -51,6 +51,7 @@ imageminOptions.get('webp').set('image/jpeg', {
 interface DownloaderOpts {
   uaString: string
   speed: number
```
benoit74: Please also rename this `speed` to `workers` for consistency (here and everywhere where it is used obviously)

src/util/const.ts
```diff
@@ -1,4 +1,4 @@
-export const MAX_CPU_CORES = 8
+export const ARTICLE_REQUEST_INTERVAL = 100 // minimum delay (ms) between action=parse API requests per worker
```
benoit74: minimum => default
delay => pause (we do not measure the time between requests issuance but the pause between one request has been processed and next one is started ; should a request takes 1.5 sec to get the response + process it, then requests will be paced at 1.6 sec intervals)

benoit74 (changes requested):
Two (hopefully last) comments, thank you

src/util/saveArticles.ts
benoit74: Okay, let's keep it like this, I don't have strong reasons to do this differently.

src/util/const.ts
triemerge: Good distinction, updated.

src/Downloader.ts
triemerge: Renamed

benoit74 (commented):
LGTM. CI errors look unrelated, we will fix `main` branch and rebase once done so that we can confirm this PR does not break the CI.

triemerge:
Rebased on `main`. Updated commit and changelog accordingly.

benoit74 (approved):
LGTM, thank you!

PR #2736, opened 2026-05-19
Make customCSS flag editable as blob

benoit74:


benoit74:
@elfkuzco FYI, I already have a `css` flag which deserves the `allowRemoteURL` flag ^^

PR #2737, opened 2026-05-21
Add support of Node.js 26

kelson42:
Node.js 26 is the new `current` release.

Blocked by https://github.com/openzim/node-libzim/pull/203 (and it has to be released and integrated)


benoit74:
Superseeded by https://github.com/openzim/mwoffliner/pull/2744

PR #2751, opened 2026-06-10
Improve error message when incorrect (or no) value is passed for --mw…

### LINKED ISSUE #2752: Improve error message when incorrect (or no) value is passed for --mwActionApiPath
Hi, first time user here! This is an awesome project, and I have an idea how you could make it slightly easier to use for first-time users.

I was trying to download https://stardewvalleywiki.com, and initially used this command:
```
docker run -v (pwd)/stardewvalley-wiki:/out -it ghcr.io/openzim/mwoffliner mwoffliner --mwUrl=https://stardewvalleywiki.com --adminEmail=<my-email>
```
which produced the following output:
```
[log] [2026-06-10T18:59:09.934Z] Starting mwoffliner v1.17.5...
[error] [2026-06-10T18:59:10.390Z] Failed to run mwoffliner after [0s]:
 Error: Mediawiki API is not reachable with https://stardewvalleywiki.com/w/api.php?action=query&format=json&formatversion=2&maxlag=5
AxiosError: Request failed with status code 404
    at <anonymous> (/tmp/mwoffliner/src/sanitize-argument.ts:189:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async check_mwApiReachability (/tmp/mwoffliner/src/sanitize-argument.ts:188:17)
    at async check_all (/tmp/mwoffliner/src/sanitize-argument.ts:108:3)
    at async Module.execute (/tmp/mwoffliner/src/mwoffliner.lib.ts:185:3)
```

I was a bit lost, and after skimming the FAQ I read https://github.com/openzim/mwoffliner/wiki/API-end%E2%80%90points, but that didn't help. I ended up searching the web for "mediawiki find API end‐points", which pointed me to the Special:Version page.

Only when I felt that the docs were a bit lacking I realised the FAQ's "Which value for `--mwActionApiPath`?" paragraph does already mention Special:Version.

I feel this can be improved - maybe I just picked a bad wiki to download, but the one other wiki I tried, https://hollowknight.wiki, also has a non-default API path. So maybe this happens often, and it should be easier to find a mention of Special:Version.

Ideally the console output should be improved.

mvforell:
…ActionApiPath

Make it easier for first-time users to find out which value to pass for --mwActionApiPath by pointing to the FAQ.

Fixes #2752.

### DIFF
--- a/src/sanitize-argument.ts
+++ b/src/sanitize-argument.ts
@@ -196,7 +196,7 @@ export async function check_mwApiReachability(mwUrl: string, mwActionApiPath: st
   // We are building an API query "by-hand" because we've not yet initialized the whole URL director / URL builder stack
   const apiQueryUrl = `${new URL(mwActionApiPath || MediaWiki.actionApiPath, mwUrl).toString()}?action=query&format=json&formatversion=2&maxlag=${config.defaults.maxlag}`
   const value = await Downloader.get(apiQueryUrl).catch((err) => {
-    throw new Error(`Mediawiki API is not reachable with ${apiQueryUrl}\n${err}`)
+    throw new Error(`Mediawiki API is not reachable with ${apiQueryUrl}\n${err}\nYou may need to pass a (different) value for --mwActionApiPath, see the FAQ.`)
   })
   const contentType = value.headers['content-type']
   if (!(contentType as string).includes('application/json')) {


benoit74 (commented):
Thank you! 

I'm not really convinced everyone will know where to look after the FAQ. I'm not sure adding a link to https://github.com/openzim/mwoffliner/wiki/Frequently-Asked-Questions#which-value-for---mwactionapipath in the codebase is ideal. WDYT?

mvforell:
The link to the FAQ is at the top of the README and I found it without problems, not sure about others. I agree linking to it in the code probably isn't the best solution.

If you're unsure about referring to the FAQ, we could mention Special:Version directly instead?

benoit74:
My point is that some users might use mwoffliner without knowing the code is hosted on GitHub. But you are right that search engines and the fact that FAQ link is present in README is probably sufficient. Let's merge this as-is.

benoit74 (approved):

kelson42:
I think this PR should not have been merged.... we have many scenarios where the --mwApiPath is not the problem. We should not tell the user what to do if we have no clue about where is the problem.

mvforell:
I think it's still valuable to try and help with an educated guess. But of course the usefulness depends on whether my assumption holds that a non-default API path is responsible for a reasonable amount of the 404 errors.

Personally I can think of two other reasons for the 404 error:
1. passing the wrong value for `--mwUrl`, or
2. passing the the URL for a non-MediaWiki site.

I think for (1) we could augment the message, e.g. something like:
"Double-check the value for --mwUrl. If that's correct, you may need to pass a (different) value for --mwActionApiPath. See the FAQ."

For (2) it's a bit more difficult. Personally I would expand the FAQ, maybe with a sentence like "If the Special:Version page doesn't exist, the page isn't using MediaWiki and can't be downloaded." (Not sure if it could also be related to an outdated version.)

Am I missing other scenarios that could lead to the 404 error?

Markus-Rost:
The value for `mwApiPath` is by far the most common cause of this issue. Mainly because our default value of `/w/api.php` doesn't match the MediaWiki default value of `/api.php`, so most non-WikiMedia wikis will need the parameter provided.

Though personally I feel like we could maybe merge mwApiPath and mwUrl parameter. As mwUrl only takes the host and mwApiPath only takes a path, combining them into the full api path url might be more user friendly. We can then pull both needed values from that single parameter.
Additionally we could add a fallback to try and pull the api path from the article html when an article url is provided. There is a meta element on every MW page which has the needed value.

kelson42:
This guess is not "educated" as there is nothing letting you think the problem comes from there. It can also be anything related to a WAF or any network problem.

What you should to here, is A FAQ entry.

@benoit74 we should really consider rollbacking OR make a piece of code which really try to diagnose the problem.

mvforell:
Ah, my bad, I still had the error message in my mind that said "404 error" but that was only the underlying error. I agree we probably need more sophisticated error handling in the code, then. Unfortunately I don't have the time for that right now.

Issue #2620, opened 2026-02-17
Add support for category pagination


Markus-Rost:
Follow-up to #2245 to fully support category pages including pagination

> I've been looking into this a bit and the main issue I'm seeing is pagination of the category members. With some categories having millions on members, we can't just ignore it. With pagination being split between subcategories, pages, and files, we also can't just have duplicate category copies for each page of the list.
> My idea would be to use iframes and load a separate page of the list for subcategories, pages, and files onto the main category page:
> ```html
> Normal parsed wikitext category
> <iframe src="/_cat_/PAGENAME/subcat/1"></iframe>
> <iframe src="/_cat_/PAGENAME/page/1"></iframe>
> <iframe src="/_cat_/PAGENAME/file/1"></iframe>
> ```
> ```html
> <head>
> <base target="_parent">
> </head>
> <body>
> <h2>Subcategories</h2>
> This category has the following 200 subcategories, out of 400 total.
> (previous page|<a href="2" target="_self">next page</a>)
> <ul>
> <li><a href="/Cookie">Cookie</a></li>
> ...
> </ul>
> </body>
> ``` 

 _Originally posted by @Markus-Rost in [#2245](https://github.com/openzim/mwoffliner/issues/2245#issuecomment-3884489936)_

> Another issue is, that we are currently not using any i18n library and as such can't support the localized messages used on category pages. Those messages need to respect language specific [plural rules](https://translatewiki.net/wiki/Plural). I've created #2614 regarding this.
> 
> Example of a message used by MediaWiki:
> `{{PLURAL:$2|This category has only the following subcategory.|This category has the following {{PLURAL:$1|subcategory|$1 subcategories}}, out of $2 total.}}` 

 _Originally posted by @Markus-Rost in [#2245](https://github.com/openzim/mwoffliner/issues/2245#issuecomment-3885050140)_

Markus-Rost:
For the iframe based pagination to look correctly, the iframed pages need to load the exact same css and js modules as the parent category pages. This will require some reworking of how the action parse renderer works.
While the renderer always returns a list of pages for each article id (probably based on the other renderers which had some basic multi-page category support), there is currently way to add additional pages to this list and the part adding the css module etc also doesn't support a list of pages yet.

benoit74:
While the iframe approach is very promising, it hits a kind of serious blocker due to sandboxing constraints on some readers (see https://github.com/kiwix/libkiwix/issues/1305).

[...]

PR #2754, opened 2026-06-11
Refactor codebase to really support rendering multiple ZIM HTML items for a single MW article

benoit74:
This is a refactoring PR to prepare for #2620 ; the core goal is to allow the renderer to return a list of HTML files to include in the ZIM for a single article retrieved from Mediawiki instance

Changes:
- simplify static files handling: input extension directly in config + stop wasting time computing the list for every article since it is static per renderer (and Mediawiki instance)
- instead of assuming an article will return a list of outputs with each a list of dependencies (images, ...) simplify the logic by assuming an article will return a list of HTML items and list of dependencies (images, ...)
- drop complexities of deeply buried promises in `saveArticle` which was quite flawed (we are already handling a promise per article, there is no point in pushing to many promises in the queue and ending with promises overflow) and useless now

In other words, main PR goal is to move from

```ts
export interface RenderSingleOutput {
  articleId: string
  displayTitle: string
  html: string
  imageDependencies: any
  videoDependencies: any
  mediaDependencies: any
  moduleDependencies: any
  staticFiles: string[]
  subtitles: any
}

export type RenderOutput = RenderSingleOutput[]
```

to

```ts

export type RenderSingleOutput = {
  articleId: string
  zimPath: string
  zimTitle?: string
  htmlContent: string
}

export type RenderOutput = {
  items: RenderSingleOutput[]
  imageDependencies: any
  videoDependencies: any
  mediaDependencies: any
  moduleDependencies: any
  subtitles: any
}
```

Everything else is expected to be pure plumbing changes.

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -24,7 +24,7 @@ import ApiURLDirector from './util/builders/url/api.director.js'
 import urlHelper from './util/url.helper.js'
 
 import ActionParseURLDirector from './util/builders/url/action-parse.director.js'
-import { Renderer } from './renderers/abstract.renderer.js'
+import { Renderer, RenderOutput } from './renderers/abstract.renderer.js'
 import { findFirstMatchingRule, renderDownloadError } from './error.manager.js'
 import RedisStore from './RedisStore.js'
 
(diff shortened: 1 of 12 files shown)


benoit74:
Without any review or even any feedback in due time, I'll consider I shall merge this

Issue #2761, opened 2026-06-16
User login does not handle session expiration


benoit74:
When a username and password are passed, the scraper handles the login operation and session cookies are persisted in the cookie JAR, transparently sent in every eligible request done.

However, there is nothing to detect / handle cookie expiration. 

Looking at Mediawiki documentation, login session expires after:
- [$wgCookieExpiration](https://www.mediawiki.org/wiki/Manual:$wgCookieExpiration) (30 days by default)
- or [$wgObjectCacheSessionExpiry](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgObjectCacheSessionExpiry) of inactivity (1 hour by default)

I strongly suspect this is the problem we faced in https://farm.openzim.org/pipeline/08d0e1b9-8707-4605-a447-66fe3854048e/debug where we wanna create two ZIMs in a row:
- scraper issues a login operation with bot password and cookie is set
- first dump proceeds with regular operation to get articles and everything is fine thanks to high API limits granted to the bot password we use
- first dump move on to downloading files, no more operations are done on mdwiki.org for more than 1 hour since all medias come from upload.wikimedia.org
- session is expired by `$wgObjectCacheSessionExpiry` since no operation occured
- second dump starts without being logged-in anymore (and nothing notices the issue)
- unauthenticated API operations are quickly throttled by unauthenticated user limits

benoit74:
I've confirmed in a local test targeting mdwiki.org that scraper is logged-out after 1h of inactivity.

PR #2762, opened 2026-06-16
Login again at every dump start to workaround issue of expired logins

benoit74:
Workaround for #2761 

Really fixing the issue is hard because we do not want to overwhelm the upstream with additional API calls, we have no clue what expiration could be (could be quite low actually) and we have no clue on regular call regarding whether we are still logged-in or not.

I hence propose to workaround known symptom: the problem occur only when we start a fresh dump, because inactivity period triggers. The other expiration is at 30 days and we have no scraper running that long in Zimfarm. Solution to these conditions is hence straightforward and lightweight: attempt to login again at every dump start and ignore "normal" errors saying you are already logged in.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -32,7 +32,7 @@ Unreleased:
 * UPDATE: Use node-libzim 4.3.0 and Node.JS 26, update dependencies (@benoit74 #2743)
 * FIX: Remove now unused resource files (@benoit74 #2748)
 * CHANGED: *** Breaking *** Use Pino log library, change log level values and CLI argument controlling log level (@benoit74 #2122)
-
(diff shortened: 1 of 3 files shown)


kelson42:
@benoit74 the expiration date is not part of the cookie!?!

benoit74:
@kelson42 it is not the cookie which expires in mdwiki.org case, but the server side session storage (the one controlled with [$wgObjectCacheSessionExpiry](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:$wgObjectCacheSessionExpiry)).

This is not something visible client-side, especially since server tracks inactivity period, not time since initial session creation. And we have no clue about the parameter value.

I still think we should aim to enhance this situation at some point (hence I won't close the issue), but it is not justified to do this in 2.0 milestone.

PR #2764, opened 2026-06-16
Introduce i18next for internationalization handling

### LINKED ISSUE #2614: Use a proper i18n library

### LINKED ISSUE #2750: Implement full i18n of category pages content
Once #2614 is fully implemented, we need to rework i18n of category pages content.

benoit74:
Fix #2614 
Fix #2750

Changes:
- use `i18next` library for internationalization handling
- use `i18next` format in translation files (which forces to update many translation strings for new placeholder format)
- plug translator in dump class (needs to be resolved async)
- fix tests + CI checks


Nikerabbit:
Hi. This format for plurals is not supported by translatewiki.net. Can you make it so that the plural variants are an array? E.g.
```
"categoryFileCount": { "one": "baa", "other": "baas" }
```
That would be supported.

PR #2765, opened 2026-06-16
Drop imagemin and use sharp for image compression

### LINKED ISSUE #1980: mwoffliner drops downloaded images from zim creation if imagemin toolchain fails processing them
https://github.com/openzim/mwoffliner/blob/ad5dc1d6071552c1a9b577fa1a619cfba0fe6938/src/Downloader.ts#L504-L521

The catch in line 505 seems to miss an error handler:  If the imagemin call fails, then a successfully downloaded image will be dropped from ZIM creation.  mwoffliner can do better in these cases: if the size reduction fails, log and take the downloaded original.

-----
As an example for a possibly failing  image see
https://upload.wikimedia.org/wikipedia/commons/thumb/1/1e/Schleiflade_Tontraktur_Animation.gif/120px-Schleiflade_Tontraktur_Animation.gif

Prior to kohler/gifsicle@0e02f7b62a9a3a344c34f92ddb7e178ad3b3e3ff  this file would fail a gifsicle run and thus not be included in any zim even if the resource successfully downloaded..

### LINKED ISSUE #2547: Installation Fails - ARM64 (II)
When trying to install mwoffliner the installtion fails - see attchaed log-file.
I try to install is on a NanoPC-T6, running Ubuntu 24.04 ARM64.

node version: v24.9.0
npm version: 11.6.1

The installation  is failing when trying to install / compile 'optipng-bin'

[2025-10-04T14_02_59_985Z-debug-0.log](https://github.com/user-attachments/files/22700923/2025-10-04T14_02_59_985Z-debug-0.log)

Thank you!

benoit74:
Fix #1980
Fix #2547 

or deprecate both issues, I'm not sure ^^ 

Anyway now ARM64 Docker image achieves to build and we should not have anymore silent conversion failures, we fail the scrape if image conversion fails. There is no known reason for conversion to fail, and fallback to original means we will upload original image to S3 and it will be very hard to know which image failed to convert. For now I prefer we block scraper on conversion issues, and then we will take an informed decision if we realize there are too many good reasons for conversion to fail we can't fix.

Changes:
- CI will run unit tests + build docker image on ARM64 as well (not e2e tests which are just already too long to run)
- CI will publish Docker image on ARM64 as well
- imagemin deps are dropped
- Fail scraper if any image conversion fails


benoit74:
Failing unit test is irrelevant, a usual flaky culprit, and succeeded on ARM64

Issue #1928, opened 2023-10-13
nopdf param doesn't work as expected for some articles


VadimKovalenkoSNF:
`nopdf` param is not working as expected, see https://github.com/openzim/mwoffliner/issues/1916#issuecomment-1746417465 for more details.

benoit74:
This is a big misunderstanding.

The `nopdf` option does not means that we want to remove all links to PDF document but that we do not want to embed PDF files in the ZIM (for PDF files which are hosted on the Mediawiki obviously).

This is covered by e2e tests in `test/e2e/multimediaContent.test.ts` based on https://en.wikipedia.org/wiki/User:Kelson/MWoffliner_CI_reference article:
- when `novid` is used, the ZIM does contain the `Kiwix_-_WikiArabia_Cairo_2017.pdf` pdf
- when `nopdf` is used, the ZIM does not contain this PDF

I will hence create a PR to cleanup test codebase of wrong comments.

PR #2766, opened 2026-06-16
Remove unneeded comments linked to misunderstandings

benoit74:
Relates to #1928

- commented out test is irrelevant, this is not the purpose of `nopdf` option
- doubts expressed around multimedia test really testing all formats one by one is irrelevant, the test even asserts that 4 ZIMs have been created


benoit74:
I won't wait for CI to pass since this is only about deleting comments

PR #2771, opened 2026-06-18
Change translation files format to match translatewiki.net expectations

### LINKED ISSUE #2768: Rework i18n translation key format to match translatewiki constraints
See https://github.com/openzim/mwoffliner/pull/2764#issuecomment-4720970754

@Nikerabbit I feel like this is not supported by i18next library we are using, but we can probably do the required transformation when feeding translations to i18next

benoit74:
Fix #2768

Changes:
- nest plural handling "inside" the parent key for proper handling at translatewiki.net
- unnest before feeding the data to i18next


benoit74:
@Nikerabbit does this look like what you would expect?

Nikerabbit:
Yeah we should be able to handle this format.

PR #2773, opened 2026-06-18
Categories pagination

### LINKED ISSUE #2620: Add support for category pagination

benoit74:
Fix #2620 

This PR contains three commits:
- first one based on original approach based on iframe but which caused https://github.com/kiwix/libkiwix/issues/1305
- second one is an adaptation to rather use JS-based approach
- third one is fixing tests and bugs

See issue for discussion of the pro and cons of the two approaches.

Commits will be squashed once PR is about to get merged and I've confirmed JS approach works fine.

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -139,6 +139,13 @@
       "title": "Add categories",
       "description": "Download category pages"
     },
+    "categoriesPageSize": {
+      "type": "integer",
+      "required": false,
+      "title": "Categories page size",
+      "description": "Number of items (subcategories, pages, files) to display per page on category articles (default is 200)",
+      "min": 1
+    },
     "keepEmptySections": {
       "type": "boolean",
       "required": false,
--- a/res/script.js
+++ b/res/script.js
@@ -1,0 +1,35 @@
+/*
+ * Functions for categories pagination
+ */
+
+document.addEventListener('DOMContentLoaded', () => {
+  showHideJs()
+})
+
+function showHideJs() {
+  for (const item of ['mw-subcategories', 'mw-pages', 'mw-category-media']) {
+    const section = document.getElementById(item)
+    if (!section) {
+      continue
+    }
+    for (const span of section.querySelectorAll('.mwo-no-js')) {
+      span.style.display = 'none'
+    }
+    for (const span of section.querySelectorAll('.mwo-js')) {
+      span.style.display = 'block'
+    }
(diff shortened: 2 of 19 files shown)


benoit74:
A significant point I had to change compared to original issue (besides the move to JS instead of iframe) is the "partials" (chunks of HTML used for pagination) ZIM path: it does not include any additional `/` (besides the ones of original page), so that relative links are easier to write *and* zimcheck doesn't complains links are broken even if they are not supposed to be used "directly" (partials are not supposed to be opened directly)

benoit74:
@Markus-Rost I will merge this to move forward, but do not hesitate to open an issue or another PR if you see something not satisfying you

PR #2778, opened 2026-06-23
Replace article with page everywhere + replace articleId with pageTitle

### LINKED ISSUE #2380: Cleanup the mess between article ID, title, displayTitle, htmlDisplayTitle
The codebase is cluttered with confusions and fallbacks around article ID, title, displayTitle, htmlDisplayTitle

- we must make the interfaces much cleaner: when do we manipulate a title as used in some APIs? when do we manipulate an ID? when do we manipulate HTML code containing article titles?
- we must implement fallbacks only in one place (e.g. if a renderer does not know how to retrieve htmlDisplaytitle, then it is null ; and some code decides what this should be **depending on how it is used**, preferably with some utility function for fallbacks ; because e.g. computing a ZIM title is different than computing a HTML title for the h1 heading, and only the one using this can know what he is dealing with)

I'm postponing this effort until https://github.com/openzim/mwoffliner/issues/2210 has matured, it could be a waste of time to implement this in all renderers if we finally decide to drop support for most of them. Not speaking about the fact that most renderers are currently not heavily tested (automated or manually).

### LINKED ISSUE #2697: Accept both article title and article id in `--articleListToIgnore`
This is probably a minor code change but accepting both article ID and article title in the `--articleListToIgnore` CLI parameter would save lots of headaches to scraper users.

E.g. on dag.wikipedia.org, ignoring https://dag.wikipedia.org/wiki/Sarika_nima_yuya article should be doable with `Sarika_nima_yuya` and `Sarika nima yuya`.

benoit74:
Fix #2697 
Fix #2380

Changes: see https://github.com/openzim/mwoffliner/issues/2697#issuecomment-4754414544


benoit74:
Sorry, just realized I did not updated tests to confirm pageListToIgnore is permissive regarding usage of space and underscore in page title, and I forgot to push the CHANGELOG entry.

docs/functional_architecture.md
```diff
@@ -28,38 +38,38 @@ Currently, scrape uses this call to retrieve:
...
+When using ActionParse renderer, we pass `usepage=1`, which means that we ask the Mediawiki to use the parser configured for this page. This allows the scraper to retrieve page `text` (inner of the HTML containing the page itself) that is as close as possible to what is used online, either Parsoid or legacy parser (see https://www.mediawiki.org/wiki/Parsoid). This is mandatory because if we use a parser different than the one used online, we will de-facto get bugs which are "normal", either because parser still has a bug, or an extension is not compatible, or because Wikitext contains some workaround understandable only by a given parser. This should be avoided at all price if we want to have a versatile scraper capable of processing any Mediawiki. Even focusing only on Wikimedia wikis, not all of them have already transitioned to Parsoid for instance.
```
kelson42: sure about the new argument name?

benoit74: Nice catch, this is a copy/paste error

Markus-Rost: Api parameter is `usearticle=1` and should keep that name.

benoit74: Yep, code is still correct (AFAIK), I just broke documentation

res/style.css
```diff
@@ -50,7 +50,7 @@
   border-bottom: 0px solid #eaecf0;
 }
 
-.mw-body h1.article-header {
+.mw-body h1.page-header {
```
Markus-Rost, line 53: This style seems to be targeted at some MW skin elements. Are you sure this should be changed?

src/util/pages.ts
```diff
@@ -4,24 +4,24 @@ import * as logger from '../Logger.js'
...
 
 /**
- * Check if a given article title or id is main page of the target ZIM
+ * Check if a given page title or id is main page of the target ZIM
  */
-export function isMainPage(articleTitleOrId: string): boolean {
-  return MediaWiki.metaData.mainPage === articleTitleOrId || MediaWiki.metaData.mainPage === articleTitleOrId.replace(/ /g, '_')
+export function isMainPage(pageTitleOrId: string): boolean {
+  return MediaWiki.metaData.mainPage === pageTitleOrId || MediaWiki.metaData.mainPage === pageTitleOrId.replace(/ /g, '_')
 }
```
Markus-Rost: this can probably be simplified now and I assume metaData.mainPage is no longer using underscores?

src/Downloader.ts
```diff
@@ -569,24 +568,23 @@ class Downloader {
       }
       if (errorRule.isHardFailure) {
         logger.info(`This is a hard ${errorRule.detailsMessageKey} error which will be replaced by a placeholder`)
-        dump.status.articles.hardFail += 1
-        dump.status.articles.hardFailedArticleIds.push(articleId)
-        if (dump.maxHardFailedArticles > 0 && dump.status.articles.hardFail > dump.maxHardFailedArticles) {
-          throw new Error('Too many articles failed to download') // eslint-disable-line preserve-caught-error
+        dump.status.pages.hardFail += 1
+        dump.status.pages.hardFailedPageIds.push(pageTitle)
```
Markus-Rost: Shouldn't `hardFailedPageIds` be `hardFailedPageTitles` or just `hardFailedPages` now?

src/Downloader.ts
```diff
@@ -569,24 +568,23 @@ class Downloader {
...
+        if (dump.maxHardFailedPages > 0 && dump.status.pages.hardFail > dump.maxHardFailedPages) {
+          throw new Error('Too many pages failed to download') // eslint-disable-line preserve-caught-error
         }
       } else {
         logger.info(`This is a soft ${errorRule.detailsMessageKey} error which will be replaced by a placeholder`)
-        dump.status.articles.softFail += 1
-        dump.status.articles.softFailedArticleIds.push(articleId)
+        dump.status.pages.softFail += 1
+        dump.status.pages.softFailedPageIds.push(pageTitle)
```
Markus-Rost: Same as for `hardFailedPageIds`

Markus-Rost (changes requested):

test/e2e/pageLists.test.ts
```diff
@@ -7,39 +7,39 @@ import { zimcheckAvailable, zimcheck } from '../util.js'
 
 jest.setTimeout(10000)
 
-const articleList = 'Kiwix,Wikipedia,Internet,Real-time computer graphics'
-const articleListToIgnore = 'Wikipedia, Internet'
+const pageList = 'Kiwix,Wikipedia,Internet,Real-time computer graphics,Wars_of_Augustus,Roman emperor'
+const pageListToIgnore = 'Wikipedia, Internet, Wars of Augustus,Roman_emperor'
```
Markus-Rost, line 11: Do we want `pageListToIgnore` to support case-insensitive first letters on wikis that have `$wgCapitalLinks` enabled? (most wikis except wiktionary)

This would be possible by querying the ignored pages through the API to get their normalized titles, but I'm unsure if this is worth the effort.

res/style.css
benoit74: More or less, I think this is related to `injectH1TitleToHtml` function which injects this class, so AFAIK we are free to choose the class name we want

test/e2e/pageLists.test.ts
benoit74: IMHO querying API looks indeed like a bit too much work and runtime consequences for limited added value.

I'll move this topic to a dedicated issue to discuss separately (and later).

benoit74: Issue is https://github.com/openzim/mwoffliner/issues/2784

PR #2780, opened 2026-06-25
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates


benoit74:
@Nikerabbit looks like we still have something broken:
- why do we have so many files deleted while it looks like they contain good translations?
- why do we still have new contributions to now gone translation keys?

I highly suspect this is my fault, but we need your guidance to fix that situation. Thanks

Nikerabbit:
In those languages, those translations are marked as outdated pending review due to the automatic changes. If there are no non-outdated translations, the files are not created (or deleted).

PR #2820, opened 2026-07-07
Walk the category tree up to find all parent categories

### LINKED ISSUE #2813: Categories: walk the category tree up
While the scraper is nicely capturing categories now, it does not "walk the tree up", i.e. if a category contains only sub-categories, it will not be captured (even if these sub-categories are captured because they contain pages which are in the ZIM).

Looks like we need some form of recursion to ensure we will really capture all categories.

This is particularly painful because these categories with only sub-categories are typically found on Mediawiki main page as "top-level categories", breaking navigation significantly.

Sample case: https://en.wikivet.net/Category:Cat is not captured when doing a full ZIM of en.wikivet.net

benoit74:
Fix #2813


Markus-Rost:
Does this handle category loops correctly?

PR #2846, opened 2026-07-20
Solve issues around empty categories

### LINKED ISSUE #2833: Cannot read properties of null (reading 'categoryinfo')
From https://farm.openzim.org/pipeline/e7b67b96-200a-48ed-a174-36f2e44607cc/debug

```
[2026-07-09 14:47:52.486] ERROR: Error downloading/rendering page Catégorie:Groupe de Saskatchewan
[2026-07-09 14:47:52.487] ERROR: Failed to run mwoffliner after [24482s]:

    err: {
      "type": "TypeError",
      "message": "Cannot read properties of null (reading 'categoryinfo')",
      "stack":
          TypeError: Cannot read properties of null (reading 'categoryinfo')
              at ActionParseRenderer.processHtml (/tmp/mwoffliner/src/renderers/abstract.renderer.ts:671:44)
              at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
              at async Downloader.getPage (/tmp/mwoffliner/src/Downloader.ts:528:14)
              at async <anonymous> (/tmp/mwoffliner/src/util/savePages.ts:162:148)
    }
[2026-07-09 14:47:52.500] INFO: Exiting with code [2]
[2026-07-09 14:47:52.500] INFO: Deleting temporary directory [/tmp/mwoffliner-1783583994179]
```

benoit74:
Fix #2833 

Three fixes:
- do not follow redirects when listing categories since we can have pages attached both to redirect source and target
- always populate categoryMembers (with empty lists at start) so that it can't be undefined, and populate arrays on-the-fly
- prefer infinite loop to list all members instead of recursion (which might ultimately go too deep and is a nightmare in logs stacktraces)

This is meant to superseed https://github.com/openzim/mwoffliner/pull/2840 which was far incomplete

@SAY-5 feedback is welcomed on this PR if any

### DIFF
--- a/src/Downloader.ts
+++ b/src/Downloader.ts
@@ -924,22 +924,27 @@ class Downloader {
     handler(err)
   }
 
-  private async getCategoryMembers(pageTitle: PageTitle, categoryMembers: GroupedCategoryMembers, continueStr = ''): Promise<void> {
+  private async getCategoryMembers(pageTitle: PageTitle, categoryMembers: GroupedCategoryMembers): Promise<void> {
     const apiUrlDirector = new ApiURLDirector(MediaWiki.actionApiUrl.href)
 
-    const { query, continue: cont } = await this.getJSON<any>(apiUrlDirector.buildCategoryMembersURL(pageTitle, continueStr))
-    const items: Array<CategoryMember> = query.categorymembers.filter((a: CategoryMember) => {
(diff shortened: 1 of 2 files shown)


benoit74:
This issue is indeed blocking to many recipes, I will merge right away so that the fix is hopefully working well everywhere. But I would still appreciate you let me know if you spot something wrong in this, we can always fix that in the aftermath.

PR #2847, opened 2026-07-20
Correctly detect SVG mime-types

benoit74:
## Context

Terraria wiki last run showed that we have issues with SVG files: https://farm.openzim.org/pipeline/5b12417a-273d-4d2f-845c-931b978ccb92/debug

Analyzing the issue, it became clear the problem occurs only when we use the optimization cache and the file to retrieve is in the cache.

The root cause is that by default we prefer to detect mime-type from content, and if not available we fallback to the server content-type return in response headers.

The problem is that for SVG the file-type library we are using is not capable to detect from content (it detects only binary formats) and when using S3 cache and file is in the cache, then the server returns a 304 and does not returns the content-type header.

We hence end-up with no content-type, and this is correctly making the scraper ignore the given image.

## Change

This PR adds `@file-type/xml` library which is an addon to `file-type` to also detect XML-based document file types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "GPL-3.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1090.0",
+        "@file-type/xml": "^1.1.1",
         "@ladjs/country-language": "^1.0.3",
         "@openzim/libzim": "4.5.0",
         "@types/async": "^3.2.25",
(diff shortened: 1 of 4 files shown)


kelson42:
Just FYI, this is not the first time we had this problem...

benoit74:
OK, thank you, glad I've probably finally fixed an old problem which only had patches with bad side-effects in edge cases so far.

PR #2866, opened 2026-07-28
add quickstart tutorial in README

### LINKED ISSUE #2797: Transform README.md into tutorial for new-comers
This is a continuation to https://github.com/openzim/zimfarm/issues/1682

README.md should be expended as a tutorial explaining what the tool is and how to use it, rather than instructions for devs.

Goal is not to cover all edge cases but rather document the standard flow to come from "I want to ZIM this Mediawiki" into "I need to run scraper with these arguments". And then advanced instructions can be found in the Wiki FAQ.

elfkuzco:
## Changes
- add quickstart section that explains the essential flags needed to run the scraper and create a ZIM file

This closes #2797

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,10 +28,132 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
 
 Run `mwoffliner --help` to see all available options.
 
-## Prerequisites
+## Quick Start
+
+### Prerequisites
 
 - [Docker](https://docs.docker.com/engine/install/) (or Docker-based engine)
 - amd64 or arm64 architecture
+
+To turn the Bambara Wikipedia into an offline ZIM:
+
+```sh
+mkdir -p output
+docker run -v $(pwd)/output:/output ghcr.io/openzim/mwoffliner \
+  mwoffliner --mwUrl=https://bm.wikipedia.org --adminEmail=you@example.com \
+  --outputDirectory=/output
+```
+
+**NOTE**: In order to avoid [429 responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429), be sure to use an
+actual email address as dummy addresses like `you@example.com` will be flagged by the Wiki servers.
+
+That's it. When it finishes you'll find a `.zim` file in the `output/` directory. Open it with
+[Kiwix Reader](https://get.kiwix.org/en/solutions/applications/kiwix-reader/).
+
+### The Four Decisions
+
+Every scrape comes down to four choices:
+
+#### 1. Which wiki?
+
+```sh
+--mwUrl=https://en.wikipedia.org        # English Wikipedia
+--mwUrl=https://fr.wiktionary.org       # French Wiktionary
+--mwUrl=https://terraria.wiki.gg        # Terraria Wiki
+--mwUrl=https://proofwiki.org           # ProofWiki
+--mwUrl=https://wiki.my-org.com         # Any MediaWiki site
+```
+
+#### 2. Which API path?
+
+Most MediaWiki sites use the default API path `/w/api.php`, but many
+don't. Check by visiting `<wiki-url>/w/api.php` in your browser. If that returns a 404, you need to set `--mwActionApiPath` to the correct path.
+
+> [!TIP]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,11 +28,96 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
...
+To turn the Bambara Wikipedia into an offline ZIM:
+
+```sh
+mkdir -p output
+docker run -v $(pwd)/output:/output ghcr.io/openzim/mwoffliner \
+  mwoffliner --mwUrl=https://bm.wikipedia.org --adminEmail=you@example.com \
+  --outputDirectory=/output --speed=0.1
+```
+
+**NOTE**: In order to avoid [429 responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429), we have set a speed of `0.1`. Also, be sure to use an
```
benoit74: We won't need to set speed to `0.1` anymore, this is now fixed in `main` (or Docker `dev`) which is gonna be released as `2.0.0` soon. So please remove this setting from "default" commands and rather add a section about this in "few more things that you might want". Default speed of 1 is supposed to be fine in most situations. See CLI doc (in `main` branch) for other values meaning, and explain reducing speed might be valuable when seeing lots of HTTP errors in the logs

README.md
```diff
@@ -28,11 +28,96 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
...
+actual email address as dummy addresses like `you@example.com` will be flagged by the
+Wiki servers.
+
+That's it. When it finishes you'll find a `.zim` file in the `output/` directory. Open it with
+[Kiwix Reader](https://get.kiwix.org/en/solutions/applications/kiwix-reader/).
+
+### The Three Decisions
+
+Every scrape comes down to three choices:
```
benoit74: You should speak as well about `--mwActionApiPath`. Default value of `/w/api.php` is fine in most situations, but in many cases it needs to be tweaked, so it is a mandatory decision to make (check if wiki at hand does use default URL). You can use https://terraria.wiki.gg/api.php as an example where `--mwActionApiPath=/api.php` is mandatory.

README.md
```diff
@@ -28,11 +28,96 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
...
+
+#### 1. Which wiki?
+
+```sh
+--mwUrl=https://en.wikipedia.org        # English Wikipedia
+--mwUrl=https://fr.wiktionary.org       # French Wiktionary
+--mwUrl=https://wiki.my-org.com         # Any MediaWiki site
```
benoit74, line 64: Please add https://terraria.wiki.gg/ and https://proofwiki.org/ as other examples

README.md
```diff
@@ -28,11 +28,96 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
...
+```
+
+**NOTE**: `--adminEmail=you@example.com` is also required. It is included in the HTTP
+User-Agent so wiki operators know who is scraping.
+
+### A Few More Things You Might Want
+
+#### Scrape only specific pages
+
+Put one page title per line in a text file, then point to it with `--pageList`:
```
benoit74: You should mention you can also directly input a CSV of page titles

README.md
```diff
@@ -28,11 +28,96 @@ User help is available in the [FAQ](https://github.com/openzim/mwoffliner/wiki/F
...
+
+```sh
+--pageList=./my-pages.txt
+```
+
+#### Scrape a private wiki
+
+```sh
+--mwDomain=corp --mwUsername=jdoe --mwPassword=s3cret
```
benoit74, line 126: Split this in two: domain is rarely used. So first specify username + password, then add that you might use domain as well.

Explain also that username and password should preferably not be the ones of a regular user but rather bot passwords and link to https://www.mediawiki.org/wiki/Manual:Bot_passwords

benoit74 (changes requested):
Thank you

README.md
benoit74: Btw, we should also mention this changed in 2.0.0

benoit74 (approved):
LGTM, thank you. This is a great foundation to build-up upon.

Issue #2798, opened 2026-07-02
Better handle Mediawiki with translated pages


benoit74:
When a Mediawiki has translated pages, we have two problems:
- Translation "widget" (the UI allowing to switch between languages on a given article, not sure it is a widget in MediaWiki sense) is not shown 
- It is not possible to filter pages to ZIM in order to have only one language per ZIM

I'm not sure we should support both. 

AFAIK in many cases translations are partial and having a "mul" ZIM with translations in all languages can help to find all languages (typically I can read French and English, I prefer French but if information is not available I don't mind to read English ; if we have one ZIM per language then I need to switch back-and-forth between them which is just too painful). Being able to filter and create one ZIM per language is maybe not that relevant.

Known cases:
- https://github.com/openzim/zim-requests/issues/1849#issue-4786050196 
- https://github.com/openzim/zim-requests/issues/7
- https://github.com/openzim/zim-requests/issues/526
 
I suspect translation "widget" is JS-based and could easily be added with 2.0, but I did not had a look yet.

Markus-Rost:
A large part would be to handle `Special:MyLanguage/...` links. They will currently all be removed as they point to a special page. We should ideally rewrite them to go directly to the base article instead.

benoit74:
Yes, if we decide to go the `mul` ZIM route, i.e. continue to create a ZIM with all languages inside. I know @Popolechien is quite opposed to this approach and would probably prefer to be able to filter articles by language. Which is maybe simpler for us to implement.

emansom:
> Being able to filter and create one ZIM per language is maybe not that relevant.

Probably not for most readers, though it turns out to matter quite a bit for machine consumers, so a concrete case might be useful input here. We index ZIM content for retrieval with `ibm-granite/granite-embedding-small-english-r2`, an English-only model, and non-English articles do not get harmlessly ignored: they are embedded anyway and then compete with the correct results in the same vector index. The same goes for a language-specific stemmer or a BM25 text search configuration. `archlinux_en_all_maxi_2026-04` declares `<language>eng`, but only 2206 of its 6666 `text/html` entries are English articles, so a large part of what we embed is unusable.

> typically I can read French and English, I prefer French but if information is not available I don't mind to read English ; if we have one ZIM per language then I need to switch back-and-forth between them which is just too painful

That makes sense, and it would be a shame to lose. I do not think these have to be either/or, though: `mul` as the default for readers, with an opt-in flag for the consumers that cannot work with one.

[...]

PR #2867, opened 2026-07-28
Add --keepPageLanguages to scrape only pages in given languages

emansom:
Partially addresses #2798.

Adds `--keepPageLanguages`, filtering on MediaWiki's `pagelanguage`, which `prop=info` already returns in the query made here, so no heuristic is involved. Pages declaring no language are kept.

Redirects need their own lookup, since `prop=redirects` returns no language. Those pointing at a section are skipped while the filter is on: libzim cannot express them, and the HTML entry written instead has the redirect's own title as its only text.

Verified by scraping wiki.archlinux.org twice with the same `--pageList` and comparing the ZIMs entry by entry. The French and Russian articles are gone, as are all entries with non-ASCII titles, while the English and undeclared ones are untouched.

`test:unit` (393 passed, 4 skipped) and `test:e2e` (191 passed, 1 skipped) both pass, with counts identical to the parent commit. 14 new unit tests cover both filters.

The skipped ones are the S3-gated suites, which `describe.skip` when `S3_URL` is unset. That is two tests in `test/unit/s3.test.ts`, two in the gated block of `test/unit/downloader.test.ts`, and the one in `test/e2e/downloadImage.e2e.test.ts`. I have no credentials for the bucket, so they did not run on my machine. Nothing in this change touches them, and CI covers them where the secret is available.

One question: would you like an e2e test for this? I left it out because it needs a wiki that keeps translations in the main namespace, which Wikipedia does not, and that would add a third-party dependency to CI.

### DIFF
--- a/src/mwoffliner.lib.ts
+++ b/src/mwoffliner.lib.ts
@@ -98,6 +98,7 @@ async function execute(argv: any) {
     addNamespaces: _addNamespaces,
     onlyNamespaces: _onlyNamespaces,
     addContentModels: _addContentModels,
+    keepPageLanguages: _keepPageLanguages,
     javaScript: _javaScript,
     addModules: _addModules,
     customZimFavicon,
@@ -261,6 +262,13 @@ async function execute(argv: any) {
         .map((a: string) => a.trim())
     : []
 
+  const keepPageLanguages = _keepPageLanguages
+    ? String(_keepPageLanguages)
+        .split(',')
+        .map((a: string) => a.trim())
+        .filter((a: string) => a.length > 0)
+    : []
+
   const onlyNamespaces = _onlyNamespaces
     ? String(_onlyNamespaces)
         .split(',')
@@ -415,7 +423,7 @@ async function execute(argv: any) {
 
   logger.debug('Getting pages details')
   let stime = Date.now()
-  await getPages(mainPage, pages, pagesToIgnore, allowedContentModels)
+  await getPages(mainPage, pages, pagesToIgnore, allowedContentModels, keepPageLanguages)
   logger.info(`Got pages details in ${(Date.now() - stime) / 1000} seconds`)
 
   // Getting total number of pages from Redis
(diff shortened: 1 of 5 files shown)


src/util/mw-api.ts
```diff
@@ -128,9 +160,11 @@ async function processPagesAndSaveToRedis(pages: QueryMwRet, pagesToIgnore: Page
     filteredPages.splice(filteredPages.indexOf(page), 1)
   }
 
+  const keptRedirects = await filterRedirectsByLanguage(redirects, keepPageLanguages)
```
benoit74, line 163: Probably worth an explanation to say that:
- this incurs additional API calls when language filtering is active (but only then)
- it is needed to cope with situation where redirect language is different than page language

src/mwoffliner.lib.ts
```diff
@@ -676,6 +684,20 @@ async function execute(argv: any) {
...
+          // redirect to an anchor, so the fallback below writes a real HTML
+          // entry whose only text is the redirect's own title. A redirect can
+          // carry a translated title while the wiki still reports it in the
+          // default language. Measured on wiki.archlinux.org, "X11 转发" and
+          // "Záloha celého systému pomocí rsync" both report pagelanguage=en,
+          // so filtering by language cannot remove them, and they would be the
+          // only text in the archive outside the requested languages. The
+          // target article stays; what is lost is a shortcut to a section of
+          // it, exactly as for 'nodet'.
+          if (keepPageLanguages.length) {
```
benoit74, line 697: I'm not really convinced by this. Why do `Záloha celého systému pomocí rsync` and `X11 转发` both report language `en`? Isn't this something which should be reconfigured at wiki level? Is it possible? Why do we have a chinese title redirecting to an english page?

I'm kinda concerned this will remove many English fragment-based redirect to an English page, which is way more legit.

The fact that libzim cannot express a redirect to an anchor and adds a real HTML entry which is empty is irrelevant, there is always a `refresh` meta in HTML head which redirects to correct ZIM entry and section.

src/parameterList.ts
```diff
@@ -36,6 +36,7 @@ export const parameterDescriptions = {
   webp: 'Convert all jpeg, png and gif images to webp format',
   addNamespaces: 'Force additional namespace (comma separated numbers)',
   addContentModels: 'Include additional content models besides wikitext (comma separated, e.g. "Scribunto,json"). By default, only wikitext pages are scraped.',
+  keepPageLanguages: 'Include only pages whose MediaWiki page language matches (comma separated ISO codes, e.g. "en,simple"). By default, pages of every language are scraped.',
```
benoit74, line 39: I would rather use `onlyPageLanguages` as parameter name to better match existing parameters naming logic.

And I don't think Mediawiki is using ISO code for this parameter, e.g. `simple` is not ISO AFAIK.

benoit74 (changes requested):
Thanks a lot for contributing this, much appreciated.

I've left some inline feedback / questions.

You will need to also add the parameters in `offliner-definition.json` (mostly identical change to what you've done in `parameterList.ts` but in a different format for usage by Zimfarm) and please also add an entry in CHANGELOG.md

It looks like this PR would work well on https://wiki.gentoo.org/, but for some reason it looks like this will completely fail on https://gramps-project.org/wiki/index.php where page language seems to always be set to `en` (any idea why?).

@kelson42 are you OK to consider merging this in 2.0.0?

PR #2888, opened 2026-08-30
Make --pageListToIgnore support case-insensitive first letter

### LINKED ISSUE #2784: Make --pageListToIgnore support case-insensitive first letter
See https://github.com/openzim/mwoffliner/pull/2778#discussion_r3474211847

sundeep8967:
## Description
MediaWiki page titles are case-insensitive on their initial letter and normalize spaces and underscores. When users specify `--pageListToIgnore`, entries previously failed to match if the first letter's case or space/underscore formatting differed between the command line and MediaWiki API responses.

This PR introduces `isPageIgnored` to normalize:
1. First-letter case variations (`Earth` / `earth`).
2. Space vs underscore formatting (`Solar System` / `Solar_System`).

Applied across `filterPages` and category traversal in `src/util/mw-api.ts`. Added unit tests in `test/unit/util/filterPages.test.ts`.

Fixes #2784

### DIFF
--- a/src/util/mw-api.ts
+++ b/src/util/mw-api.ts
@@ -4,7 +4,7 @@ import Downloader from '../Downloader.js'
 import Timer from './Timer.js'
 import RedisStore from '../RedisStore.js'
 import MediaWiki from '../MediaWiki.js'
-import { cleanupAxiosError } from './misc.js'
+import { cleanupAxiosError, lcFirst, ucFirst } from './misc.js'
 
 const MAX_TITLES_PARAM_SIZE = 7400
 const MAX_BATCH_SIZE = 50
@@ -57,6 +57,24 @@ export async function getPagesByTitle(
   )
 }
 
+export function isPageIgnored(title: string, pagesToIgnore: PageTitle[]): boolean {
+  if (!pagesToIgnore || pagesToIgnore.length === 0 || !title) {
+    return false
+  }
+  const normTitle = title.replace(/_/g, ' ')
+  if (MediaWiki.case === 'case-sensitive') {
+    return pagesToIgnore.some((ignored) => (ignored as string).replace(/_/g, ' ') === normTitle)
+  }
+
+  const normTitleUc = ucFirst(normTitle)
+  const normTitleLc = lcFirst(normTitle)
+
+  return pagesToIgnore.some((ignored) => {
+    const normIgnored = (ignored as string).replace(/_/g, ' ')
+    return normIgnored === normTitle || normIgnored === normTitleUc || normIgnored === normTitleLc
+  })
+}
+
 export function filterPages(pages: QueryMwRet, pagesToIgnore: PageTitle[], allowedContentModels: string[]) {
   function revisionFilter(page: PageInfo & QueryRet): boolean {
     return !page.revisions
@@ -67,7 +85,7 @@ export function filterPages(pages: QueryMwRet, pagesToIgnore: PageTitle[], allow
   }
 
   function ignoredPagesFilter(page: PageInfo & QueryRet): boolean {
-    return pagesToIgnore.includes(page.title) || pagesToIgnore.includes((page.title as string).replace(/ /g, '_') as PageTitle)
+    return isPageIgnored(page.title as string, pagesToIgnore)
   }
 
   // Filter pages without revisions (#2091)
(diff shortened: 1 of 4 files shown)


src/util/mw-api.ts
```diff
@@ -57,6 +57,36 @@ export async function getPagesByTitle(
   )
 }
 
+export function isPageIgnored(title: string, pagesToIgnore: PageTitle[]): boolean {
+  if (!pagesToIgnore || pagesToIgnore.length === 0 || !title) {
+    return false
+  }
+  const titleSpace = title.replace(/_/g, ' ')
```
benoit74: This logic seems extremely complex and compute intensive.

I feel like we should:
- normalize pagesToIgnore list when received (I would suggest normalization is always upper-case first letter, always underscore instead of spaces) at scraper startup
- normalize title with same logic in this function
- simply check if normalized title is included in pagesToIgnore

benoit74 (changes requested):
Thank you. I feel like we could do something simpler.

benoit74:
Please also add an entry in CHANGELOG.md

sundeep8967:
thanks for the feedback! simplified the matching logic using `ucFirst`/`lcFirst` and added an entry to `Changelog`.

benoit74:
Please take https://github.com/openzim/mwoffliner/issues/2784#issuecomment-5476690395 into account.

Please do not merge from main branch to feature branch, this is against our policy. We will *rebase* once the PR is ready to be merged (you might rebase only if there is some change in `main` you need for PR to work).

sundeep8967:
Thanks @benoit74 for the feedback and guidance.

- **Rebase/Merge Policy:** Apologies for the merge commit — I have reset the branch to remove the merge from `main`. The PR is now a single clean commit on top of the original base.
- **Case-Sensitive Wikis:** Updated `isPageIgnored` to check the wiki's case configuration from MediaWiki `siteinfo` (`MediaWiki.case`). On case-sensitive wikis (`case === 'case-sensitive'`), first-letter folding is skipped so exact title case is preserved, while on standard wikis (`case === 'first-letter'`), first-letter case variations are normalized. Added unit test coverage for both modes.

src/util/mw-api.ts
```diff
@@ -57,6 +57,24 @@ export async function getPagesByTitle(
   )
 }
 
+export function isPageIgnored(title: string, pagesToIgnore: PageTitle[]): boolean {
```
benoit74, line 60: `title` should be a `PageTitle` as well, especially since all its callers have a `PageTitle` available

src/util/mw-api.ts
```diff
@@ -57,6 +57,24 @@ export async function getPagesByTitle(
...
+  if (!pagesToIgnore || pagesToIgnore.length === 0 || !title) {
+    return false
+  }
+  const normTitle = title.replace(/_/g, ' ')
```
benoit74, line 64: title being a `PageTitle`, you do not need to normalize anymore

src/util/mw-api.ts
```diff
@@ -57,6 +57,24 @@ export async function getPagesByTitle(
...
+  if (MediaWiki.case === 'case-sensitive') {
+    return pagesToIgnore.some((ignored) => (ignored as string).replace(/_/g, ' ') === normTitle)
```
benoit74, line 66: `ignored` is a `PageTitle`, so it is already normalized (there is no `_` left)

src/util/mw-api.ts
```diff
@@ -57,6 +57,24 @@ export async function getPagesByTitle(
...
+  }
+
+  const normTitleUc = ucFirst(normTitle)
+  const normTitleLc = lcFirst(normTitle)
+
+  return pagesToIgnore.some((ignored) => {
+    const normIgnored = (ignored as string).replace(/_/g, ' ')
```
benoit74, line 73: no need to normalize, same remark

benoit74 (changes requested):

src/util/mw-api.ts
sundeep8967: Good catch. Since `extractPageList` and upstream API handlers already normalize titles to `PageTitle` (spaces instead of underscores), I have removed the redundant regex replacements and typed `title` as `PageTitle` directly.

sundeep8967:
Updated the PR to address the review feedback:

- Simplified `isPageIgnored` by removing redundant `.replace(/_/g, ' ')` calls on `title` and `pagesToIgnore` elements, relying directly on the normalized `PageTitle` guarantee.
- Typed `title: PageTitle` in `isPageIgnored` and removed redundant type-casts across callers.
- Re-verified unit tests and linting cleanly.

src/util/mw-api.ts
sundeep8967: Acknowledged and updated. Removed the redundant normalization on `ignored` as well as on `title`, performing direct comparison against `title`, `ucFirst(title)`, and `lcFirst(title)`.

src/util/mw-api.ts
sundeep8967: Acknowledged. Removed `.replace(/_/g, ' ')` on `ignored` across the function.

PR #2889, opened 2026-08-30
Add support for namespaces by name in CLI arguments

### LINKED ISSUE #2879: Add support for namespaces by name in CLI arguments
Bouncing off https://github.com/openzim/zim-requests/issues/1885, I suspect there is no easy way of knowing which namespaces are active. 

Would it be possible to enter the namespace's name as a text string instead of its number, and have mwoffliner identify that  `thesaurus` is 106 in French and 110 in English and act accordingly?

sundeep8967:
## Description
Allows specifying namespaces by text string names (e.g. `thesaurus`, `portal`) in addition to numerical IDs in `--addNamespaces` and `--onlyNamespaces`.

Namespace names are matched case-insensitively against:
- Local language namespace names (e.g. `Thésaurus`)
- English canonical names (e.g. `Thesaurus`)
- Namespace aliases (e.g. `Wiktionnaire:Thésaurus`)
- Numerical IDs (e.g. `106`)

Added unit tests in `test/unit/namespaces.test.ts`.

Fixes #2879

### DIFF
--- a/src/MediaWiki.ts
+++ b/src/MediaWiki.ts
@@ -344,7 +344,21 @@ class MediaWiki {
     }
   }
 
-  public setNamespaces(json: SiteInfoQueryResponse, addNamespaces: number[], onlyNamespaces: number[]) {
+  public setNamespaces(json: SiteInfoQueryResponse, addNamespaces: (number | string)[], onlyNamespaces: (number | string)[]) {
+    const matchesNamespace = (list: (number | string)[], id: number, name: string, canonical?: string, alias?: string): boolean => {
+      return list.some((item) => {
+        if (typeof item === 'number' || !isNaN(Number(item))) {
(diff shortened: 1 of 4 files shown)


benoit74 (changes requested):
I know this was here before you came, but I feel it's very weird that `Mediawiki.namespacesToMirror` contains a list of "names" which could be either name or aliases ... and then we only use the number of the namespaces which matches this name.

Wouldn't this PR be a good opportunity to change this and rather add namespaces numbers in `Mediawiki.namespacesToMirror`?

sundeep8967:
thanks for the great suggestion! storing namespace numbers directly in `Mediawiki.namespacesToMirror` makes total sense and simplifies the resolution logic. will update this PR accordingly.

PR #2883, opened 2026-08-28
fix(savePages): do not trim down mini ZIM files landing page

sundeep8967:
## Problem
When building `mini` ZIM files with `--nodet`, `getPageRenderUrl` restricted all pages to `config.filters.leadSectionId` ('0'), stripping anything below the lead paragraph from the bespoke landing / main page.

## Solution
Updated `getPageRenderUrl` in `src/util/savePages.ts` to verify `!isMainPage(pageDetail.title)` before applying the `leadSectionId` filter, allowing the bespoke landing page to render with its full curated layout.

Fixes #2882

PR #2890, opened 2026-08-30
Do not trim down `mini` ZIM files' landing page

### LINKED ISSUE #2882: Do not trim down `mini` ZIM files' bespoke landing page
Because of the way `mini` is built anything below the lead paragraph is removed:

Image

Image

Image

Maybe it's an issue for mwoffliner, but could we make it so that `mini`-building rules do not apply to the landing page?

sundeep8967:
## Description
When generating `mini` / `nodet` flavour ZIMs, articles are trimmed to the lead section (section 0) to reduce size. However, bespoke landing pages (the main page of the ZIM) contain essential portal structures, introductory navigation, or curated categories that should not be truncated to only the lead paragraph.

This PR:
1. Updates `getPageRenderUrl` in `src/util/savePages.ts` to only request the `leadSectionId` when `!isMainPage(pageDetail.title)`.
2. Updates DOM cleaning in `src/renderers/abstract.renderer.ts` to exempt the landing page from section trimming and citation stripping in `nodet` mode.

Fixes #2882


benoit74:
@sundeep8967 looks like you forgot you've already opened https://github.com/openzim/mwoffliner/pull/2890 ; please make a single contribution per issue! Your additional changes in this PR makes sense btw.

sundeep8967:
Thanks @benoit74 for pointing this out. I have closed #2883 and will maintain all work for #2882 here in #2890.

src/renderers/abstract.renderer.ts
```diff
@@ -661,7 +661,7 @@ export abstract class Renderer {
     // applyOtherTreatments must run before treatMedias so that iframe
     // placeholders (e.g. YouTube thumbnails) are already in the DOM
     // when treatMedias processes <img> tags for download into the ZIM.
-    doc = await this.applyOtherTreatments(doc, dump, pagePath)
+    doc = await this.applyOtherTreatments(doc, dump, pagePath, isMainPage(pageTitle))
```
benoit74: `isMainPage` is different from `isLandingPage` ; it does not take into account a custom main page which might have been set or computed automatically.

test/e2e/miniLandingPage.e2e.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { zimdump, zimcheck } from '../util.js'
+import 'dotenv/config.js'
+import { jest } from '@jest/globals'
+import { rimraf } from 'rimraf'
+
+jest.setTimeout(60000)
+
+const parameters = {
+  mwUrl: 'https://en.wikipedia.org',
+  pageList: 'Earth,Main Page',
```
benoit74: With this page list, there is no landing page but a custom list of these articles ; this test makes no sense, the only reason it pass is because you've included `Main Page` in the list. But in fact here both pages should still be trimmed to lead section.

Please fix test (both pages are trimmed) + add other tests (no page list passed, should be done in bm.wikipedia.org test case already existing to avoid ZIMing yet another full Mediawiki, + a custom main page set + a single page passed in pageList

Changelog
```diff
@@ -19,6 +20,7 @@ Unreleased
 * CHANGED: *** Breaking *** Remove deprecated CLI parameters `mwWikiPath`, `mwIndexPhpPath`, `keepEmptyParagraphs` (@Markus-Rost #2489)
 * CHANGED: *** Breaking *** Reintroduce support of --customFlavour option (beware API has changed due to #2380) (@benoit74 #2707)
 * CHANGED: *** Breaking *** Drop webp polyfill (@benoit74 #2758)
+>>>>>>> upstream/main
```
benoit74: ?

benoit74 (changes requested):
Thank you! See inline comments.

sundeep8967:
Thanks @benoit74 for the thorough review!

I have updated the implementation to address all review points:

1. **`isMainPage` vs `isLandingPage`:**
   - Introduced `isLandingPage(pageTitle: PageTitle, dump?: Dump): boolean` in `src/util/pages.ts` which inspects `dump?.opts?.mainPage` (falling back to `MediaWiki.metaData?.mainPage`). This properly accounts for `customMainPage`, single-entry `pageList`, and multi-page `pageList` (where `mainPage` is empty string).
   - Updated `savePages.ts` (`getPageRenderUrl`) and `abstract.renderer.ts` (`processHtml`) to use `isLandingPage(pageTitle, dump)`.

2. **Test Coverage:**
   - Updated `test/e2e/miniLandingPage.e2e.test.ts` to cover:
     - Multi-page `pageList` without `customMainPage` (both pages are trimmed to lead section).
     - `customMainPage` set (custom landing page is preserved while standard articles are trimmed).
     - Single page passed in `pageList` (automatically designated as landing page and preserved).
   - Added landing page preservation verification in `test/e2e/bm.e2e.test.ts` for when no `pageList` is passed.
   - Added unit tests in `test/unit/renderers/processHtml.test.ts` and `test/unit/util/pages.test.ts` to verify `isLandingPage` resolution and HTML section/citation trimming rules across all combinations.

3. **Changelog:**
   - Removed the stray merge marker (`>>>>>>> upstream/main`).

test/e2e/bm.e2e.test.ts
```diff
@@ -113,3 +122,41 @@ await testAllRenders('bm-wikipedia-only-ns-12', { ...parameters, onlyNamespaces:
...
+ * The wiki's default main page should be the ZIM landing page and NOT be trimmed.
+ * A known regular article should be trimmed to the lead section only.
+ */
+await testAllRenders('bm-wikipedia-nodet', { ...parameters, format: ['nodet,nopic'] }, async (outFiles) => {
+  test(`test ZIM integrity for ${outFiles[0]?.renderer} renderer (nodet)`, async () => {
+    await expect(zimcheck(outFiles[0].outFile)).resolves.not.toThrow()
+  })
+
+  test(`landing page (wiki main page) is not trimmed in nodet mode for ${outFiles[0]?.renderer} renderer`, async () => {
+    // bm.wikipedia.org main page is "Hadamantan"
```
benoit74: I don't get this, `Nyɛ_fɔlɔ` is the main page of bm.wikipedia.org, isn't it?

test/e2e/bm.e2e.test.ts
```diff
@@ -113,3 +122,41 @@ await testAllRenders('bm-wikipedia-only-ns-12', { ...parameters, onlyNamespaces:
     }
   })
 })
+
+/**
+ * Scenario: no pageList (full wiki scrape in nodet mode).
+ * The wiki's default main page should be the ZIM landing page and NOT be trimmed.
+ * A known regular article should be trimmed to the lead section only.
+ */
+await testAllRenders('bm-wikipedia-nodet', { ...parameters, format: ['nodet,nopic'] }, async (outFiles) => {
```
benoit74: Why running this separately, with both `nodet` and `nopic` formats and finally testing only `outFiles[0]` so only the nodet file?

I feel like it would make more sense to add `nodet` to the list of formats on line 16, and test both `Nyɛ_fɔlɔ` and `Mali` page content in both nopic and nodet mode.

I'm also wondering if using bm.wikipedia.org makes any sense after all, since at first glance it looks like it has only one section on main page.

test/e2e/bm.e2e.test.ts
```diff
@@ -71,6 +73,13 @@ await testAllRenders('bm-wikipedia', parameters, async (outFiles) => {
     expect(redirectFromDump).toContain('<a href="./Bamakɔ#Monumentsa_ni_yɔrɔ">Bamakɔ Monumentsa</a>')
   })
 
+  test(`test landing page is not trimmed in nodet mode for ${outFiles[0]?.renderer} renderer`, async () => {
+    // When no page list is passed, the wiki main page is the landing page and is not trimmed in nodet mode
+    const mainPageFromDump = await zimdump(`show --url "Nyɛ_fɔlɔ" ${outFiles[0].outFile}`)
+    const mainPageDoc = domino.createDocument(mainPageFromDump)
+    expect(mainPageDoc.body.innerHTML.length).toBeGreaterThan(500)
```
benoit74: This is a weak test, should the lead section grow bigger than 500 then the test will fail while scraper still works correctly. Can't we search for sections in the page and ensure there is multiple when needed, and only one otherwise?

test/e2e/miniLandingPage.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  'mini-custom-main-page',
+  {
+    mwUrl: 'https://en.wikipedia.org',
+    pageList: 'Earth,Main_Page',
+    customMainPage: 'Main_Page',
+    adminEmail: 'test@kiwix.org',
+    format: ['nodet,nopic'],
+  },
+  async (outFiles) => {
+    const mainPageFromDump = await zimdump(`show --url "Main_Page" ${outFiles[0].outFile}`)
```
benoit74: please test both formats files, here you create both nopic and nodet but you test only nodet

test/e2e/miniLandingPage.e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    pageList: 'Main_Page',
+    adminEmail: 'test@kiwix.org',
+    format: ['nodet,nopic'],
+  },
+  async (outFiles) => {
+    const mainPageFromDump = await zimdump(`show --url "Main_Page" ${outFiles[0].outFile}`)
+    const mainPageDoc = domino.createDocument(mainPageFromDump)
+
+    describe('single page pageList landing page preservation', () => {
+      test(`zimcheck for ${outFiles[0].renderer} renderer`, async () => {
```
benoit74: please test both formats files, here you create both nopic and nodet but you test only nodet

benoit74 (changes requested):
Please see comments below

I also find the term "landing page" a bit confusing. I would prefer we use the term "ZIM main page" for clarity. I.e. rename `isLandingPage` method to `isZimMainPage`

sundeep8967:
Thanks for the review @benoit74!

I have addressed all the feedback in the latest commit:
- Renamed `isLandingPage` to `isZimMainPage` across the codebase and tests for consistency and clarity.
- Updated `test/e2e/miniLandingPage.e2e.test.ts` to loop over and assert both `nodet` and `nopic` dump files across all test scenarios.
- Reverted the additions in `test/e2e/bm.e2e.test.ts` to keep the test suite clean and avoid redundant full-wiki scrapes on a wiki whose main page only has a lead section.

All typechecks, linting, formatting, and unit tests pass cleanly.

src/util/pages.ts
```diff
@@ -2,14 +2,23 @@ import MediaWiki from '../MediaWiki.js'
...
  */
 export function isMainPage(pageTitle: PageTitle): boolean {
   return MediaWiki.metaData.mainPage === pageTitle
 }
 
+/**
+ * Check if a given page title is the main page of the target ZIM
+ */
+export function isZimMainPage(pageTitle: PageTitle, dump?: Dump): boolean {
+  const landing = dump?.opts?.mainPage ?? MediaWiki.metaData?.mainPage
```
benoit74: Under which condition do we have `dump?.opts?.mainPage` being `null` or `undefined`? I don't get when this could happen, and why we would fallback to `MediaWiki.metaData?.mainPage` for checking ZIM main page.

src/util/pages.ts
```diff
@@ -2,14 +2,23 @@ import MediaWiki from '../MediaWiki.js'
...
+  return Boolean(landing && landing === pageTitle)
```
benoit74: why not just `return landing === pageTitle`?

benoit74 (changes requested):
Thank you, we are on the good track, very close to merge actually, but I have two more remarks.

src/util/pages.ts
sundeep8967: You are completely right. Every `Dump` instance has its `opts.mainPage` initialized during creation in `mwoffliner.lib.ts` (set to either `customMainPage`, the single-entry page, or `mwMetaData.mainPage`, or `''` for multi-page lists). `dump` is always provided at all call sites, so there is no reason to fall back to `MediaWiki.metaData?.mainPage` or use optional chaining.

I've simplified this to `return dump.opts.mainPage === pageTitle` directly.

src/util/pages.ts
sundeep8967: Updated! Simplified directly to `return dump.opts.mainPage === pageTitle`.

sundeep8967:
Thanks @benoit74!

Addressed both comments in the latest commit:
1. Simplified `isZimMainPage(pageTitle: PageTitle, dump: Dump): boolean` to directly evaluate `return dump.opts.mainPage === pageTitle`.
2. Removed redundant fallback to `MediaWiki.metaData?.mainPage` and unnecessary `Boolean()` wrapping.
3. Updated unit tests in `test/unit/util/pages.test.ts` accordingly.

All unit tests, linter (`eslint`), and formatting (`prettier`) pass cleanly.

benoit74 (changes requested):
Thank you, please squash commits into a single one.

sundeep8967:
@benoit74 Done! Commits squashed into a single clean commit with verified DCO sign-off.

benoit74:
@sundeep8967 do not hesitate to rebase on main and force push to your branch, I've fixed CI on `main`.

sundeep8967:
@benoit74 Thanks for the update! Rebased onto latest main and force-pushed.

PR #2891, opened 2026-08-30
Selectively remove page title from bespoke landing page

### LINKED ISSUE #2870: Selectively remove page title from bespoke landing page
Because in many case we end up creating bespoke landing page, we arrive to situations where a "wrong" page title is displayed and this looks Deeply Unserious. Case in point:

Image

Turns out that this page title is currently already hidden from the canonic / default landing page, be it on-wiki or on-zim:

Image

or

Image

Shouldn't MWoffliner take into account that any page listed in the `Main page` field should get a similar treatment?

sundeep8967:
## Description
MediaWiki skins hide the top level page title (`#firstHeading`) on the canonical `Main_Page` when configured with an empty `mainpage-title` system message. When bespoke landing pages are configured for a ZIM, they previously displayed their page heading because only the hardcoded CSS class `page-Main_Page` was checked.

This PR updates `ActionParseRenderer` in `src/renderers/action-parse.renderer.ts` to check `isMainPage(pageTitle)` in addition to `page-Main_Page`, ensuring custom landing pages properly inherit the configured display title or hidden heading behaviour.

Fixes #2870

### DIFF
--- a/src/renderers/action-parse.renderer.ts
+++ b/src/renderers/action-parse.renderer.ts
@@ -232,7 +232,10 @@ export class ActionParseRenderer extends Renderer {
 
     // Main page display title
     let hideFirstHeading = false
-    if (isMainPage(pageTitle) || bodyCssClass.split(' ').includes('page-Main_Page')) {
+    const isCustomMainPage = Boolean(dump?.opts?.customMainPage && (pageTitle === dump.opts.customMainPage.replace(/_/g, ' ') || isMainPage(pageTitle)))
+    if (isCustomMainPage) {
+      hideFirstHeading = true
+    } else if (bodyCssClass.split(' ').includes('page-Main_Page')) {
       const mainpageTitle = await getMainpageTitle()
       if (mainpageTitle === '') {
         hideFirstHeading = true
--- a/src/renderers/action-parse.renderer.ts
+++ b/src/renderers/action-parse.renderer.ts
@@ -9,7 +9,7 @@ import { htmlVectorLegacyTemplateCode, htmlVector2022TemplateCode, htmlFallbackT
 import Downloader, { DownloadError } from '../Downloader.js'
 import { customCssUrlToFilename, customJsUrlToFilename } from '../util/customCssJs.js'
 import Gadgets from '../Gadgets.js'
-import { extractJsConfigVars, extractBodyCssClass, extractHtmlCssClass, getMainpageTitle } from '../util/pages.js'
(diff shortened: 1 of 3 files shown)


benoit74:
This is a bit different than what I have in mind, but there could be added value in this as well.

With this PR change, instead of systematically hiding the custom main page title as proposed in the issue, the proposed code is going to either replace the custom main page title with the Wiki main page title if configured (AFAIK this is a Mediawiki setting, not something configured on the main page itself), or hide the custom main page title when the Wiki main page title is missing / empty.

I see added value in this behavior (automatically fetching "proper" title even when using a custom main page) but risks as well (we still cannot fully control the custom main page look and feel, something is automatically changed by the scraper). I feel like the risk is higher than the added value and would propose to go back to original idea of automatically hiding the custom main page title in all cases. Builders of custom main page can always add their own "fake title" in the page body

@Popolechien @kelson42 WDYT?

Popolechien:
> hide the custom main page title when the Wiki main page title is missing / empty

Fine by me, though I can't think of a page (besides landing) where the title would be missing.

I was also going to ask about making a different / bespoke main page title and this seems to address it. Thanks.

sundeep8967:
thanks for the discussion! happy to align with automatically hiding the custom main page title in all cases as originally suggested.

benoit74:
@sundeep8967 please do so, automatically hiding the custom main page title in all cases as originally suggested (and only on custom main page, not wiki default main page).

sundeep8967:
Updated the implementation as requested: the page title heading is now automatically hidden unconditionally when a custom main page is rendered, while retaining the standard `getMainpageTitle()` lookup for the wiki default main page.

src/renderers/action-parse.renderer.ts
```diff
@@ -232,8 +232,10 @@ export class ActionParseRenderer extends Renderer {
 
     // Main page display title
     let hideFirstHeading = false
-    if (bodyCssClass.split(' ').includes('page-Main_Page')) {
-      // Check for class to avoid custom main pages
+    const isCustomMainPage = Boolean(dump?.opts?.customMainPage && (pageTitle === dump.opts.customMainPage.replace(/_/g, ' ') || isMainPage(pageTitle)))
```
benoit74, line 235: Can you explain this? I don't get the rationale behind this complex calculation. 

I'm also concerned about the fact that we parse again the customMainPage argument in a very low level function

